PR #225, opened 2014-12-31
Fix B64 & improve _oasis

Chris00:


Chris00:
I removed the `B64` commit and reformatted the `Description:` so that the itemize is not lost when oasis is parsing it.

rgrinberg:
@Chris00 looks good. Did you run `oasis setup` btw?

Chris00:
On Thu, 1 Jan 2015 11:15:28 -0800, Rudi Grinberg wrote:
> 
> @Chris00 looks good. Did you run oasis setup btw?

I did not commit the generated files — as I wonder why you do not use dynamic mode for development (which means pinning to Git will require oasis to work but that not a big deal to me).

PR #226, opened 2015-01-01
Closes #225

rgrinberg:


rgrinberg:
@Chris00 i took your PR and regenerated the oasis files. What's the cause of this failure?

Chris00:
It says ``ocamlfind: Package `conduit.async' not found`` which is strange.  Can you reproduce it on your machine?  On mine
```
$ ocamlfind query conduit.async
/home/chris/.opam/system/lib/conduit
```
and cohttp compiles fine.

rgrinberg:
Seems to work fine for me as well.

avsm:
I think this is because of the addition of constraints in the _oasis file.  It seems to generate different (and buggy) code in that case.

Chris00:
@gildor478 ?

gildor478:
@Chris00 I acknowledge the problem, and it seems related to oasis 0.4.5. The bug seems to not be present in 0.4.4.

Can you generate a minimal _oasis file with those constraint so that I can check and create a test ?

gildor478:
(and followup on:
https://forge.ocamlcore.org/tracker/index.php?func=detail&aid=1447&group_id=54&atid=291 )

PR #230, opened 2015-01-07
Lwt 2.4.7 renamed blit_bytes_string -> blit_to_bytes

Leonidas-from-XIV:
Thus tests fail to compile. This PR just updates to the new name. This probably also requires a version bump of Lwt im `_opam`, let me know if I should add it.


samoht:
+1 just got that error as well. Would be great to cut a new release with that fix!

rgrinberg:
Yes the constraints will have to be updated. I'm a bit confused why we don't have a lower bounds in the depopts however.

PR #236, opened 2015-01-21
Add `post_form' for async

### LINKED ISSUE #235: No `post_form` for async
It looks like this method is only implemented for lwt. It shouldn't be hard to add it for async. I worked around by by manually doing sth like

```ocaml
let body_str = Uri.encoded_of_query [
    "key1" ["value1"];
    "key2", ["value2"];
    ] in
  let body = Body.of_string body_str in
  Cohttp_async.Client.post ~headers:(Cohttp.Header.init_with "content-type" "application/x-www-form-urlencoded") ~body (Uri.of_string "http://example.com")
```

marklrh:
Resolves #235

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -210,6 +210,13 @@ module Client = struct
 
   let delete ?interrupt ?headers uri =
     call ?interrupt ?headers ~chunked:false `DELETE uri
+
+  let post_form ?interrupt ?headers ~params uri = 
+    let headers = Cohttp.Header.add_opt headers "content-type" "application/x-www-form-urlencoded" in
+    let q = List.map ~f:(fun (k,v) -> k, [v]) (Cohttp.Header.to_list params) in
+    let body = Body.of_string (Uri.encoded_of_query q) in
+    post ?interrupt ~headers ~chunked:false ~body uri
+
 end
 
 module Server = struct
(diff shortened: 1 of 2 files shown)


rgrinberg:
1) The original function is very strange to me. Why would `~params` be typed as a header? It should simply be a key-value list. Nevermind that `Header.t` coincides with that now. @avsm ?

2) I don't like the copy pasting between lwt/async. Can we simply have a backend independent `Request.url_encoded` that will construct the request for lwt/async?

marklrh:
Yeah, simple copy pasting is so easy to do. I think cohttp needs to find a better way to solve those inconsistencies.

rgrinberg:
In this particular case pulling out the request construction to `Request` works good. Still doesn't address 1 however.

avsm:
iirc, the form handling is vestigial from the original ocaml-http library.  I see no reason why it shouldn't be built over the core HTTP library, as it's just a client that adds specific headers and a body format.

rgrinberg:
@avsm and why is `~params` of type `Header.t`? That seems wrong to me.

avsm:
@rgrinberg I agree that `~params` being a `Header.t` seems wrong (although it's almost the same data type, with a slightly different encoding, so it's not fully weird, just named wrongly).

rgrinberg:
@marklrh Can you have a look #257 and make the same change in this PR. Should be good to merge after that.

I've tried to think of a way to get rid of the code duplication but I don't want `Request` to depend on `Body`.

marklrh:
since i deleted the repo days ago, I have to create a new pull at #258

Issue #239, opened 2015-01-26
Problem downloading URL


Chris00:
The following code
```ocaml
open Lwt
open Printf
module Client = Cohttp_lwt_unix.Client

let get url =
  Client.get url >>= fun (_, b) ->
  Cohttp_lwt_body.to_string b >>= fun data ->
  Lwt_io.printf "%s\n%!" data

let () =
  Lwt_main.run(get (Uri.of_string "http://blog.camlcity.org/blog/rss"))
```
fails to download the complete body of the URL.

rgrinberg:
Thanks for the report. This is indeed an issue.

The cause of this is Gerd's blog not returning any `content-length`. In that case cohttp simply tries to fill some fixed size buffer that ends up being a truncated version of the body [1]. We should probably try to read the content more eagerly here with some timeout for nothing being read perhaps. I'm open to suggestions.

P.S. You can reproduce these problems much easier if you simply use the `cohttp_curl_{lwt,async}.native` utility rather than writing up a snippet.

[1] https://github.com/mirage/ocaml-cohttp/blob/master/lib/transfer_io.ml#L80

Chris00:
On Sun, 25 Jan 2015 16:59:09 -0800, Rudi Grinberg wrote:
> 
> Thanks for the report. This is indeed an issue.
> 
> The cause of this is Gerd's blog not returning any
> content-length. In that case cohttp simply tries to fill some fixed
> size buffer that ends up being a truncated version of the body [1].

One should read until the fd gives an end-of-file condition.  And,
yes, a timeout would be nice — but that is not too difficult for the
user to add one with Lwt.

> P.S. You can reproduce these problems much easier if you simply use the
> cohttp_curl_{lwt,async}.native utility rather than writing up a snippet.

Thanks for letting me know.

PR #241, opened 2015-01-26
Transfer_io: read until EOF when body size is unknown

Chris00:
Fixes https://github.com/mirage/ocaml-cohttp/issues/239

### DIFF
--- a/_tags
+++ b/_tags
@@ -555,3 +555,4 @@ true: annot, bin_annot
 <examples/async/receive_post.{native,byte}>: custom
 # OASIS_STOP
 true: principal, strict_sequence, debug, short_paths
+true: annot
--- a/lib/code.mli
+++ b/lib/code.mli
@@ -29,8 +29,10 @@ type success_status =
 
 type redirection_status =
   [ `Multiple_choices (** multiple options for the resource delivered *)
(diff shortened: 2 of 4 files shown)


rgrinberg:
This is good, thanks.

avsm:
This also changes the default read chunk size from 16384 to 4096. Not sure it makes a difference, but worth noting.

Chris00:
I did some (quick nonscientific) testing and, if anything, it seems to be faster (IIRC, 4k is the right amount for disk IO according to “Programmation système en OCaml”).  I though it would put less pressure on the GC.

avsm:
Values greater than 2k (256 words) get allocated directly onto the major heap, so I actually wonder if reading just below that threshold would keep the majority of such buffers in the minor heap.

Chris00:
On Mon, 26 Jan 2015 06:47:05 -0800, Anil Madhavapeddy wrote:
> 
> Values greater than 2k (256 words) get allocated directly onto the major heap, so I actually wonder if reading just below that threshold would keep the majority of such buffers in the minor heap.

OK.  I think it is better to try using read with a byte buffer before optimizing its size.

PR #243, opened 2015-01-26
Add `Cohttp.Header.mem` to check if a header exists

avsm:
This is quicker than retrieving the header, which constructs a
list of values.


rgrinberg:
Yeah this is nicer. 

I'm in the process of fixing all of this automatic header override stuff. Are there any instances where it's legitimate to override the user's value? I can't think of any.

avsm:
Nope...there *might* be a case for appending a header for a multiheader one, but I can't think of any that we override like that.

PR #244, opened 2015-01-26
Improve header handling

rgrinberg:
`add_maybe` and `add_opt` are both bad names I feel. But it seems like enough people are already using `add_opt` https://github.com/search?q=add_opt+language%3AOCaml&type=Code&utf8=%E2%9C%93

Any suggestions for `add_maybe` though?

EDIT: there are a few cases remaining where I've allowed the old "overriding" behavior because the user can call it explicitly. E.g. `add_authorization{,_req}`.

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -252,7 +252,7 @@ module Server = struct
         let keep_alive = Request.is_keep_alive req in
         let flush = Response.flush res in
         let res =
-          let headers = Cohttp.Header.add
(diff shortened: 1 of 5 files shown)


rgrinberg:
@SGrondin you may try this PR as well.

avsm:
How about just being explicit and calling it `Header.add_unless_exists` or something similar?

rgrinberg:
`add_unless_exists` is good with me

rgrinberg:
@avsm ping :)

Issue #198, opened 2014-11-08
1.0 release checklist


avsm:
For Cohttp 1.0 we probably want to fix these issues:
- [ ] bad request handling #146
- [x] empty chunk handling #136 
- [ ] normalize path segments #150 
- [ ] API inconsistencies #200 

Add examples:
- [ ] multipart form request #181 
- [ ] example of http fetch for lwt thats functorised like github (so compiles to js_of_ocaml)
- [x] add examples to README #43 

A test harness to run it through some stress tests to check for fd leakage would be very nice: http://andreas.github.io/2014/09/18/how-to-test-async-ocaml-code/

rgrinberg:
I've added #281 to the milestone since not being able to make AWS is a little embarrassing.

yomimono:
I think this issue is not useful since cohttp 1.0.0 has been released.  If it's tracking pieces of work that are still planned or there is some other purpose served by keeping it around, please feel free to reopen this issue (possibly with a more descriptive title).

PR #245, opened 2015-01-27
raise (Failure ..) -> failwith ..

rgrinberg:
simple tweak

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -25,14 +25,14 @@ module Net = struct
   let lookup uri =
     let host = Option.value (Uri.host uri) ~default:"localhost" in
     match Uri_services.tcp_port_of_uri ~default:"http" uri with
-    | None -> raise (Failure "Net.connect") (* TODO proper exception *)
+    | None -> failwith "Net.connect" (* TODO proper exception *)
     | Some port ->
       let open Unix in
       Addr_info.get ~host [Addr_info.AI_FAMILY PF_INET; Addr_info.AI_SOCKTYPE SOCK_STREAM]
       >>= function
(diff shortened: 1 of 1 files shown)


rgrinberg:
The travis failure is unrelated and should be ignored.

async/cohttp_async.ml
```diff
@@ -25,14 +25,14 @@ module Net = struct
...
-    | None -> raise (Failure "Net.connect") (* TODO proper exception *)
+    | None -> failwith "Net.connect" (* TODO proper exception *)
     | Some port ->
       let open Unix in
       Addr_info.get ~host [Addr_info.AI_FAMILY PF_INET; Addr_info.AI_SOCKTYPE SOCK_STREAM]
       >>= function
       | {Addr_info.ai_addr=ADDR_INET(addr,_)}::_ ->
         return (host, Ipaddr_unix.of_inet_addr addr, port)
-      | _ -> raise (Failure "resolution failed")
+      | _ -> failwith "resolution failed"
```
dbuenzli: I'd rather say you are using exceptions for something unexceptional.

rgrinberg: Yes all of this stuff is very much less than ideal. I think we are saving the refactoring post 1.0 however.

hannesm: why post 1.0? shouldn't such API changes (I suspect `raise` will be transformed to use an `'a option` return type instead) be before 1.0?

rgrinberg: Well we did ask people to participate in making a checklist for 1.0:

https://github.com/mirage/ocaml-cohttp/issues/198

and the issue wasn't included. I guess it's not too late now :P

rgrinberg: @hannesm @dbuenzli #247 addresses this

PR #247, opened 2015-01-27
Net.lookup returns Or_error rather than raising

rgrinberg:
I don't think Net.lookup can throw any other way so Or_error is appropriate.


dbuenzli:
Don't you want to be able to distinguish the errors programatically ? If that's the case you should rather use a proper variant for the error report (sorry don't have time to have a close look at that).

rgrinberg:
@dbuenzli In this case we don't care about distinguishing what particular error it is because the function itself isn't even exported. Also, I agree with using variants for representing errors but it seems like it's core's convention to use `Error.t`.

I'll fix the message however.

https://realworldocaml.org/v1/en/html/error-handling.html

PR #251, opened 2015-02-06
Improve handling of list valued headers

rgrinberg:
Use a hashtable to check if a header key is list valued rather than a
map

### DIFF
--- a/lib/header.ml
+++ b/lib/header.ml
@@ -34,13 +34,16 @@ type t = string list StringMap.t
 
 let user_agent = Conf.user_agent
 
-let headers_with_list_values = List.map LString.of_string [
+let headers_with_list_values = Array.map LString.of_string [|
   "accept";"accept-charset";"accept-encoding";"accept-language";
   "accept-ranges";"allow";"cache-control";"connection";"content-encoding";
   "content-language";"expect";"if-match";"if-none-match";"pragma";
   "proxy-authenticate";"te";"trailer";"transfer-encoding";"upgrade";
-  "vary";"via";"warning";"www-authenticate";
-]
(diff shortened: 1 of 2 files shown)


avsm:
@dsheets, happy with this?  Looks ok to me but deferring to 0.16.0

dsheets:
Does this actually work? We should have test coverage for this... but we probably don't.

Immediately, I don't see where the hash table is initialized with the contents of the list-valued headers.

avsm:
Woops, I thought that happened in the Hashtbl.create section, but it doesn't.

rgrinberg:
@dsheets agreed :confused: i'll fix this and add some tests as well

rgrinberg:
@avsm I've added a (passing) test.

Note that we are getting back the header values in reverse order as we've added them which could be surprising to users.

dsheets:
The order probably matters for some of the headers... I don't know which and I don't know which direction (or if they're even homogeneous). We should probably do a header research/RFC dive to see if the order is specified somewhere.

dsheets:
Actually, I know the order matters for `Accept`. Dunno which way it is supposed to be when there are dupes, though.

rgrinberg:
Unfortunately, I don't much about this subtleties. I'll check out how HTTP is done in other languages. As a first step however, I don't really like how `Header.add` leaves the order up in the air like that. The user should be able to construct a list valued header by providing a list of values.

That being said, this patches does not change the ordering in any way. So perhaps we should create a new issue for this.

rgrinberg:
@avsm the travis failure should be fixed by #261 right?

avsm:
yep.

rgrinberg:
Rebased to rerun travis

PR #255, opened 2015-02-07
Copy basic auth from Uri.userinfo into Authorization header

vklquevs:
I have been wondering the best place for this for a while -- it seems to me like this form of credentials is a common enough case to warrant being here.

This will take e.g. `user:pass` from `http://user:pass@host` and add it to the headers as basic authorization, without overriding any credentials in the supplied headers. It also only works with `user:pass`, not just `user`.

### DIFF
--- a/lib/request.ml
+++ b/lib/request.ml
@@ -30,6 +30,19 @@ let make ?(meth=`GET) ?(version=`HTTP_1_1) ?encoding ?headers uri =
     match headers with
     | None -> Header.init ()
     | Some h -> h in
+  let headers =
+    (* Add user:password auth to headers from uri
+     * if headers don't already have auth *)
+    match Header.get_authorization headers, Uri.userinfo uri with
+      | None, Some userinfo -> begin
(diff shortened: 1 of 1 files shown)


dsheets:
LGTM. I've made a note to actually expose the username/password combo at mirage/ocaml-uri#62. Sorry about this old interface. :-/ @rgrinberg, comments?

rgrinberg:
I think this is good as well but I'm not sure whether not overriding the header is the right behavior. In any case, this is better than not supporting it all.

vklquevs:
I wasn't sure for a while either, but figured it might work better this way round for something along the lines of `wget user@host --ask-password`.

rgrinberg:
Looks good to me. Merging this in and this should be slated for 0.16

@vklquevs for future reference, have a look at https://github.com/OCamlPro/ocp-indent for formatting your code.

vklquevs:
Good stuff, & noted - it's time to stop awkwardly fighting with vim's default indentation!

PR #257, opened 2015-02-07
Fix type in post_form

rgrinberg:
type of params is (string * string list) list and not Header.t

@avsm in theory this is a breaking change


avsm:
This does lead me to wonder why we need a `post_form` in the core library at all, since it's something that can just be built as a normal request.  Or is it just that it's so common...?

rgrinberg:
It's definitely just for convenience. Doesn't make sense to have it in only 1 backend. I think these helpers are useful and I'd like to add more, e.g. multipart. This one in particular was request by a user so I'm inclined to add it.

I'd really like to think of a way to reduce to reduce the duplication though. Something to think about for post 1.0.

avsm:
Yeah, in agreement with all that.

PR #259, opened 2015-02-08
Add version to User-Agent taken from _oasis

Chris00:


avsm:
Thanks! I'm merging this in with a couple of changes:

- Conf shouldnt be an InternalModule; it needs to be packed with the rest of Cohttp.
- Added an ocamldoc header to conf.mli

Chris00:
On Sun, 15 Feb 2015 10:15:53 -0800, Anil Madhavapeddy wrote:
> 
> Thanks! I'm merging this in with a couple of changes:
> 
>   * Conf shouldnt be an InternalModule; it needs to be packed with the rest of
>     Cohttp.
>   * Added an ocamldoc header to conf.mli

I would need to check to be sure but the fact that it is internal does not mean that it isn't packed, right?  It is just that it is not accessible from the outside (no interface).

PR #262, opened 2015-02-15
Add magic mime detection for static file serving

avsm:
Use the `magic-mime` library to add a MIME type by probing filename during static serving in the Lwt/Async backends


lwt/cohttp_lwt_unix.ml
```diff
@@ -75,7 +75,9 @@ module Server = struct
       Lwt_stream.on_terminate stream (fun () ->
         ignore_result (Lwt_io.close ic));
       let body = Cohttp_lwt_body.of_stream stream in
-      let res = Cohttp.Response.make ~status:`OK ~encoding ?headers () in
+      let mime_type = Magic_mime.lookup fname in
+      let headers = Cohttp.Header.add_opt headers "content-type" mime_type in
```
rgrinberg: Aren't we possibly overriding a user provided header value here?

async/cohttp_async.ml
```diff
@@ -304,7 +304,9 @@ module Server = struct
          Reader.open_file filename
          >>= fun rd ->
          let body = `Pipe (Reader.pipe rd) in
-         respond ?flush ?headers ~body `OK
+         let mime_type = Magic_mime.lookup filename in
+         let headers = Cohttp.Header.add_opt headers "content-type" mime_type in
```
rgrinberg: Same comment

lwt/cohttp_lwt_unix.ml
SGrondin: Agreed. I think it's a useful feature, as long as it's only for respond_file and that it doesn't override any headers.

avsm: Yes indeed, it shouldn't be.  I'll make it conditional on an `add_if_exists`

avsm: fixed in bc7438d093cbf90b2911dcbec37017794fb22cd0 and new header function in d00503b48e5162e24b12135852db37409ae866e8

async/cohttp_async.ml
avsm: also fixed in bc7438d093cbf90b2911dcbec37017794fb22cd0

PR #264, opened 2015-02-16
Cohttp proxy

avsm:


_oasis
```diff
@@ -282,6 +282,15 @@ Executable "cohttp-server-lwt"
   Install:            false
   BuildDepends:       cohttp, cohttp.lwt, cmdliner (>= 0.9.4)
 
+Executable "cohttp-proxy-lwt"
```
rgrinberg: Should this be added to the `.install` file?

lib/request.ml
```diff
@@ -107,11 +107,14 @@ module Make(IO : S.IO) = struct
...
+      let uri =
+        match Header.get headers "host" with
         | None -> Uri.of_string path
-        | Some h ->
-          (* XXX: Could be https:// as well... *)
-          Uri.of_string ("http://" ^ h ^ path) in
+        | Some host ->
+           let host_uri = Uri.of_string host in
+           Uri.of_string path |> fun uri ->
+           Uri.with_host uri (Uri.host host_uri)
```
dsheets: Isn't `host` just a bare hostname here? `Uri.of_string host` would then be a relative path...

_oasis
avsm: done in 3dff7769ee314ed64de64ddfbbe32ce29e140d90

lib/request.ml
avsm: Good catch; addressed in aacf6e3ae39d10374a70545cb7c182d7f67added (we need to parse Host:Port as well, hence my desire to use Uri.of_string)

PR #266, opened 2015-02-17
Make Cohttp_lwt use the lwt monad

samoht:
All the IO.t in Cohtt_lwt should be equal to Lwt.t (that's the point of the module)


samoht:
This is necessary to write portable code over cohttp client implementations using lwt (such as mirage and lwt_io).

rgrinberg:
Looks good. Thanks.

Issue #246, opened 2015-01-27
Transfer-Encoding Header is injected into responses


SGrondin:
Another bad case of Header injection.
This one completly breaks the application I'm trying to build.
Once again I don't know the codebase subtleties enough to fix it myself, so I debugged it for you guys as much as I could.

Basically, `Server.respond`ing with a stream will set the `Transfer-Encoding` to chunked. That makes sense, but if I already know the expected length of the stream I'd like to be able to specify the `Content-Length` myself and not have Cohttp inject the `Transfer-Encoding` header for me.

[Server.respond](https://github.com/mirage/ocaml-cohttp/blob/d0aa1cb55b364ba085384ef0fe3eaa6dce9554be/lwt/cohttp_lwt.ml#L298)
calls
[Cohttp_lwt_body.transfer_encoding](https://github.com/mirage/ocaml-cohttp/blob/86b4042c0e95f00eeeb7a02a474a3c7b8091e968/lwt/cohttp_lwt_body.ml#L87)

It would make more sense if it ONLY injected `Transfer-Encoding` if `Content-Length` is not present. In my current application, I have to manually delete `Content-Length` because it breaks `curl` and others if `Transfer-Encoding` is also present. But that also breaks other clients that did not request a Chunked response and got one (the extra control bytes for Chunked blow up everything).

Sure, I could buffer the stream and only respond with a string once I have everything, but that's ugly and horrible for performance.

Thanks for looking into it, it's probably not a huge fix and it would help me tremendously.

talex5:
You should be able to make the response manually. I use:

    let body = Cohttp_lwt_body.of_stream data in
    let headers = Cohttp.Header.init_with "Content-Length" (Int64.to_string size) in
    let encoding = Cohttp.Transfer.Fixed size in
    let res = H.Response.make ~status:`OK ~encoding ~headers () in
    return (res, body)

SGrondin:
Thanks @talex5 that worked. I'm not sure how I missed that feature from reading the docs. Obviously it's low priority, but I think it could be good if Server.respond didn't use Chunked mode when a length is already provided.

Nice blog, btw.

rgrinberg:
Yes cohttp should be smarter and default to fixed if the response is provided as a string.

avsm:
I've actually just fixed this in my proxy branch. Commit incoming

avsm:
My fix was for requests, not responses (mirage/ocaml-cohttp#268).

[...]

PR #268, opened 2015-02-17
Better Request handling of transfer-encoding

avsm:
* `Request.has_body` does not permit a body to be set for methods that RFC7231 forbids from having one (`HEAD`, `GET` and `DELETE`). 
* `Request.make` does not inject a `transfer-encoding` header if there
is no body present in the request (#246).

### DIFF
--- a/lwt/cohttp_lwt.ml
+++ b/lwt/cohttp_lwt.ml
@@ -298,7 +298,14 @@ module Make_server(IO:IO)
     Filename.concat docroot rel_path
 
   let respond ?headers ?(flush=true) ~status ~body () =
-    let encoding = Cohttp_lwt_body.transfer_encoding body in
+    let encoding = 
+      match headers with
+      | None -> Cohttp_lwt_body.transfer_encoding body
+      | Some headers ->
+         match Header.get_transfer_encoding headers with
+         | Transfer.Unknown -> Cohttp_lwt_body.transfer_encoding body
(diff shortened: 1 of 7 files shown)


lwt/cohttp_lwt.ml
```diff
@@ -298,7 +298,14 @@ module Make_server(IO:IO)
     Filename.concat docroot rel_path
 
   let respond ?headers ?(flush=true) ~status ~body () =
-    let encoding = Cohttp_lwt_body.transfer_encoding body in
```
rgrinberg: I wouldn't mind seeing this in its own function for testing purposes. Although I wouldn't want the user to be exposed to it...

avsm: It's probably better to test the expected behaviour of `respond` isnt it?  With and without user-specified headers and a few different body formats.

rgrinberg: It's a little more work but you're right that it's better from a rigor perspective.

I did originally think that this logic could be reused between async/lwt. Not the case however.

rgrinberg:
+1

lwt/cohttp_lwt.ml
avsm: That's a very good point though. There's quite a bit of subtle logic that could be pulled into Cohttp itself.  Am putting together an Async proxy now

Issue #136, opened 2014-05-05
Empty chunk


dinosaure:
Hello,

During my migration with ocsigenserver and cohttp when we had a chunked transfer with `Lwt_stream.t` we had some blocks that were empty (but that does not mean as long as the transfer was completed).

Thus it happened that sent cohttp `0\r\n` (which corresponds to the end of the chunk mode sent in the HTTP protocol) when we had not finished. Cohttp could very well ignore empty chunks (at least, that's what we do on our side here ocsigen/ocsigenserver@dadf5623).

I think the line to edit it [here](https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/transfer_io.ml#L48).

rgrinberg:
Fixed in master. Will be released in 0.16.0

PR #270, opened 2015-03-01
Skip empty chunk in Transfer_io.write

rgrinberg:
Addresses #136 I've decided against flushing on `""` because we already have a dedicated flush function.

### DIFF
--- a/lib/transfer_io.ml
+++ b/lib/transfer_io.ml
@@ -95,20 +95,25 @@ module Make(IO : S.IO) = struct
     | Fixed len -> Fixed.read ~remaining:(ref len)
     | Unknown -> Unknown.read
 
+  let write_ignore_blank writer io s =
+    if String.length s = 0
+    then return ()
+    else writer io s
+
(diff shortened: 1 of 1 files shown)


lib/transfer_io.ml
```diff
@@ -95,20 +95,25 @@ module Make(IO : S.IO) = struct
     | Fixed len -> Fixed.read ~remaining:(ref len)
     | Unknown -> Unknown.read
 
+  let write_ignore_blank writer io s =
+    if String.length s = 0
+    then return ()
```
avsm: prefer `return_unit` here to minimise allocation

rgrinberg: This code is functorised over `IO` so we'd need to add `return_unit` in `IO`.

avsm: good point...

rgrinberg: Any other objections?

avsm: lgtm, will merge

PR #273, opened 2015-03-04
Factor cohttp_server and add file sizes to listings

dsheets:
No longer a single file but much easier to maintain and with much less code duplication. Also, I wanted to know the sizes of downloads before I attempted them.

### DIFF
--- a/bin/cohttp_server_lwt.ml
+++ b/bin/cohttp_server_lwt.ml
@@ -23,28 +23,11 @@ open Lwt
 open Cohttp
 open Cohttp_lwt_unix
 
-let ( / ) = Filename.concat
+open Cohttp_server
 
 let serve_file ~docroot ~uri =
   let fname = Server.resolve_local_file ~docroot ~uri in
   Server.respond_file ~fname ()
-
-let compare_kind = function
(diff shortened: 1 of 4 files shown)


bin/cohttp_server_lwt.ml
```diff
@@ -81,58 +64,31 @@ let handler ~info ~docroot ~verbose ~index (ch,conn) req body =
...
-      ~body:(sprintf "<html><body><h2>Not Found</h2>
-      <p><b>%s</b> was not found on this server</p>
-      <hr />%s</body></html>" path info)
+      ~body:(html_of_not_found path info)
       ()
     else fail e
   | e -> fail e
   )
 
-let string_of_sockaddr = function
```
rgrinberg: So this was unused I assume.

rgrinberg:
+1 - Very nice.

bin/cohttp_server_lwt.ml
dsheets: Yes, unused but perhaps should be used somewhere. The async and lwt servers still don't have equivalent console logging.

PR #274, opened 2015-03-04
Simplify the Net signature which needs to be provided for lwt servers

samoht:
Not sure to understand why the `Net` module needs to be passed as an argument of the server functor. It is not supposed to connect to anything (I think).

### DIFF
--- a/lwt/cohttp_lwt.ml
+++ b/lwt/cohttp_lwt.ml
@@ -219,9 +219,6 @@ module type Server = sig
   module Request : Request
   module Response : Response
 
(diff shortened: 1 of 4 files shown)


rgrinberg:
+1. Surprised this doesn't break anything downstream. @avsm please review as well

samoht:
ping? Would be very handy to have that in the next release so I can merge https://github.com/mirage/mirage-http/pull/10

avsm:
This makes sense to me. Will merge shortly.

rgrinberg:
ping

PR #276, opened 2015-03-06
Fix typo

leowzukw:
Forgotten a ' in examples/async/receive_post.ml


rgrinberg:
Hmm I think these nested quotes are just confusing. Would it be possible to change it to `$ curl -X POST -d 'foo bar' http://localhost:%d` instead?

leowzukw:
Why not!

PR #277, opened 2015-03-11
Extend async file server example to run in http or https mode

trevorsummerssmith:
As per our conversation on the mailing list -- here is a minor change to let the async file server server ssl.


avsm:
Thanks! Merged into 0.16.0 with slight tweaks (651cbfeda3aecd1c7283e0b5ba21277e8465617e)

PR #282, opened 2015-03-22
Lwt_unix_io: Split large reads

j0sh:
With the Lwt backend, reads hang if trying to fetch more than
Sys.max_string_length due to a byte allocation failure in
Lwt_io.read. Problematic for large uploads, etc.

Upper layer callers (eg, Transfer_io) seem to handle recombining
these split reads without any issues.


avsm:
Looks good to me, thanks! Will merge shortly

Issue #260, opened 2015-02-12
Will respond_file use magic-mime ?


zoggy:
Hello,

Do you have plans to make respond_file add a mime-type in the response header (for example using magic-mime ? :-) )

avsm:
Yep, that is sensible.  It didn't make it into 0.15.2, but I'll put that into a point release shortly.

dsheets:
Slightly off-topic but it seems that magic-mime should actually be named something like "mime-ext-db" and the "magic" component should be reserved for the actually magical function like Unix `file(1)`.

Issue #252, opened 2015-02-06
Uninstalling cohttp doesn't remove binaries


rgrinberg:
E.g.

```
$ opam install cohttp lwt
$ which `cohttp-server-lwt`
$ opam remove cohttp
$ which `cohttp-server-lwt`
```
That binary is still there.

Issue #248, opened 2015-02-03
Proxied requests confuse the URI parser


SGrondin:
When I set the system-wide HTTP proxy on Debian to forward all traffic to a Cohttp application, the calls come in as `GET //simongrondin.name/files/fifty.txt` instead of `GET http://simongrondin.name/files/fifty.txt`.

The URI parser gets confused and interprets them like this:
```
Uri.to_string (Request.uri req)
> http://simongrondin.namehttp//simongrondin.name/files/fifty.txt
```

rgrinberg:
Are there any headers that come with this `GET //simongrondin.name/files/fifty.txt` request? I'm looking at   a `host` header specifically.

SGrondin:
Host is, as expected, `Host: simongrondin.name`

rgrinberg:
OK, so this is the culprit: https://github.com/mirage/ocaml-cohttp/blob/master/lib/request.ml#L97

SGrondin:
By the way it's easier to reproduce by setting the Firefox HTTP proxy setting to a local cohttp server.

dsheets:
@rgrinberg are you on this? It looks like <https://github.com/mirage/ocaml-cohttp/commit/d04701fa4ae1161eec148ad5871647ef08109bc1#diff-daa4e580cf195143c70372b6183a5a40R140> should really be using  proper `Uri` functional updates rather than string concatenation (will handle relative paths correctly).  Also, I'd recommend switching to a scheme-relative identifiers.

SGrondin:
This is the absolutely horrible hack I'm using for now, but it works until a fix is (hopefully) released in 0.16, in case anyone googled and finds this issue.

```ocaml
let fix_uri uri =
	let ungarble uri host protocol_len =
		uri
		|> fun uri -> ((Uri.with_host uri (Some (String.slice host 0 (-protocol_len)))), (String.length host - protocol_len))
		|> fun (uri, len) -> Uri.with_path uri (String.slice (Uri.path uri) (len + 2) 0)
	in
	match (uri |> Uri.host |> Option.value ~default:"") with
	| host when String.is_suffix ~suffix:"http" host -> ungarble uri host 4
	| host when String.is_suffix ~suffix:"https" host -> ungarble uri host 5
	| host when (((String.length host) / 2) mod 2 = 0) && ((String.slice host (String.length host / 2) 0) = (String.slice host 0 (String.length host / 2))) ->
[...]

PR #283, opened 2015-03-22
0.16.0

avsm:
Consolidated pull request for:

Compatibility breaking interface changes:
* [lwt] Fix types in `post_form` to be a `string * string list` instead
  of a `Header.t` (#257)
* Simplify the `Net` signature which needs to be provided for Lwt servers
  to not be required.  Only the Lwt client needs a `Net` functor argument
  to make outgoing connections. (#274)

New features and bug fixes:
* Copy basic auth from `Uri.userinfo` into the Authorization header
  for HTTP requests. (#255)
* Install binaries via an OPAM `.install` file to ensure that they are
  reliably uninstalled. (#252)
* Use the `magic-mime` library to add a MIME type by probing filename
  during static serving in the Lwt/Async backends. (#260)
* Add `Cohttp.Header.add_opt_unless_exists` to set a header only if
  an override wasn't supplied, and to initialise a fresh Header value
  if none is present.
* Do not override user-supplied headers in `post_form` or `redirect`.
* `Request.has_body` does not permit a body to be set for methods that
  RFC7231 forbids from having one (`HEAD`, `GET` and `DELETE`).
* `Request.make` does not inject a `transfer-encoding` header if there
  is no body present in the request (#246).
* `Server.respond` no longer overrides user-supplied headers that
  specify the `content-length` or `transfer-encoding` headers (#268).
* `cohttp_server_lwt` and `cohttp_server_async` now include sizes in
  directory listing titles

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ script: bash -ex .travis-opam.sh
 env:
   - PACKAGE="cohttp" OCAML_VERSION=4.01   DEPOPTS="async lwt js_of_ocaml" POST_INSTALL_HOOK="opam install github cowabloga" 
   - PACKAGE="cohttp" OCAML_VERSION=latest DEPOPTS="async lwt js_of_ocaml" POST_INSTALL_HOOK="opam install github cowabloga"
+  - PACKAGE="cohttp" OCAML_VERSION=latest DEPOPTS="async_ssl" POST_INSTALL_HOOK="opam install github cowabloga"
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,7 @@ Compatibility breaking interface changes:
   to make outgoing connections. (#274)
 
 New features and bug fixes:
+* Add a `cohttp-lwt-proxy` to act as an HTTP proxy. (#248)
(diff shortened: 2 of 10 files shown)


rgrinberg:
Don't forget https://github.com/mirage/ocaml-cohttp/pull/278 in CHANGES

avsm:
#248 is the only release blocker for 0.16.0 now i think.

SGrondin:
@rgrinberg thanks, I'm glad to see it'll be included in 0.16.0

@avsm let me know if you're having issues reproducing #248 and I can try to make a small test file to reproduce the bug. Currently it happens if I remove the call to `fix_uri` from [this line](https://github.com/Mashape/HARchiver/blob/a64def990ee6b357e9113268128444983a846ac4/src/proxy.ml#L72). If you end up running that codebase locally I apologize for the non-idiomatic code in places, I'm still learning the language

rgrinberg:
@SGrondin I think making a small test case for us would greatly ease fixing this problem. Note to mention that we'll be able to add it to our test suite to prevent regression.

PR #284, opened 2015-03-23
Add new method Client.call which takes a Request.t

### LINKED ISSUE #281: Add generic execute Request.t function to Cohttp_async.Client
The current client interface does not allow one to exercise full control over the headers that are sent. This is limiting. For example, if one wants to send requests to the AWS S3 API one must create a [signed hash the contains all headers prior to transmission] (http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html). It is not possible to do this with the current, Client interface, because headers are added after the user's call. I could imagine other situations where it is desirable to construct a Request.t and then send it.

I would propose a call that takes a Request.t and executes it. I am happy to submit a pull request for this if you all like this idea. Suggested signature:
```ocaml
val call :
   ?interrupt:unit Deferred.t ->
   ?body:Body.t ->
   Request.t ->
   (Response.t * Body.t) Deferred.t
```

trevorsummerssmith:
** Interface breaking change **

The old Client.call is renamed to Client.call_infer_encoding.
The new method is what you send is what you get -- no headers
are added/removed or edited.

Closes #281

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -155,26 +155,9 @@ end
 
 module Client = struct
 
-  let call ?interrupt ?headers ?(chunked=false) ?(body=`Empty) meth uri =
-    (* Figure out an appropriate transfer encoding *)
-    let req =
-      match chunked with
-      | false ->
-          Body.disable_chunked_encoding body
-          >>| fun (body, body_length) ->
-          Request.make_for_client ?headers ~chunked ~body_length meth uri
-      | true -> begin
-          Body.is_empty body
-          >>| function
-          | true -> (* Dont used chunked encoding with an empty body *)
-            Request.make_for_client ?headers ~chunked:false ~body_length:0L meth uri
-          | false -> (* Use chunked encoding if there is a body *)
-            Request.make_for_client ?headers ~chunked:true meth uri
-      end
-    in
-    req >>= fun req ->
+  let request ?interrupt ?(body=`Empty) req =
     (* Connect to the remote side *)
-    Net.connect_uri ?interrupt uri
+    Net.connect_uri ?interrupt (Request.(req.uri))
     >>= fun (ic,oc) ->
     Request.write (fun writer -> Body.write Request.write_body body writer) req oc
     >>= fun () ->
@@ -191,6 +174,26 @@ module Client = struct
           Deferred.all_ignore [Reader.close ic; Writer.close oc]
         );
         res, `Pipe rd
+
+  let call ?interrupt ?headers ?(chunked=false) ?(body=`Empty) meth uri =
+    (* Create a request, then make the request.
+       Figure out an appropriate transfer encoding *)
+    let req =
+      match chunked with
+      | false ->
+          Body.disable_chunked_encoding body
+          >>| fun (body, body_length) ->
+          Request.make_for_client ?headers ~chunked ~body_length meth uri
+      | true -> begin
+          Body.is_empty body
+          >>| function
+          | true -> (* Dont used chunked encoding with an empty body *)
+            Request.make_for_client ?headers ~chunked:false ~body_length:0L meth uri
+          | false -> (* Use chunked encoding if there is a body *)
+            Request.make_for_client ?headers ~chunked:true meth uri
+      end
+    in
+    req >>= fun req -> request ?interrupt ~body req
 
   let get ?interrupt ?headers uri =
     call ?interrupt ?headers ~chunked:false `GET uri
--- a/async/cohttp_async.mli
+++ b/async/cohttp_async.mli
@@ -52,6 +52,25 @@ end
 
 module Client : sig
 
+  (** Send an HTTP request with an arbitrary body
+      The request is sent as-is. *)
+  val request :
+    ?interrupt:unit Deferred.t ->
+    ?body:Body.t ->
+    Request.t ->
+    (Response.t * Body.t) Deferred.t
+
+  (** Send an HTTP request with arbitrary method and a body
+      Infers the transfer encoding *)
+  val call :
+    ?interrupt:unit Deferred.t ->
+    ?headers:Cohttp.Header.t ->
+    ?chunked:bool ->
+    ?body:Body.t ->
+    Cohttp.Code.meth ->
+    Uri.t ->
+    (Response.t * Body.t) Deferred.t
+
   (** Send an HTTP GET request *)
   val get :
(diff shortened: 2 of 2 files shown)


trevorsummerssmith:
I didn't add any tests because 1) it seems that there's not much unit test coverage in the codebase already and 2) It seems you want to get this added asap and 3) all the current client tests now call through this function.

I'd be happy to add a more complete test (of the style already in the codebase) soon. I was thinking a simple AWS S3 get example would be exemplary, however it would require dependencies on something with a sha256 so that's actually not such a good idea.

Let me know your thoughts. Thanks.

trevorsummerssmith:
The build failed due to the test taking over 50 minutes. It didn't appear to be due to the pull request.

rgrinberg:
Looks good but is it possible to think of a name other than call to avoid the interface breakage? Leave `call` as is and use another name. @avsm @dbuenzli what do you think?

avsm:
Why not just add an optional label onto `call` to disable the coding detection?  That would be an interface breakage, but much easier to fix up.

avsm:
> however it would require dependencies on something with a sha256 so that's actually not such a good idea.

You can grab this from `nocrypto`, which is fine to have as a test-only dependency in cohttp. (it's used in the pure OCaml TLS stack as well)

rgrinberg:
I'm not such a fan of adding more optional arguments because we already have plenty and they're not really aesthetically pleasing when used this much IMHO. Also, there's no default value here that makes sense anyway and we're only choosing one to avoid breakage.

Regardless, I'm interested to hear what others have to say.

trevorsummerssmith:
@rgrinberg As a contributor I agree with your take.

For what it's worth -- I think it is bad to have generic names mean something specific.

There's not a clear generic name (call, exec, do_request) to share with both the current `call` and this new `call`. If you were to do a multi-release break we could have this release keep `call` as `call`, have the new function be named `call_new` or something else that is clear, then next release go with `call` and `call_infer_encoding`.

dbuenzli:
Sorry I don't have a clear view of the API here, it would be nice to publish the ocamldoc somewhere (whichever state it may be in) to make it easier to have a browsable bird's eye view over it. But IIUC the optional arguments would be mutually exclusive no ? So you'd have to check them for consistency an raise `invalid_arg` if they are inconsistant which seems a little bit silly (if that's really the case).  

From a non HTTP-expert user point of view (and not having these things in my mind at the moment) I find `call_infer_encoding` more confusing that clarifying. I would simply keep the old one as `call` and name the new one `call_with_request`. The names may not be perfect but from a user's point of view their signature make it clear what both functions require from you and you don't break compat (if that's important).

dbuenzli:
And maybe `call_uri` could be added as a synonym for `call` and `call` be deprecated.

dbuenzli:
And then maybe ponder `call_with_request` vs `call_request` to be more regular with `call_uri`.

trevorsummerssmith:
@dbuenzil I like that idea `call_uri` and `call_request`. I dislike having `call` do some other things but be the most simple function.

avsm:
I'm pretty torn here.  The issue with HTTP is that it's intrinsically a very complex protocol with a loose specification, and there are many combinations of valid APIs that might work for different endpoints. In this particular case, `call` is currently doing the right thing by figuring out the right transfer encoding headers.  But then we have a case of an endpoint API in Amazon that forbids proxies and the (perfectly valid) addition of extra headers!

With additional optional arguments to `call`, we can specify various quirks that most callers will not care about.  In this case, the only thing that `call` does is to figure out transfer encoding before serialising the request, and so having an optional argument lets us drop down to "bare" behaviour where it does nothing special.  I guess we could have a `direct_call` function that always guarantees that it does nothing else other than a serialisation, and have that invoked by the existing `call` logic.

I wonder what this API will look like with the addition of HTTP2... at least body handling will be simpler!

dbuenzli:
@avsm It seems to me that this quite fits the `call_uri` and `call_request` idea, in the sense that "calling an `uri`" can be seen as a higher-level concept (i.e. the library will do more for you), than using a raw `Request.t` value (from what I see from this PR). In any case if you prefer to keep `call` I'm not very fond of `direct_call` and would prefer `raw_call`.

dbuenzli:
@avsm Actually I'm not sure I understood your point. Is it that you'd prefer to have `call` and `{raw,direct}_call` to make sure that casual users prefer to use `call` because it will do the right thing(tm) most of the time ? If that's the case then I think that's a good point and that this scheme should be preferred.

rgrinberg:
@avsm I agree with the sentiment but I don't see how it supports adding an optional argument rather than a new call. Daniel also points that some optional arguments can be mutually exclusive. For example, `?chunked` and a potential `?infer_encoding` argument. You could change `?chunked` to be some sort of variant to better represent the options but that's adding more breakage that a new call will simply avoid.

I have no preference between `direct_call`, `raw_call`, and `call_request` (please no `with`s anywhere)

trevorsummerssmith:
The `direct_call/raw_call/call_request` and `call` I support. Someone make a final decision here and I'll resend the pull request.

fxfactorial:
@trevorsummerssmith I just tried your code and for whatever reason I also was getting compilation errors. (silly question, did you add that code on a fresh clone of cohttp?) In any case I did basically the same thing on a branch and got to talking to s3, so that's nice. I can give the pull request too if desired.

trevorsummerssmith:
@fxfactorial what were the errors?

With this pull request, and any pull request, you can see where the pull request was created by looking at the commit. The parent of this commit is 7f2d04bd3d69b18038feb2838ec9aed326027552, which still appears to be most recent.

If you look at the [Travis log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/55478911/log.txt) you will that the reason it failed is because it took longer than 50 minutes.

fxfactorial:
@trevorsummerssmith Sorry, I don't have them left, I was getting them when using opam pin add and having opam try to compile the package.

trevorsummerssmith:
@avsm Pushed amended commit with new naming. Once this PR, and #290 are merged in #281 should be able to be closed with the semantics you discuss in #281 for both functions. I will add the s3 example later today.

avsm:
@trevorsummerssmith looks perfect, thanks! If you push your S3 test, I'll merge it after adding an Lwt equivalent and testing them both.

avsm:
(Would also be good if @rgrinberg could ack the semantics I described in #281 before merging)

PR #288, opened 2015-03-25
Add curl like -X, --request param to cohttp curl

rgrinberg:
Allows to set http methods on requests


rgrinberg:
Merging as this isn't too controversial

avsm:
lgtm!

PR #290, opened 2015-03-28
Supplying a `content-range` header in any client request overrides any other encoding preference.

avsm:
Previously, the `~encoding` label took precedence, which could be
overridden by intermediate calls like `Request.make_for_client`.
Now, the client-supplied headers always take effect over anything
else.

Partially addresses #281

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,7 +8,7 @@ Compatibility breaking interface changes:
   to make outgoing connections. (#274)
 
 New features and bug fixes:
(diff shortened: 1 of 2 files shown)


rgrinberg:
+1 looks good. IMO we should consider getting rid of the encoding field altogether and control it solely through the headers.

Also, we really need some tests for all of this

avsm:
The encoding field was there for convenience, so that could be subsumed by `Header` helper functions to generate appropriate ones.

PR #292, opened 2015-03-29
Add Header_io.write

rgrinberg:
Reduce copy pasting between request and response


rgrinberg:
Merging as this just gets rid of simple code duplication

avsm:
+1

PR #293, opened 2015-03-29
Basic multipart support for lwt client

rgrinberg:
- Doesn't support JS
- Doesn't support multipart/mixed
- Doesn't do base64/whatever encoding

The curl support was mostly added to help me test.

But is good enough for my own needs for now. Feedback welcome.


avsm:
Deferring to 0.18.0 until there are tests included.

PR #294, opened 2015-03-29
Add async client example s3_cp

trevorsummerssmith:
Showcases real world usage of the async client library to access
AWS S3. This requires a dependency on nocrypto. This has been added
to _oasis and merlin.

This got to be much larger than I was hoping so feel free to let me know that this is a bit off the deep end and you all want something different. I thought about it for a while though and do think this represents simple usage of the api, even though the file is longer than I would have liked.

There's a few "compat" functions in there. I'm going to open a few tickets in the relevant projects to see if there is interest in adding functionality that will address those needs (or if I'm missing something). Hopefully some of that will lighten the code length a bit.

Eager to hear your thoughts!


_oasis
```diff
@@ -327,6 +327,15 @@ Executable "async-hello-world"
   Install:            false
   BuildDepends:       cohttp, cohttp.async
 
+Executable "async-s3-cp"
+  Path:               examples/async
+  MainIs:             s3_cp.ml
+  Build$:             flag(async)
```
rgrinberg: We should probably have another flag for this.

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+    (* Core.Std.Time doesn't have a format function that takes a timezone *)
+    let d, s = Time.to_date_ofday ~zone:Time.Zone.utc t in
+    let open Core.Span.Parts in
+    let {hr; min; sec; _} = Time.Ofday.to_parts s in
+    Printf.sprintf "%sT%.2d%.2d%.2dZ"
+      (Date.to_string_iso8601_basic d) hr min sec
+end
+
+type region = [
+  | `ap_northeast_1 (* Asia Pacific (Tokyo) *)
```
rgrinberg: I strongly prefer making all polymorphic variants start with an uppercase letter although I know other people don't really care.

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+    let auth_header = Auth.auth_request ~now:time
+        ~hashed_payload ~region:conf.region ~service:`S3
+        ~aws_access_key:conf.aws_access_key
+        ~aws_secret_key:conf.aws_secret_key request
+    in
+    let headers' = headers @ auth_header in
+    Cohttp.Request.(request.headers <- (Header.of_list headers'));
+    match meth with
+    | `PUT -> Client.request
+                ~body:(Option.value ~default:"" body |> Body.of_string)
```
rgrinberg: `~default` should probably be ````Empty```. Although would be best if we'd expose that constructor as `Body.empty`

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+          dst;
+        Printf.printf "Wrote s3://%s to %s\n" (src.bucket ^ src.objekt) dst;
+        return ()
+      | _ -> Printf.printf "Error: %s\n" (Sexp.to_string (Response.sexp_of_t resp));
+        return ()
+    end
+  | LocaltoS3 (src, dst) ->
+    begin
+      let body = In_channel.with_file src ~f:(fun ic -> In_channel.input_all ic) in
+      make_request ~body conf ~meth:Code.(`PUT) ~bucket:dst.bucket ~objekt:dst.objekt
```
rgrinberg: You don't need to qualify polymorphic variant constructors.

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+type cmd =
+    S3toLocal of s3path * string
+  | LocaltoS3 of string * s3path
+
+let determine_s3_parts s =
+  (* Takes: string of the form s3://<bucket>/<object> *)
+  let s = String.drop_prefix s 5 in
+  let parts = String.split ~on:'/' s in
+  match parts with
+  | bucket::rst -> {bucket=bucket; objekt=(String.concat ~sep:"/" rst)}
```
rgrinberg: You can pun the `bucket` away here

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+    let request = Request.make ~meth
+        ~headers:(Header.of_list headers)
+        uri in
+    let auth_header = Auth.auth_request ~now:time
+        ~hashed_payload ~region:conf.region ~service:`S3
+        ~aws_access_key:conf.aws_access_key
+        ~aws_secret_key:conf.aws_secret_key request
+    in
+    let headers' = headers @ auth_header in
+    Cohttp.Request.(request.headers <- (Header.of_list headers'));
```
rgrinberg: Any reason for not using a functional update here?

I'm really itching towards getting rid of request/response mutability altogether so I'm just warning you not to rely on it :D

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+
+let string_of_service = function
+  | `S3 -> "s3"
+
+module Auth = struct
+  (** AWS S3 Authorization *)
+
+  let digest s =
+    (* string -> sha256 as a hex string *)
+    Nocrypto.Hash.(digest `SHA256 (Cstruct.of_string s))
```
rgrinberg: Can you add cstruct to the list of dependencies in oasis?

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+let run region_str aws_access_key aws_secret_key src dst () =
+  (* nb client does not support redirects or preflight 100 *)
+  let open S3 in
+  let region = region_of_string region_str in
+  let conf = {region; aws_access_key; aws_secret_key} in
+  match determine_paths src dst with
+  | S3toLocal (src, dst) ->
+    begin
+      make_request conf ~meth:Code.(`GET) ~bucket:src.bucket ~objekt:src.objekt
+      >>= fun (resp, body) ->
```
rgrinberg: if you change this to `>>|` you won't need the `return ()` in the match cases

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+    let date_str = Date.to_string_iso8601_basic date' in
+    let date_key = mac (Cstruct.of_string ("AWS4"^secret_access_key)) date_str in
+    let date_region_key = mac date_key (string_of_region region) in
+    let date_region_service_key = mac date_region_key (string_of_service service) in
+    let signing_key = mac date_region_service_key "aws4_request" in
+    signing_key
+
+  let auth_request ?now ~hashed_payload ~region ~service ~aws_access_key ~aws_secret_key request =
+    (* Important use the same time for everything here *)
+    let time = match now with None -> (Time.now ()) | Some t -> t in
```
rgrinberg: `Option.value` should simplify this a little (never mind that `Time.now` will always be evaluated)

rgrinberg:
On the plus side, this is a nice and educational example of a very typical use case of cohttp. On the other hand, there's quite a lot of cruft here to workaround deficiencies in various api's. I vote to include this but we should really work towards simplifying this so that cohttp doesn't look bad to the unexpected by passer.

@trevorsummerssmith This is more of a side note regarding your email. IMO, this example actually shows that it's relatively easy to construct backend independent code in simple cases such as this. Simple clients such as this one should just provide functions to construct `Request.t`'s and translate `Response.t`'s. Backend specific code (if it's even necessary) is only responsible for doing the `Request.t -> Response.t` conversion.

rgrinberg:
Finally, in the future please separate out `oasis setup` changes into it's own commit.

examples/async/s3_cp.ml
trevorsummerssmith: We don't -- we do need some way to take in the user's input as a string. I was conflicted about including it. Given that this is a short example I'll nix it.

examples/async/s3_cp.ml
trevorsummerssmith: Thanks.

examples/async/s3_cp.ml
trevorsummerssmith: Done.

_oasis
trevorsummerssmith: Agreed.

examples/async/s3_cp.ml
trevorsummerssmith: Thanks.

examples/async/s3_cp.ml
trevorsummerssmith: Nope, no good reason. I would prefer that as well. Updated.

examples/async/s3_cp.ml
trevorsummerssmith: Done.

examples/async/s3_cp.ml
trevorsummerssmith: There's the Body.to_string below that. So it seems that one of them can be >>| but the other needs to be >>=. Regardless, it cleans up the code. Thanks!

examples/async/s3_cp.ml
trevorsummerssmith: Thanks.

examples/async/s3_cp.ml
```diff
@@ -0,0 +1,390 @@
...
+        return ()
+      | _ -> Printf.printf "Error: %s\n" (Sexp.to_string (Response.sexp_of_t resp));
+        return ()
+    end
+  | LocaltoS3 (src, dst) ->
+    begin
+      let body = In_channel.with_file src ~f:(fun ic -> In_channel.input_all ic) in
+      make_request ~body conf ~meth:Code.(`PUT) ~bucket:dst.bucket ~objekt:dst.objekt
+      >>= fun (resp, body) ->
+      match Cohttp.Response.(resp.status) with
```
trevorsummerssmith: Done.

trevorsummerssmith:
@rgrinberg pushed updates responding to your comments. Thanks for taking the time to provide all of your feedback. Much appreciated.

trevorsummerssmith:
Also in response to the cruft. Here's relevant tickets:

https://github.com/mirage/ocaml-uri/issues/66
https://github.com/mirage/ocaml-uri/issues/65
https://github.com/mirleft/ocaml-nocrypto/issues/50

trevorsummerssmith:
@rgrinberg pushed a minor change that uppercased the variants (forgot this last time) and slightly simplified the encode_query_string function, after a discussion with the Uri folks https://github.com/mirage/ocaml-uri/issues/66

avsm:
This is a fantastic PR!

rgrinberg:
@trevorsummerssmith Thanks again!

PR #295, opened 2015-03-30
Actually installs binaries fixing regression in bd626e for #252

dsheets:
The previous version of `cohttp.install` broke binary installation. This fixes it for me, at least.

### DIFF
--- a/cohttp.install
+++ b/cohttp.install
@@ -1,7 +1,16 @@
 bin: [
-  "?cohttp-server-async"
-  "?cohttp-server-lwt"
-  "?cohttp-curl-async"
-  "?cohttp-curl-lwt"
-  "?cohttp-proxy-lwt"
+  "?cohttp_server_async.byte"   { "cohttp-server-async" }
+  "?cohttp_server_async.native" { "cohttp-server-async" }
+
+  "?cohttp_server_lwt.byte"     { "cohttp-server-lwt" }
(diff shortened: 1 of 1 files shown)


cohttp.install
```diff
@@ -1,7 +1,16 @@
 bin: [
-  "?cohttp-server-async"
-  "?cohttp-server-lwt"
-  "?cohttp-curl-async"
-  "?cohttp-curl-lwt"
-  "?cohttp-proxy-lwt"
+  "?cohttp_server_async.native" { "cohttp-server-async" }
+  "?cohttp_server_async.byte"   { "cohttp-server-async" }
```
avsm: does the ordering matter? I'd prefer to install native code ones rather than bytecode.

dbuenzli: The opam manual says that files will be installed in sequence. https://opam.ocaml.org/doc/manual/dev-manual.html#sec25

dsheets: Maybe the order matters but `_oasis` is targeting `best` so only one of these should exist.

dbuenzli: (though personally I don't find that relying on this property is very clean, that's the reason why .install files should really be generated by the build system).

dsheets: Well, I've reordered it but GitHub is having some trouble with the force push.

dsheets:
Updated.

rgrinberg:
+1

PR #296, opened 2015-03-30
Remove mutability of request/response

rgrinberg:
Mutability of request/response really isn't buying anything over just using a ref when it's actually needed. I think we should just get rid of it unless we can justify it with an actual use case.

The purpose of this PR is to really just discuss the possibility and see
if/how many downstream projects rely on it.

### DIFF
--- a/lib/request.ml
+++ b/lib/request.ml
@@ -18,11 +18,11 @@
 open Sexplib.Std
 
 type t = {
-  mutable headers: Header.t;
-  mutable meth: Code.meth;
-  mutable uri: Uri.t;
-  mutable version: Code.version;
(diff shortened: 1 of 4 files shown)


rgrinberg:
Looks like this doesn't break anyone!

avsm:
I'm fine with this change.  Might as well get it in before 1.0

rgrinberg:
Yep. Merge it and push off 0.16.0? Multipart can wait until 0.17.0

avsm:
Yep, although I can't test the 0.16.0 cut as I'm on vacation in Singapore this week! Want to release it?

rgrinberg:
Sure!

PR #298, opened 2015-04-01
Fix cohttp-server-lwt bind

dsheets:
I also put an interface on `Cohttp_lwt_unix_net` and left doc comments noting it as private. I believe it is intended to be private but... maybe not?


samoht:
If it is really private, can you also update the `_oasis` file to move it into some `InternalModules`?

dsheets:
I don't think it actually is private any more. If you want to use `Cohttp_lwt.Make_client` with alternative `IO`, `Request`, or `Response` but not `Net` then it would be handy to have it around. I updated the branch but I'm still tracking down the 4.01.0 camlp4 syntax error.

dsheets:
times out on async_extra

PR #300, opened 2015-04-03
String_io: recognize CRLF is a line terminator

nojb:
The current code in `String_io` terminates lines only with `'\n'`.  This fixes it.


rgrinberg:
Thank you!

PR #301, opened 2015-04-03
Remove nettests

rgrinberg:
useless flag


avsm:
these are useful for offline development

rgrinberg:
The tests are still there but there's no use in having oasis run them

Issue #308, opened 2015-04-09
More robust handling of paths beginning with //


zoggy:
Hello,

When an uri such as `http://localhost//foo/bar` is requested, `Uri.of_string` is used on the requested path, i.e. `//foo/bar`, which returns an uri with path `/bar`, because it handles the whole string as a uri. This should be parsed as a path, to handle "incorrect" paths beginning with "//".

The call to `Uri.of_string` is here: https://github.com/mirage/ocaml-cohttp/blob/master/lib/request.ml#L113

dsheets:
Indeed you are correct. I am preparing a patch now.

dsheets:
#309 fixes this issue. When that PR is merged, this issue will close. Thanks for reporting the problem!

zoggy:
The fix does not seem to handle query params correctly, i.e. with the following url:
````
http://localhost/test/?formula=foo
````
the path obtained with `Uri.path` is `/test%3Fformula=foo` instead of "/test".

dsheets:
Indeed. :-( Sorry about that. It's the same problem as <https://github.com/mirage/irmin/pull/187#issuecomment-92703466>.

zoggy:
I'm trying to fix it.

zoggy:
Here is a PR: https://github.com/mirage/ocaml-cohttp/pull/317
"make test" was OK and it seems to fix my problem.

PR #309, opened 2015-04-09
Fix request path parsing

dsheets:
Addresses #308.

Perhaps we should parse auth headers here and reconstruct the actual request URI for the request record?

I tried to write some request parsing tests in  c429f5e but encountered what appears to be a long-standing compiler bug:

```
Fatal error: exception File "typing/printtyp.ml", line 583, characters 12-18: Assertion failed
```

which occurs with any of these `parse_request_uri_*` test cases on 4.01.0, 4.02.1, and 4.03.0+trunk. I'll report that to Mantis as soon as Travis fails.

This probably shouldn't be merged until we can confirm this bugfix works as expected and we can work-around the compiler bug.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+0.16.1 (trunk):
+New features and bug fixes:
+* Fix handling of request paths starting with multiple slashes (#308)
+
 0.16.0 (2015-03-23):
 
 Compatibility breaking interface changes:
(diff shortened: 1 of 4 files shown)


dsheets:
Reported in Mantis [PR6836](http://caml.inria.fr/mantis/view.php?id=6836).

dsheets:
I've reverted the addition of `short_paths` to `_tags` and the build (and new tests) now succeed except for the persistent ctypes 1.5.0 async_ssl failure.

rgrinberg:
There's tests. Yay.

:100:

PR #311, opened 2015-04-13
Link header

dsheets:
This begins Link header ([RFC 5988](https://tools.ietf.org/html/rfc5988)) and header i18n ([RFC 5987](https://tools.ietf.org/html/rfc5987)) support. I converted the `test_header` suite to use alcotest.

There are lots of potential issues with this support but I believe it is usable as-is. The parsing isn't perfect or particularly optimized, the i18n header parsing is slightly off, quote escaping probably doesn't work correctly, I'm not certain about the link relation type, the tests are probably more verbose than necessary, etc.

Feedback welcome. Support for RFC 5988 will be used in ocaml-github 1.0 to implement support for GitHub's pagination feature.


dsheets:
Build failure at <https://travis-ci.org/mirage/ocaml-cohttp/jobs/58307674#L2216> may be fixed by ocaml/ocaml-travisci-skeleton#37 as it appears to be a `js_of_ocaml` uninstall issue.

dsheets:
After ocaml/ocaml-travisci-skeleton#37 was merged, only the unknown cstruct 1.5.0 dll error remains.

dsheets:
I want to use alcotest for other cohttp test suites so I'm merging this. Please leave feedback here or open new issues if you would like to participate in discussion about this PR or feature.

lib/link.mli
```diff
@@ -0,0 +1,92 @@
...
+  | Successor_version
+  | Up
+  | Version_history
+  | Via
+  | Working_copy
+  | Working_copy_of
+
+type language = string
+
+type charset = string
```
rgrinberg: Are these aliases strictly necessary?

lib/link.mli
```diff
@@ -0,0 +1,92 @@
...
+  hreflang : string option;
+  media : string option;
+  title : string option;
+  title_ext : string ext option;
+  media_type : (string * string) option;
+  extensions : (string * string) list;
+  extension_exts : (string * string ext) list;
+}
+
+type t = { context : Uri.t; arc : arc; target : Uri.t; }
```
rgrinberg: Are we certain we want to expose these records? I guess I'm asking how stable do you expect the interface to stay @dsheets

rgrinberg:
+1

@dsheets could you add sexp converters please?

lib/link.mli
dsheets: Not necessary but both are structured values defined by RFCs that I don't want to implement right now. I will make them abstract and provide `to_string`/`of_string`.

lib/link.mli
dsheets: That is a good point. I could make these abstract as well and use fieldslib for the getters and setters?

dsheets:
Yep, I'll add the sexp converters with the abstractions.

PR #317, opened 2015-04-14
fixing parse of requests (#308)

zoggy:


lib/request.ml
```diff
@@ -108,13 +108,19 @@ module Make(IO : S.IO) = struct
...
       Header_IO.parse ic >>= fun headers ->
-      let empty = Uri.of_string "" in
+      let uri = Uri.of_string path in
+      let uri =
+        match Uri.scheme uri with
+          Some _ -> uri (* we have an absoluteURI *)
+        | None ->
+           let uri = Uri.of_string ("//localhost"^path) in
+           Uri.with_host uri None
+      in
```
avsm: This really ought to be in Uri itself eventually. Complex semantics :/

dsheets:
This fails against #318.

zoggy:
Which tests are failing ?

zoggy:
I think some tests should be fixed. I'll take the tests @dsheets added and commit some modifications.

rgrinberg:
There's not enough context here for me to understand what's going on. @dsheets I assume you'll take care of this.

zoggy:
I made some changes and added/fixed some tests. Now `Invalid "..."` is raised when the given path is not an absoluteURI and it does not begin with a '/'. (e.g. like in `GET foo HTTP/1.1`).

PR #318, opened 2015-04-14
Add more request URI tests for URI components

dsheets:
These are tests that I should have written before. Several fail (see #308).


dsheets:
This is still not [RFC 2616 Sec 5.1.2](http://tools.ietf.org/html/rfc2616#section-5.1.2) compliant. Absolute URIs are allowed in the `Request-URI` production for proxies and does not indicate a path-like-URI. `authority` alone and `*` are also allowed for special methods. Servers should return the right error codes when these failures occur.

Compliance may require larger refactoring.

dsheets:
I believe this is a reasonable implementation now.

dsheets:
I added `CONNECT` and `TRACE` HTTP 1.1 standard methods and then used them to manage the parsing of the `CONNECT authority` request line which is ambiguous otherwise. I also tested `OPTIONS *` and ensured that the requested URI's path component is never empty.

dsheets:
@samoht oops. I've changed the generate.ml script now. Thanks!

rgrinberg:
Sorry, it's too much to review closely. Are there any breaking changes?

dsheets:
I believe adding `CONNECT` and `TRACE` may be breaking if you rely on exhaustive method matches. Everything else should be correctness improvements.

dsheets:
I believe I have addressed any outstanding issues. `*` is now handled correctly, non absolute paths now result in ``` `Invalid``` values from the parser. Directory traversal is handled directly in the parser.

dsheets:
With no objections, I am merging this so we can release a version of cohttp that isn't horribly broken.

avsm:
+1

PR #319, opened 2015-04-15
Link header: add sexp converters, clean up interface

dsheets:
Moved some types into submodules and abstracted some (Language, Charset).
Added getters/setters.

The record fields are still exposed for pattern matching. I'm not sure about the `Rel.t` type. Should there be an `Other of string` constructor? Should it just be a `private variant` with a normalized string constructor and an `Extension of Uri.t` constructor?

### DIFF
--- a/lib/link.ml
+++ b/lib/link.ml
@@ -61,6 +61,47 @@ module Rel = struct
     | Working_copy
     | Working_copy_of
   with sexp
+
+  let extension uri = Extension uri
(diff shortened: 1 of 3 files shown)


rgrinberg:
If you're not sure you can always make the variant abstract and expose all the constructors as functions. I think we can omit `Other` until it's asked for.

dsheets:
I've abstracted `Link.Rel.t` into values now.

PR #320, opened 2015-04-15
Fix premature end of chunked body read

dsheets:
callv requests are explicitly POST; server responds to shutdown and then delays
100ms before exiting

### DIFF
--- a/lib/transfer_io.ml
+++ b/lib/transfer_io.ml
@@ -35,8 +35,15 @@ module Make(IO : S.IO) = struct
           try Some (int_of_string ("0x" ^ hex)) with _ -> None
         in
         match chunk_size with
-        |None | Some 0 -> return Done
-        |Some count -> begin
+        | None -> return Done
+        | Some 0 -> (* TODO: Trailer header support *)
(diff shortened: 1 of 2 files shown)


dsheets:
I've just committed a fix for missing `\r\n` consumption after the chunked encoding trailer fields. This corrects a defect that was causing premature end-of-pipeline events when using pipelined requests (e.g. `callv`). This defect is masked by the silent consumption of ``` `Invalid``` at <https://github.com/mirage/ocaml-cohttp/blob/master/lwt/cohttp_lwt.ml#L351>.

rgrinberg:
@dsheets looks good.

I'd also suggest to have a look at my integration testing initiative here https://github.com/mirage/ocaml-cohttp/blob/master/lib_test/test_sanity.ml

Tests done this way run automatically so they're more useful

dsheets:
Yes, that was very useful when I was debugging this issue. It needs to have a test case added for chunked + pipelined requests.

dsheets:
This got rolled into #321 so I'm closing this one. I'll add the test case to that PR.

PR #321, opened 2015-04-15
Fix our broken HEAD

### LINKED ISSUE #313: cohttp HEAD shouldn't use chunked mode
When you ask cohttp (0.16.0) to send a HEAD request by default it also sends a '0' as the body which confuses servers as they send back 400 reply with a body after the real reply.

Also if you call Cohttp_lwt_body.to_string on a HEAD request it just blocks indefinetely. Its probably ok to allow to check the length of the body for a HEAD (it might be as Content-Length in the headers) but it should probably raise an exception if you actually try to read the body of a HEAD reply.

[there are some broken servers though that do send a body for a HEAD, like some old version of nginx with gzip enabled: it sends the gzip/zlib headers in the body of the reply, that uncompresses to 0 bytes, not sure if it'd be worth supporting that]

```sh
$ cohttp-curl-lwt http://example.com -X HEAD
HTTP/1.0 400 Bad Request
Content-Type: text/html
Content-Length: 349
Connection: close
Date: Mon, 13 Apr 2015 16:07:05 GMT
Server: ECSF (ewr/15DC)

<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
	<head>
		<title>400 - Bad Request</title>
	</head>
	<body>
		<h1>400 - Bad Request</h1>
	</body>
</html>
```

### LINKED ISSUE #315: HEAD request file descriptor leak (waiting for body that never arrives)
When not using chunked requests (see https://github.com/mirage/ocaml-cohttp/issues/313) HEAD requests work, but the body is still created and it leaks a file descriptor.
I cannot use Cohttp_lwt_unix.drain_body because that would block indefinetely, and I don't see other functions to forcibly close the connection either.

dsheets:
`read_response` now requires a method and, if the method is ``` `HEAD```, no body
will be read. A method stream is produced from a request pipeline and this
result is now bound to the `Lwt` monad. A `last_body` reference has been introduced
which allows the response stream to finish drainage of any unhandled body.

Fixes #313. Fixes #315.

async implementation needs investigation.


dsheets:
I've added a sanity test case and expanded the command interface of `bin/cohttp_curl_async` to include a `-X METH` option. The async client code does not appear to suffer from the same bug as the Lwt implementation. I believe this PR is ready for merge.

lwt/cohttp_lwt.ml
```diff
@@ -194,21 +210,32 @@ module Make_client
...
-            Cohttp_lwt_body.write_body (Request.write_body writer) body) req oc)
-        reqs in
-    (* Read the responses. For each response, ensure that the previous response
-     * has consumed the body before continuing to the next response, since HTTP/1.1
-     * pipelining cannot be interleaved. *)
+    lwt meths = Lwt_stream.fold_s (fun (req,body) meths ->
+      Request.write (fun writer ->
+        Cohttp_lwt_body.write_body (Request.write_body writer) body
+      ) req oc >>= fun () ->
+      return ((Request.meth req)::meths)
```
rgrinberg: Why do you need to fold over requests to get the methods. Can't you just Lwt_stream.map_s and return the method one by one?

dsheets: I chose the fold to make sure the requests were actually written. I wasn't clear on the meaning of `Lwt_stream.map_s` because it seems like it simply records the mapping but doesn't force its application. The `fold_s` looked like the functional equivalent of the original `iter_s`.

Maybe this is OK but then a request may not be written to the wire until its response is trying to be read which sort of defeats the purpose of pipelining.

Basically, it comes down to ensuring the correct pipelining behavior. If there is a way to eagerly consume requests and lazily consume responses, we should use that. Perhaps we need some kind of "pump" lwt thread between the request stream and response stream to ensure that requests are dispatched ASAP and response consumption does not "pull" requests out to the wire?

rgrinberg: I'm also wondering if a stream is really the correct abstraction here for the input. We want something whose lifetime we can tie to the lifetime of our active connections but stream elements are produced lazily. We'll have to think more about that but this PR is a strict improvement.

PR #322, opened 2015-04-15
Respect HEAD in cohttp_server_{lwt,async}

### LINKED ISSUE #316: cohttp_server_{lwt,async} doesn't respond to HEAD correctly
`HEAD` is not `GET`!

dsheets:
I had to expose the representation of the `Cohttp_async.Server.response` type as a pair of response and body in order to write the `HEAD` postprocessor. I don't think this is a big deal but maybe there was a good reason that it was abstract.

I added 405 Method Not Allowed error functionality to the demonstration servers and `HEAD` support via response post-processing.

Fixes #316.

### DIFF
--- a/bin/cohttp_server_lwt.ml
+++ b/bin/cohttp_server_lwt.ml
@@ -25,6 +25,10 @@ open Cohttp_lwt_unix
 
 open Cohttp_server
 
+let method_filter meth (res,body) = match meth with
+  | `HEAD -> return (res,`Empty)
+  | _ -> return (res,body)
+
 let serve_file ~docroot ~uri =
   let fname = Server.resolve_local_file ~docroot ~uri in
   Server.respond_file ~fname ()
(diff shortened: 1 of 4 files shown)


bin/cohttp_server_lwt.ml
```diff
@@ -89,6 +81,30 @@ let handler ~info ~docroot ~verbose ~index (ch,conn) req body =
...
   )
 
+let handler ~info ~docroot ~verbose ~index (ch,conn) req body =
+  let uri = Cohttp.Request.uri req in
+  let path = Uri.path uri in
+  (* Log the request to the console *)
+  printf "%s %s %s %s\n%!"
+    (Cohttp.(Code.string_of_method (Request.meth req)))
+    path
+    (match verbose with
```
rgrinberg: What's this match statement for?

dsheets: I have no idea. It's been there since <https://github.com/dsheets/ocaml-cohttp/blob/00fbdbebe6c6f30c832615fd52b478c41e0e8710/bin/cohttp_server_lwt.ml> (first commit of this code) and should probably be cleaned up. I think there was some idea about verbose/non-verbose console logging but it never really materialized.

rgrinberg:
+1 waiting for travis.

PR #325, opened 2015-04-17
0.17.0

dsheets:
We should release this, I think.


dsheets:
Failing due to mirage/ocaml-conduit#58 and ocaml/opam-repository#3940.

rgrinberg:
Doesn't look like you've updated https://github.com/mirage/ocaml-cohttp/blob/master/lib/conf.ml

dsheets:
Good catch! Thanks! I regenerated oasis but didn't clean and re-configure it.

Issue #328, opened 2015-04-20
Latency benchmarking cohttp shows queueing


jlouis:
Hi,

So I'm in the process of doing latency-benchmarks for a couple of different web server frameworks. One of the frameworks is cohttp. I have the following code for the test:

```ocaml
(* This file is in the public domain *)

open Core.Std
open Async.Std
open Cohttp_async

(* given filename: hello_world.ml compile with:
   $ corebuild hello_world.native -pkg cohttp.async
*)

let handler ~body:_ _sock req =
  let uri = Cohttp.Request.uri req in
  match Uri.path uri with
  | "/" -> Server.respond_with_string "CHAPTER I. Down the Rabbit-Hole  Alice was beginning to get very tired of sitting by her sister on the bank, and of having nothing to do: once or twice she had peeped into the book her sister was reading, but it had no pictures or conversations in it, <and what is the use of a book,> thought Alice <without pictures or conversations?> So she was considering in her own mind (as well as she could, for the hot day made her feel very sleepy and stupid), whether the pleasure of making a daisy-chain would be worth the trouble of getting up and picking the daisies, when suddenly a White Rabbit with pink eyes ran close by her. There was nothing so very remarkable in that; nor did Alice think it so very much out of the way to hear the Rabbit say to itself, <Oh dear! Oh dear! I shall be late!> (when she thought it over afterwards, it occurred to her that she ought to have wondered at this, but at the time it all seemed quite natural); but when the Rabbit actually took a watch out of its waistcoat-pocket, and looked at it, and then hurried on, Alice started to her feet, for it flashed across her mind that she had never before seen a rabbit with either a waistcoat-pocket, or a watch to take out of it, and burning with curiosity, she ran across the field after it, and fortunately was just in time to see it pop down a large rabbit-hole under the hedge. In another moment down went Alice after it, never once considering how in the world she was to get out again. The rabbit-hole went straight on like a tunnel for some way, and then dipped suddenly down, so suddenly that Alice had not a moment to think about stopping herself before she found herself falling down a very deep well. Either the well was very deep, or she fell very slowly, for she had plenty of time as she went down to look about her and to wonder what was going to happen next. First, she tried to look down and make out what she was coming to, but it was too dark to see anything; then she looked at the sides of the well, and noticed that they were filled with cupboards......"
  | _ ->
    Server.respond_with_string ~code:`Not_found "Route not found"

let start_server port () =
  eprintf "Listening for HTTP on port %d\n" port;
[...]

PR #330, opened 2015-04-22
Limited read buffer size in Async backend to 32K.

artemkin:
There was unlimited buffer allocation in Async backend. It affected uploading of big files a lot. It was at least 10 times slower than downloading the same file. After the fix, uploading and downloading speeds seem to be on par.

Anyway, it is still not that good as new buffer is allocated for every single chunk. The buffer should be reused for all chunks for better performance.

### DIFF
--- a/async/cohttp_async_io.ml
+++ b/async/cohttp_async_io.ml
@@ -49,14 +49,19 @@ let read_line =
        |`Eof -> eprintf "<<<EOF\n"; None
     )
 
+let create_buf len =
+  let len' = min len 0x8000 in
+  String.create len', len'
+
 let read ic len =
-  let buf = String.create len in
+  let buf, len = create_buf len in
   Reader.read ic ~len buf >>| function
-  | `Ok len' -> String.sub buf 0 len'
+  | `Ok len' ->
+    if len' = len then buf else String.sub buf 0 len'
   | `Eof -> ""
 
 let read_exactly ic len =
-  let buf = String.create len in
+  let buf, len = create_buf len in
   Reader.really_read ic ~pos:0 ~len buf >>|
   function
   |`Ok -> Some buf


rgrinberg:
Yes the buffer allocation is very much not ideal. There's no "easy" way to fix it either. The whole IO layer should be reworked to stop allocating stuff with more efficient buffer management reimplemented on top of that.

Your fix is good however as we really should have a limit on the buffer size that we allocate. I'm merging this in.

Thanks.

artemkin:
Yes, I see. Maybe add it to #198 or report a separate issue? Performance is a feature for production systems. It seems people (including myself) wish to use `cohttp` for real world tasks (see #328).

I would contribute more, but the biggest problem for me is actually reading the codebase. A lot of functors + module openings just blow my mind. Any tips how to read that?

rgrinberg:
#198 only talks about interface breaking changes. I would hope that sane buffer management is not going to break people's code (but I've been wrong about this before).

I'll try and write something up on how to contribute/understand cohttp a little more. We do end up with a lot of complexity for trying to support multiple backends. But there's method to this madness.

avsm:
@artemkin I feel your pain.  One good way is to work from the `lib/` directory into `async/` and then into `lwt/`.  The Merlin IDE tool helps unwrap the types.  Documentation would most definitely help though.

trevorsummerssmith:
@avsm Getting off topic now but I think the high level design you wrote in this thread: http://lists.xenproject.org/archives/html/mirageos-devel/2015-03/msg00137.html would be very helpful to drop into the codebase basically verbatim.

artemkin:
@rgrinberg @avsm Could you release v0.17.1 with this fix? It is critical defect, and it is not convenient to use pinned repo. Guys who admin the server wish to build it locally, and they're not OCaml-ers yet)

avsm:
Yes; can do this afternoon.

avsm:
See mirage/ocaml-cohttp#332 and `DESIGN.md` skeleton committed, @trevorsummerssmith

trevorsummerssmith:
Thanks I think that will be a great help to new contributors!

artemkin:
Great! Thank you

vbmithr:
Please revert this. The semantics of read / read_exactly is not conserved anymore and this introduce bugs (Like #357).

PR #341, opened 2015-05-05
Specify content-length when request method is HEAD

### LINKED ISSUE #339: Hanging whenever having an empty body in Response
I'm encountering an issue where using, 0.17.1, `Cohttp_async.Body` on responses that have an empty body always just hangs indefinitely. 

I put some print statements in to_string around lines 101 of cohttp_async.ml but they don't even get executed, so I suppose the issue is happening even before that.

objmagic:
fix #339 

or is there any better way to solve this problem?

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -263,10 +263,13 @@ module Server = struct
         let keep_alive = Request.is_keep_alive req in
         let flush = Response.flush res in
         let res =
-          let headers = Cohttp.Header.add_unless_exists
-              (Cohttp.Response.headers res)
-              "connection"
-              (if keep_alive then "keep-alive" else "close") in
+          let headers = 
+            let hdrs = 
+              Cohttp.Header.add_unless_exists (Cohttp.Response.headers res)
+              "connection" (if keep_alive then "keep-alive" else "close") in
+            match res_body with
+            | `Empty -> Cohttp.Header.add hdrs "content-length" "0"
+            | _ -> hdrs in
           { res with Response.headers } in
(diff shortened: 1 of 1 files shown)


rgrinberg:
Thanks for the PR, I'm hesitant to merge because

1) The root of the issue is with the client. It should not simply block like it does.

2) Tests. I want to see this issue reproduced in a test case before accepting any potential "fixes"

marklrh:
agree. that's why I see it may not be a good answer. I need to dive deep into Client.request and I can do it on Friday.

as for the tests, I can put it in my [cohttp-test](https://github.com/marklrh/ocaml-cohttp-test) repo?

rgrinberg:
I'd suggest to add a test case here https://github.com/mirage/ocaml-cohttp/blob/master/lib_test/test_async_integration.ml#L23

rgrinberg:
Revisiting this, I just tried it with curl and it seems to get stuck as well. So it's indeed a problem with our server.

HOWEVER, 

1) Your commit message is confusing. You're specifying the content-length when the body's empty. Regardless of what the method is

2) After a quick look in the lwt server, I see no so similar workaround (perhaps @dsheets can illuminate). This leads me to believe that the workaround belongs in cohttp_async_server itself.

marklrh:
@rgrinberg 

I'll study carefully on Friday.

artemkin:
I use cohttp async, and it seems content-length is specified correctly. See this test case https://github.com/artemkin/git-lfs-server/blob/master/tests/test_head_metadata.ml#L14

I just respond with `\`Empty` body on server side (content-length is set to 0 automatically).

PR #344, opened 2015-05-07
[RDY] Quick n dirty attempt at async callv

rgrinberg:
That I've given up on because my test case is segfaulting on OSX

@trevorsummerssmith as we've discussed...

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -156,25 +156,40 @@ end
 
 module Client = struct
 
-  let request ?interrupt ?(body=`Empty) req =
-    (* Connect to the remote side *)
-    Net.connect_uri ?interrupt req.Request.uri
-    >>= fun (ic,oc) ->
-    Request.write (fun writer -> Body.write Request.write_body body writer) req oc
-    >>= fun () ->
-    Response.read ic
-    >>| function
+  let read_request ic =
+    Response.read ic >>| function
     | `Eof -> failwith "Connection closed by remote host"
     | `Invalid reason -> failwith reason
     | `Ok res ->
       (* Build a response pipe for the body *)
       let reader = Response.make_body_reader res ic in
       let rd = pipe_of_body (fun ic -> Response.read_body_chunk reader) ic in
+      (res, `Pipe rd)
+
+  let request ?interrupt ?(body=`Empty) req =
+    (* Connect to the remote side *)
+    Net.connect_uri ?interrupt req.Request.uri
+    >>= fun (ic,oc) ->
+      Request.write (fun writer -> Body.write Request.write_body body writer) req oc
+    >>= fun () ->
+    read_request ic >>| fun (resp, body) ->
       don't_wait_for (
-        Pipe.closed rd >>= fun () ->
+        body |> Body.to_pipe |> Pipe.closed >>= fun () ->
+          Deferred.all_ignore [Reader.close ic; Writer.close oc]);
+    (resp, body)
+
+  let callv ?interrupt uri reqs =
+    Net.connect_uri ?interrupt uri >>| fun (ic, oc) ->
+    reqs |> Pipe.iter ~f:(fun (req, body) ->
+      Request.write (fun writer -> Body.write Request.write_body body writer)
+    req oc) |> don't_wait_for;
+    let responses = Reader.read_all ic (fun ic ->
+      ic |> read_request >>| (fun x -> `Ok x)) in
+    don't_wait_for (
+      Pipe.closed responses >>= fun () ->
         Deferred.all_ignore [Reader.close ic; Writer.close oc]
-      );
-      res, `Pipe rd
+    );
+    responses
 
   let call ?interrupt ?headers ?(chunked=false) ?(body=`Empty) meth uri =
     (* Create a request, then make the request.
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -179,18 +179,6 @@ module Client = struct
       Deferred.all_ignore [Reader.close ic; Writer.close oc]);
     (resp, body)
 
-  (* Like Reader.read_all but allows you to control delay between reads *)
-  let read_all_delayed rd ~f =
-    let wait_for = ref Deferred.unit in
-    Reader.read_all rd (fun ic ->
-      !wait_for >>= fun () ->
-      f ic >>| function
-      | `Ok (result, wait_for') ->
-        wait_for := wait_for';
-        `Ok result
-      | `Eof -> `Eof
-    )
-
   let callv ?interrupt uri reqs =
     let reqs_c = ref 0 in
     let resp_c = ref 0 in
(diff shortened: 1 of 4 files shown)


marklrh:
I also tested on my OS X 10.10 and Ubuntu 14.04 (in VMWare)

For OS X 10.10
````bash
 $ ./test_async_integration.native 
2015-05-09 22:10:00.997878-07:00 Info Test Cohttp Server Test running on http://0.0.0.0:9193
Segmentation fault: 11
$ uname -a
Darwin warroom 14.3.0 Darwin Kernel Version 14.3.0: Mon Mar 23 11:59:05 PDT 2015; root:xnu-2782.20.48~5/RELEASE_X86_64 x86_64
````


For Ubuntu
````bash
$ ./test_async_integration.native 
2015-05-09 16:12:00.952429-07:00 Info Test Cohttp Server Test running on http://0.0.0.0:9193
Segmentation fault (core dumped)
$ uname -a
Linux ubuntu 3.13.0-32-generic #57-Ubuntu SMP Tue Jul 15 03:51:08 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
````

**Comment by rgrinberg** [comment]
@avsm any hints as to how to proceed here?

**Comment by rgrinberg** [comment]
Hmm I have a feeling that the segfault is occuring because I'm committing the same old mistake of not draining the body buffer before reading the next request.

Nevertheless, this shouldn't cause a segfault.

**Comment by dsheets** [comment]
The lwt implementation is now impervious to this issue. See <https://github.com/mirage/ocaml-cohttp/commit/c5f8f37b1ecdfb02aac37e91551089894786c6bb#diff-eae4e70a2c36d784c88a9039805fca28R224> for the fix. I'm not entirely happy with it for aesthetic reasons but I believe it works as intended.

**Comment by rgrinberg** [comment]
@trevorsummerssmith I've pushed out a set of commits that I think fixes the behavior. This implementation now passes my tests.

@dsheets I've adapted your body handling method to the async code. The only difference being that my criteria for a body being read is when writing to the body pipe has been finished rather than simply draining the body. What do you think about that?

I don't think we've found the ideal behavior here yet. Correct me if I'm wrong by a situation like this can give a really unintuitive result:

``` ocaml
let response_stream = Cohttp_lwt_unix.callv (* a few requests here *) in
response_stream >>= fun responses ->
responses |> Lwt_stream.to_list
```

will now return a list of responses with empty bodies.

dsheets:
Interesting. Has the pipe-read segfault been reported upstream?

If I understand correctly, the async and lwt versions now have the same body draining semantics after this pull request except the use of Async pipes means that you must wait for write completion before draining whereas Lwt streams will delay until stream end. Is that right?

Regarding your unintuitive example, I agree that this behavior is not ideal but it is an improvement over silent truncation of the response stream. In a perfect world, we could give the response stream a linear type and require the consumption of every element, in order, when traversing the stream. Unfortunately, we don't have that option available to us.

I see a few solutions:

1. Change the output type from `Cohttp_lwt_body.t` to `Cohttp_lwt_body.t option` to explicitly signal that the body was dropped rather than empty.
2. Change the output type from `Lwt_stream.t` to a special purpose stream-like construct that forces the user to traverse its cells manually.
3. Accept a function `Cohttp_lwt_body.t -> 'a` as argument to `callv` which processes the bodies and then return `(Response.t * 'a) Lwt_stream.t Lwt.t`
4. Give up on avoiding copies and create intermediate strings for every response body.

I think I'm in favor of 3.

SGrondin:
I'm in favor of 3 as well and I think that function should be a named argument.

rgrinberg:
> Has the pipe-read segfault been reported upstream?

It was not. I figured it was not worth it until I have a minimal test case.

> If I understand correctly, the async and lwt versions now have the same body draining semantics after this pull request except the use of Async pipes means that you must wait for write completion before draining whereas Lwt streams will delay until stream end. Is that right?

That's correct but I don't think the async implementation will drop any bodies like lwt's stream based implementation. However, it does mean that async implementation will block on reading the next request if the previous body pipe is too big to buffer the whole body (depends on the size budget).

Answering your proposals one by one:

1. This is does not seem satisfactory

2. I also think that Lwt_stream.t might not be the correct data structure here. One thing to ponder about streaming is whether we want to expose a "push" or "pull" based body. By pull based I mean something like we have currently with stream. While a push based api would require the user to provide some sort of chunk processing function that would be invoked. You could imagine that such a function could collect the chunks into a buffer and return a string to get back the simpler api.

3. Aaarg, I have similar thoughts that I've written up in 2.

4. I think that this is not so bad as the default behavior for small bodies actually. If we had some sort of request processing buffer (reusable or not, perhaps a cstruct) we'd imagine that a good portion of the time the body would simply fit into this buffer. I guess the challenge here is to have an API that makes both use cases convenient.

rgrinberg:
@marklrh would you mind giving the updated PR a quick test (and perhaps a review)

rgrinberg:
Note that this implementation is not yet ready. Currently, it will close the connection as soon as the response to the last request is obtained. This is unwanted because it might still be useful to reuse this connection and send new requests.

trevorsummerssmith:
Awesome work Rudi! It seems to me that reading the response bodies as strings is the best option for this first pass implementation. That works and has straightforward performance profile. Getting the trickier-optimized-for-large-bodies (ie constant memory in the user process) working should be left for another day.

marklrh:
the test passes on my machine. good work!

rgrinberg:
@trevorsummerssmith I think you can actually get that behavior rather easily with something like:

```
responses |> Pipe.map ~f:(fun (resp, body) -> (resp, Body.to_string body))
```

Do you think this is satisfactory for now?

It's not that I don't want to improve the behavior in the simple case, I'd just rather make this change in lockstep with Lwt which will take a little longer.

async/cohttp_async.ml
```diff
@@ -156,25 +156,60 @@ end
 
 module Client = struct
 
+  let read_request ic =
+    Response.read ic >>| function
+    | `Eof -> failwith "Connection closed by remote host"
+    | `Invalid reason -> failwith reason
```
avsm: Is the Async convention here to raise `Failure` or is there some other more descriptive exception?  I'm not sure what more idiomatic Async code does.

rgrinberg: I think that defining an `Invalid_response` exception would be fine.

But this PR doesn't introduce any kind of new error handling. This function was simply extracted out of request to be reusable in `callv`.

However, you remind of a good point. I think that both async's and lwt's callv shouldn't raise at all because it might be possible that raising on a bad response could possibly make it impossible for the user to retrieve the previous good responses (although the connection should be terminated regardless)

trevorsummerssmith: @rgrinberg Yes that's fine. Would your thought be that another PR would introduce a new error handling mechanism? As you point out the current error handling strategy is not very useful in the wild. It seems that one wants a ` (Response.t * Body.t) Or_error.t Pipe.Reader.t Deferred.t` instead which would allow one to figure out what is going on with each response.

rgrinberg: Yes @trevorsummerssmith I think that's a better return type. However the connection will probably be terminated after the first error. If the body is corrupted or unreadable somehow it probably doesn't make much sense to attempt reading the next body. So you will get at most one error.

Just making sure that's clear.

trevorsummerssmith: Yes that is clear, and seems to me to be the correct behavior. That allows one to process, and make decisions about retries.

rgrinberg: Actually, I don't think that my previous characterization of the issue is correct. The current implementation makes it impossible to read all responses that were successfully written to the pipe.

Some care will be necessary however because an exception might be thrown before the corresponding response is actually processed in the pipe. But you should be able to catch it and iterate over the pipe regardless.

rgrinberg:
@avsm @trevorsummerssmith I will merge this soon and leave out testing invalid responses for now. I don't believe the types will have to change even if there's a problem with the error handling in the future. But even if we must change the types, we'll have to change lwt's callv as well so it's not a big deal.

PR #346, opened 2015-05-16
Tests used polymorphic equality on URI

dsheets:
They now use `Uri.equal` which is guaranteed to have the correct semantics.

Also, I simplified the user/pass extraction from request URIs and converted the request tests to alcotest. Repeated percent-decoding is now tested.


rgrinberg:
+1, good to merge after CI passes

I think we also need `equal` on `Header.t` as well (but that's for a separate PR)

PR #348, opened 2015-05-17
fix typo

mkoconnor:
Just fix a typo

### DIFF
--- a/examples/async/receive_post.ml
+++ b/examples/async/receive_post.ml
@@ -21,7 +21,7 @@ let start_server port () =
 
 let () =
   Command.async_basic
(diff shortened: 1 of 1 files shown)


rgrinberg:
thanks

examples/async/receive_post.ml
```diff
@@ -21,7 +21,7 @@ let start_server port () =
 
 let () =
   Command.async_basic
-    ~summary:"Simple http server that ouputs body of POST's"
+    ~summary:"Simple http server that outputs body of POST's"
```
avsm: Another minor typo there is the apostrophe after `POST`, I think

PR #349, opened 2015-05-20
Remove Request/Response in functors

rgrinberg:
Exposing these modules on every functor application is not necessary.
The underlying {Request.Response}.t is always the same and there's no
need to expose the low level IO operations used.

@avsm @samoht Was there a particular reason these were exported? I looked for
a reason by trying to remove them and it looks like I've succeeded. The less
stuff we expose, the better right?

### DIFF
--- a/lwt/cohttp_lwt.ml
+++ b/lwt/cohttp_lwt.ml
@@ -56,8 +56,6 @@ end
 
 module type Client = sig
   module IO : IO
-  module Request : Request
-  module Response : Response
 
   type ctx with sexp_of
   val default_ctx : ctx
@@ -121,13 +119,11 @@ end
 
 module Make_client
     (IO:IO)
-    (Request:Request with module IO = IO)
-    (Response:Response with module IO = IO)
     (Net:Net with module IO = IO) = struct
 
   module IO = IO
-  module Request = Request
-  module Response = Response
+  module Response = Make_response(IO)
+  module Request = Make_request(IO)
 
   type ctx = Net.ctx with sexp_of
   let default_ctx = Net.default_ctx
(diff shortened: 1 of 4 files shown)


avsm:
It looks like when I exposed them in c71cb2b0a4b9d931bfb02a779e2d8c4b68e0a321, the `Request` module type didn't have the `Lwt.t` equivalence via `Http_IO`:

```
module type Request = sig
  include module type of Cohttp.Request with type t = Cohttp.Request.t
  include Cohttp.Request.S
end
```

It now reads:

```
module type Request = sig
  type t = Cohttp.Request.t with sexp
  include Cohttp.S.Request with type t := Cohttp.Request.t
  include Cohttp.S.Http_io with type t := Cohttp.Request.t
                            and type 'a IO.t = 'a Lwt.t
end
```

which is specific enough that there's no need to expose the module.  So I think this is safe, but it would be good to check that `mirage-http` doesn't break as a result (it's still fine if it does as that can be patched, but it's the usual source of breakage when these module types change).

rgrinberg:
I just tried `mirage-http` manually and it was ok.

I don't understand the travis failure.

dsheets:
`ocaml-github` expects `Cohttp_lwt_unix.Client.Response` to exist but I can change it to `Cohttp.Response` for 1.0.0.

rgrinberg:
@dsheets NP. We'll ping you once this is merged.

avsm:
Yeah, I'm ok with this change but prefer to delay it to 0.18.0 for some semblance of semantic versioning :)

rgrinberg:
There's no rush on this, I don't want to break github with the latest cohttp for people until 1.0 is out.

dsheets:
github 1.0 will be compatible with cohttp both before and after this change. I've just tested and committed a patch that works with this PR.

rgrinberg:
@dsheets can we merge this now?

dsheets:
No blockers from me.

samoht:
I think that's fine to remove these modules from the signature. That might cause `ocaml-git` to break but that's not a big deal, I'll fix that.

PR #350, opened 2015-05-20
Add "all" oasis flag

rgrinberg:
Makes it easier to test if everything compiles after:

```
$ ocaml setup.ml -configure --enable-all
$ ocaml setup.ml -all
```


rgrinberg:
Taking silence for approval

PR #351, opened 2015-05-20
add test out dir to gitignore

rgrinberg:


rgrinberg:
I'll wager and say that the gitignore change did not cause the travis failure here

Issue #357, opened 2015-05-29
Async client does not return the whole of a webpage's text.


vbmithr:
```ocaml
open Core.Std
open Async.Std
open Cohttp_async

let main () =
  Client.get (Uri.of_string Sys.argv.(1)) >>= fun (r, b) ->
  Body.to_string b >>= fun s ->
  Printf.printf "%s\n%!" s;
  Shutdown.shutdown 0;
  Deferred.unit

let _ =
  don't_wait_for @@ main ();
  never_returns @@ Scheduler.go ()
```

Try this on `https://api.bitfinex.com/v1/trades/BTCUSD`

rgrinberg:
You can also reproduce this with `$ ./cohttp_curl_async.native 'https://api.bitfinex.com/v1/trades/BTCUSD'

vbmithr:
Cool, thanks for the tip. But what about the bug :p

rgrinberg:
Seems like async's chunked encoding reader is buggy since `$ ./cohttp_curl_lwt.native` works`

vbmithr:
Yeah, I'm looking at this now.

vbmithr:
Works on 0.17.0, broken on 0.17.1, I'll soon figure out.

rgrinberg:
@vbmithr Awesome, thanks.

rgrinberg:
The only thing that's suspicious here is @artemkin 's commit

https://github.com/mirage/ocaml-cohttp/compare/v0.17.0...v0.17.1

vbmithr:
On 29/05/2015 17:07, Rudi Grinberg wrote:
> The only thing that's suspicious here is @artemkin
> <https://github.com/artemkin> 's commit
> 
> v0.17.0...v0.17.1
> <https://github.com/mirage/ocaml-cohttp/compare/v0.17.0...v0.17.1>
> 

The semantic of those functions are wrong, anyway. `len` is not respected.

-- 
Vincent

rgrinberg:
Yes that's a good point. I actually wonder if we should just remove that commit after https://github.com/mirage/ocaml-cohttp/pull/353 was done. It seems unnecessary.

vbmithr:
Reverted. And we need a new release for this :)

rgrinberg:
That's a little too fast...

I'd like to write a test for this first.

rgrinberg:
Also, are we sure there isn't a performance regression with respect to @artemkin 's original issue?

[...]

PR #358, opened 2015-05-29
Attempt to test #357

rgrinberg:
Large chunks in chunked encoding being truncated. This test attempts to
reproduce the behavior.

@vbmithr this is the after mentioned test. What's surprising is that it passed
even before your revert.


rgrinberg:
@vbmithr this test works (fails before your revert), it's just the chunk size here was too small.

PR #359, opened 2015-05-29
Patched Chunked.read so that it never allocates more than 32Kbytes.

vbmithr:
Please review. I did not write a test yet, but it works on a large page (https://api.bitfinex.com/v1/trades/BTCUSD).


rgrinberg:
Merged in #360

PR #363, opened 2015-06-01
Top level printers for the main cohttp types

### LINKED ISSUE #352: Add top level printers for things
* Headers
* Requests
* Responses
* Transfer encoding stuff (perhaps)

rgrinberg:
The main types are: `Header.t`, `Request.t`, `Response.t`.

This is basically a rip off of how @dsheets accomplished this in Uri. I did not
choose the prettiest representations but it's a good enough first start. Makes
interactive use much easier.

Fixes #352


avsm:
This is nice!

top/cohttp_top.ml
```diff
@@ -0,0 +1,20 @@
+let printers = [ "Cohttp.Header.pp_hum"
```
dbuenzli: There's a much simpler scheme to do this which is to write the installation bits in a source `PKG_top_init.ml` file and simply have a library `PKG_top` with a call to `Toploop.use_file`. The advantage is that it's easier to test that the init works before making a release if you wish, you get good error reporting and you don't `sprintf` code which is kind of ugly. 

You can have a look at [`gg_top`](https://github.com/dbuenzli/gg/blob/master/src/gg_top.ml) and [`gg_top_init.ml`](https://github.com/dbuenzli/gg/blob/master/src/gg_top_init.ml) or [`mtime_top`](https://github.com/dbuenzli/mtime/blob/master/src-os/mtime_top.ml) and [`mtime_top_init.ml`](https://github.com/dbuenzli/mtime/blob/master/src-os/mtime_top_init.ml).

dbuenzli:
Why don't you call the pretty printers simply `pp` ? It seems that when you do "one module per type" that's an established convention (or at least one that I use...). This would make the modules compatible with a hypothetical `PRINTABLE` signature.

rgrinberg:
I almost missed thes comments. Thanks @dbuenzli

I'll make a ticket for this stuff to keep track.

PR #370, opened 2015-06-04
Manually pack Cohttp module

rgrinberg:
* Prefixed every module with `Cohttp_`
* This seemed to expose some non packed usages in the xhr lib which i fixed
* The mli's in Cohttp.* now use the fully qualified which makes them a little
  uglier
* Move Cohttp_async.Body to own module


samoht:
Ideally cohttp.mli could also directly include cohttp_s with the full documentation. Then you can make all the `cohttp_*` modules private.

avsm:
Lets merge this for 0.19.0 along with the support for lwt 2.5.0. Doing a point release in the meanwhile

rgrinberg:
@avsm do you think it's worth it to go through the signature renaming? If we go 4.02+ we'll be able to avoid renaming by doing `open Cohttp` inside the pack with `-no-alias-deps`

avsm:
I think we should keep using pack for 1.0, and immediately apply this for 2.0 (which can be 4.02+ only).

rgrinberg:
OK Sure. I don't want to have to figure out how to use `-no-alias-dep` with oasis so I'd rather just switch to topkg before doing this.

avsm:
Yep, or Assemblage more likely.  Should speed up the build as well.

SGrondin:
Assemblage would be awesome, but it seems to be still in development and on some sort of hiatus? I know people look at popular libraries to see how their build works. Assemblage is very promising and the lack of use in popular libraries is probably slowing down its adoption and contributing to the horrible mess right now. From my experience trying to get some people into OCaml, the build process is the top reason people give up on the entire language!

PR #372, opened 2015-06-04
Fix whitespace handling and concatenation order in headers

edwintorok:
Cohttp concatenated header values in reverse order, with this patch we have the correct order.

### DIFF
--- a/lib/header_io.ml
+++ b/lib/header_io.ml
@@ -19,7 +19,7 @@
 
 let split_header str =
   match Stringext.split ~max:2 ~on:':' str with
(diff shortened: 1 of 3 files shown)


lib/header_io.ml
```diff
@@ -27,11 +27,13 @@ module Make(IO : S.IO) = struct
 
   module Transfer_IO = Transfer_io.Make(IO)
 
+  let rev _k v = List.rev v
```
rgrinberg: Did you forget to remove this?

rgrinberg:
:100:

PR #375, opened 2015-06-06
Fix 204 beavior

rgrinberg:
cc @edwintorok

### DIFF
--- a/lib/response.ml
+++ b/lib/response.ml
@@ -67,18 +67,26 @@ module Make(IO : S.IO) = struct
        let flush = false in
        return (`Ok { encoding; headers; version; status; flush })
 
-  let has_body {status; encoding} =
-    (* rfc7230#section-5.7.1 *)
-    match status with
-    | #Code.informational_status | `No_content | `Not_modified -> `No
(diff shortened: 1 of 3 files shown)


lib/response.ml
```diff
@@ -67,18 +67,26 @@ module Make(IO : S.IO) = struct
...
   let make_body_reader {encoding} ic = Transfer_IO.make_reader encoding ic
   let read_body_chunk = Transfer_IO.read
 
   let write_header res oc =
     write oc (Printf.sprintf "%s %s\r\n" (Code.string_of_version res.version)
       (Code.string_of_status res.status)) >>= fun () ->
-    let headers = Header.add_transfer_encoding res.headers res.encoding in
+    let headers =
+      if allowed_body res
+      then Header.add_transfer_encoding res.headers res.encoding
```
avsm: shouldn't this check to see if the response already has a Transfer-encoding header and not override it if so?

rgrinberg: Probably. I'll take a look at this separately since this PR doesn't change this aspect of the behavior

PR #379, opened 2015-06-28
Attempt to fix lwt's janky callv

rgrinberg:
@dsheets

@edwintorok 's implementation seems to work

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -206,32 +206,29 @@ module Make_client
   let callv ?(ctx=default_ctx) uri reqs =
     Net.connect_uri ~ctx uri >>= fun (conn, ic, oc) ->
     (* Serialise the requests out to the wire *)
-    Lwt_stream.fold_s (fun (req,body) meths ->
+    let meth_stream = Lwt_stream.map_s (fun (req,body) ->
       Request.write (fun writer ->
         Cohttp_lwt_body.write_body (Request.write_body writer) body
       ) req oc >>= fun () ->
-      return ((Request.meth req)::meths)
-    ) reqs [] >>= fun meths ->
+      return (Request.meth req)
+    ) reqs in
     (* Read the responses. For each response, ensure that the previous
        response has consumed the body before continuing to the next
        response because HTTP/1.1-pipelining cannot be interleaved. *)
-    let meth_stream = Lwt_stream.of_list (List.rev meths) in
     let read_m = Lwt_mutex.create () in
     let last_body = ref None in
-    let resps = Lwt_stream.from (fun () ->
-      let closefn () = Lwt_mutex.unlock read_m in
-      Lwt_stream.get meth_stream >>= function
-      | None -> return_none
-      | Some meth ->
-        begin match !last_body with None -> return_unit | Some body ->
-          Cohttp_lwt_body.drain_body body
-        end >>= fun () ->
-        Lwt_mutex.with_lock read_m (fun () -> read_response ~closefn ic oc meth)
-        >|= (fun ((_,body) as x) ->
-          last_body := Some body;
-          Some x
-        )
-    ) in
+    let closefn () = Lwt_mutex.unlock read_m in
+    let resps = Lwt_stream.map_s (fun meth ->
+      begin match !last_body with
+      | None -> return_unit
+      | Some body -> Cohttp_lwt_body.drain_body body
+      end >>= fun () ->
+      Lwt_mutex.with_lock read_m (fun () -> read_response ~closefn ic oc meth)
+      >|= (fun ((_,body) as x) ->
+        last_body := Some body;
+        x
+      )
+    ) meth_stream in
     Lwt_stream.on_terminate resps (fun () -> Net.close ic oc);
     return resps
 end
--- a/lib_test/test_sanity.ml
+++ b/lib_test/test_sanity.ml
@@ -20,7 +20,11 @@ let server =
     Server.respond ~status:`OK ~body:(Body.of_string_list chunk_body) ();
     Server.respond ~status:`OK ~body:(Body.of_string "") ();
     (* not modified *)
-    Server.respond ~status:`Not_modified ~body:Body.empty ()
+    Server.respond ~status:`Not_modified ~body:Body.empty ();
+    (* pipelined_interleave *)
+    Server.respond_string ~status:`OK ~body:"one" ();
+    Server.respond_string ~status:`OK ~body:"two" ();
+    Server.respond_string ~status:`OK ~body:"three" ();
   ]
   |> List.map const
   |> response_sequence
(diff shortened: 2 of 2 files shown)


edwintorok:
Doing pipelining correctly is actually more complicated than I initially thought, I have another WiP branch (but perhaps we can merge it after this one):
https://github.com/mirage/ocaml-cohttp/compare/master...edwintorok:lwt-callv-4?expand=1

* there should be  a limit on pipelining depth, you don't want to pipeline 1000 requests only to find out the server dropped the connection and you have to send them all again
* idempotent and non-idempotent requests can be mixed on the same persistent connection, but only idempotent requests should be pipelined according to the RFC (makes sense, connection might drop and you have to repeat the requests)
 * we can only pipeline on HTTP/1.1, but we only find out whether the server (and proxy) supports that after the first reply
* the read mutex seems to be unlocked too early, should be unlocked only after entire body is read/drained. Although map_s should ensure we don't attempt to read the body concurrently, so this point is moot
* all this makes the implementation (and usage) quite complicated, especially trying to match which reply goes with which request, your suggestion to change the API to supply a callback for parsing the body makes using `callv` a lot easier

lwt-core/cohttp_lwt.ml
```diff
@@ -206,32 +206,29 @@ module Make_client
...
       Request.write (fun writer ->
         Cohttp_lwt_body.write_body (Request.write_body writer) body
       ) req oc >>= fun () ->
-      return ((Request.meth req)::meths)
-    ) reqs [] >>= fun meths ->
+      return (Request.meth req)
+    ) reqs in
     (* Read the responses. For each response, ensure that the previous
        response has consumed the body before continuing to the next
        response because HTTP/1.1-pipelining cannot be interleaved. *)
```
dsheets: This comment seems a little wrong. Does it mean HTTP/1.1 pipelining cannot be *out-of-order*? It is called *pipelining*...

edwintorok: On 06/29/2015 04:08 PM, David Sheets wrote:
> In lwt-core/cohttp_lwt.ml <https://github.com/mirage/ocaml-cohttp/pull/379#discussion_r33461134>:
> 
>>        Request.write (fun writer ->
>>          Cohttp_lwt_body.write_body (Request.write_body writer) body
>>        ) req oc >>= fun () ->
>> -      return ((Request.meth req)::meths)
>> -    ) reqs [] >>= fun meths ->
>> +      return (Request.meth req)
>> +    ) reqs in
>>      (* Read the responses. For each response, ensure that the previous
>>         response has consumed the body before continuing to the next
>>         response because HTTP/1.1-pipelining cannot be interleaved. *)
> 
> This comment seems a little wrong. Does it mean HTTP/1.1 pipelining cannot be /out-of-order/? It is called /pipelining/...

The server is free to execute the requests in whatever order as long as it ensures that the replies are sent in the same order
as the requests: https://tools.ietf.org/html/rfc7230#page-54

There is nothing special in HTTP/1.1 pipelining that would allow clients to match responses with requests[1],
hence the pipelineing blacklists you see in tools like curl to deal with buggy servers that send replies out-of-order.

FWIW Firefox has pipelineing disabled by default, and chromium is about to remove it too: https://bugzilla.mozilla.org/show_bug.cgi?id=264354
Still I think it'd be useful for cohttp in situations where you know the server works correctly, in other situations you can still use persistence by specifying pipeline depth of 0.

[1] there was (a now expired) draft to add a header Assoc-Req: https://tools.ietf.org/html/rfc7230#page-54

SGrondin:
I'd like to make a suggestion here: HTTP/2 uses pipelining extensively and I think it could be a good use of time to read up on it and see if the HTTP pipelining @edwintorok is working on could be protocol-agnostic so that the code could be reused for HTTP/2. Thanks

avsm:
@SGrondin is spot on -- I think there are two layers here (which is why my original `callv` didnt do some things).

- A low-level pipelining layer that makes sure that Requests on a single connection are interleaved with Responses.  This is essentially like a ring buffer of RPCs.

- A higher-level API that can enforce a depth limit per connection, but also schedule a top-level request stream across multiple connections.  This logic could definitely be shared with HTTP2, which can do out-of-order pipelining as well.

There's increasing pressure on Cohttp to create the higher-level "just works" API that should look something like Python's Requests module. How much of the logic in this patch should go into that higher level API instead?

dsheets:
HTTP2's feature is called *multiplexing* and it may be able to be used via an HTTP 1.1-like high-level API but that cannot be the *only* interface to HTTP2 multiplexing. In particular, it seems like a bad idea to have different ordering semantics depending on which protocol version was negotiated. That is, the HTTP2 pipelining API should enforce HTTP 1.1 in-order return semantics. Additionally, many users of HTTP2 multiplexing will expect to be able to open multiple persistent multiplexed requests and keep them free from blocking each other. This is not achievable with HTTP 1.1 pipelining alone (you need multiple connections).

HTTP 1.1 and HTTP 2 capabilities are different enough that I believe both clients and servers should treat them as functionally equivalent transports for only the simplest of applications. In any application that deals with concurrent requests or is otherwise time or performance sensitive, the author should understand and specifically use the capabilities of each protocol. Only at a very high level (e.g. "download this list of URLs") can these transports be freely interchanged.

edwintorok:
@avsm: Most of the depth enforcement should go into a higher-level API,  in this sense `call` and `callv` are not low-level enough, perhaps low-level would be: establish connection, send request, read reply, shutdown connection (with the mutexes that you already have to enforce ordering).

One of the reasons why I like Cohttp is that it provides you access to the low-level and allows you to
compose your application the way you want to. I'm also using OCamlnet in another project, and although it does provide a high-level API for pipelining and scheduling over multiple connections I always stumble upon some hardcoded behaviour (e.g. 1 second reconnect pause in 3.7.x, fixed in 4.x).
If Cohttp provides a higher-level API I'd like to keep the ability to customize it, either via functors, or optional parameters to methods (that allow to customize not only specific values like timeouts, but the actual behaviour), i.e. if I want to make a small change in the default behaviour I should still be able to reuse the rest of the default behaviour that I didn't change.

I've  been experimenting with a `callp` built on top of `callv` that uses `Lwt_pool` (in fact multiple pools, one for HEAD/OPTIONS, one for idempotent and another for unsafe methods) that does part of the 'schedule a top-level request stream across multiple connections'. Problem is those decisions (how many pools to use, how to separate requests, how to schedule them, what timeout to use, how many times to retry) are application-specific and I wouldn't want to hardcode my specific choice in Cohttp:
```ocaml
val callp :
  ?headers:Cohttp.Header.t ->
  ?body:Cohttp_lwt_body.t ->
  Cohttp.Code.meth ->
  Uri.t -> (Cohttp.Response.t -> Cohttp_lwt_body.t -> 'a Lwt.t) -> 'a Lwt.t
```

rgrinberg:
This is a good discussion but I'm merging this. I'm pretty certain that this `callv` is a strict improvement over the old `callv` in an essential way (it actually allows you to reuse the connection).

@edwintorok I agree that changing the interface for `callv` could make implementation a lot easier but we need to mind backwards compatibility as well. I think at this point we're not going to break `callv`'s interface (hopefully when we finally switch to `4.02+` to use the deprecation annotation) but rather deprecate it if something better comes along.

edwintorok:
agreed on backwards-compat. Another question is how to implement these pipelining improvements or  higher-level API without having to reimplement it again for Async.
Could Async and Lwt based Cohttp's signatures be made close enough now that they could satisfy the same module signature?

PR #380, opened 2015-07-03
Fix compilation with lwt 2.5.0

vbmithr:


vbmithr:
To be merged after lwt 2.5.0 has made its way in the opam repository.

dsheets:
Can `buffer_size` remain so that `16384` isn't repeated?

vbmithr:
Were does this constant comes from btw ?

vbmithr:
Ah yeah, 4 pages.

avsm:
Yep, just a reasonable page multiple.

rgrinberg:
@vbmithr Should you also fix the constraints in `opam`?

mlin:
bump. `opam install lwt cohttp` is failing as of this morning...

https://github.com/ocaml/opam-repository/pull/4365
https://travis-ci.org/mlin/narcissome-beacon/builds/70516962

dsheets:
ocaml/opam-repository#4427

Sorry about that.

dsheets:
This PR also needs to update the cohttp opam pin file with the new constraint.

dsheets:
Should be fixed in opam-repository now. Thanks for your report!

rgrinberg:
Am I crazy to suggest using some preprocessing to make cohttp compatible with both older and newer lwt?

I really hate creating opam hell for everyone with this.

vbmithr:
What's the problem exactly ? Older version of cohttp will not work with newer lwt versions, is there more issues ?

rgrinberg:
The problem is we're making it difficult for people who depend on older versions of cohttp and newer versions of lwt and vice versa.

vbmithr:
On 16/07/2015 15:36, Rudi Grinberg wrote:
> The problem is we're making it difficult for people who depend on older
> versions of cohttp and newer versions of lwt and vice versa.

Personally I would not add cruft in our code in order to accomodate such
use cases.

It should be enough that for each version of cohttp, you need a certain
range of versions of lwt.

rgrinberg:
I agree it adds cruft but I wouldn't trivialize the use case. Giving our users a smooth upgrade path and keeping the range of compatibility wide for 2 very popular libraries on opam is useful and important IMO.

Putting it another way, w/e ugliness and time spent it would cost us to keep this compatibility, I'm sure it is dwarfed by the combined pain of our users.

I will merge this and amend the opam file however so at least users who want to use the latest lwt can use cohttp#master.

rgrinberg:
Applied. Thanks @vbmithr

PR #382, opened 2015-07-08
Add some REVDEPS

mor1:
Trying out using REVDEPS, as well as adding at least one more significant user of `cohttp`.


avsm:
`mirage-www` isnt actually a valid package in recent mirage releases -- this should probably be removed from opam rather than being made a gating dependency in cohttp

PR #383, opened 2015-07-11
Allow DELETE requests to have bodies

dsheets:
https://tools.ietf.org/html/rfc7231#section-4.3.5 paragraph 5 does not disallow bodies for `DELETE` requests. We should give the user the option of sending and receiving bodies with `DELETE` requests.

I believe this patch adds this functionality and does not break previous functionality. I'm not 100% certain that pipelined DELETE requests work with/without body in all circumstances.

### DIFF
--- a/lib/request.ml
+++ b/lib/request.ml
@@ -161,8 +161,8 @@ module Make(IO : S.IO) = struct
   (* Defined for method types in RFC7231 *)
   let has_body req =
     match req.meth with
-    | `GET | `HEAD | `DELETE | `CONNECT | `TRACE -> `No
-    | `POST | `PUT | `PATCH | `OPTIONS | `Other _ ->
+    | `GET | `HEAD | `CONNECT | `TRACE -> `No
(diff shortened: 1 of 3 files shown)


edwintorok:
Should the same apply to GET? I've only encountered this in [the ElasticSearch API](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-request-body.html) where its used to overcome URL length limitations with large queries. The wording in the RFC is similar to DELETE (i.e. no defined semantics and some implementations might reject it)

dsheets:
Probably.

samoht:
The issue with `GET` is that proxy are apparently often caching the body (http://stackoverflow.com/questions/978061/http-get-with-request-body) so it's a bit hard to predict how it will work.

samoht:
@dsheets need a version constraint in the `opam` file to not use the latest cohttp.

rgrinberg:
@dsheets Looks good. Can you fix this for async as well?

samoht:
I can try to propose a patch for async.

samoht:
See https://github.com/dsheets/ocaml-cohttp/pull/1

PR #389, opened 2015-07-14
Don't catch End_of_file for Lwt_io.read

fdopen:
Lwt_io.read doesn't raise End_of_file; it alread returns "".

https://github.com/ocsigen/lwt/blob/b4af03f0ea2b54b8f6470367bbe44153e7b3d89d/src/unix/lwt_io.mli#L256


rgrinberg:
Good catch.

PR #391, opened 2015-07-15
Improve a bit the documentation of Accept and add sorting.

Drup:
We could put the q-stuff in a submodule. Not sure if necessary.

### DIFF
--- a/lib/accept.mli
+++ b/lib/accept.mli
@@ -26,6 +26,11 @@ type q = int with sexp
 (** Lists, annotated with qualities. *)
 type 'a qlist = (q * 'a) list with sexp
 
+(** Sort by quality, biggest first.
+    Respect the initial ordering.
+*)
+val qsort : 'a qlist -> 'a qlist
+
 type pv = Accept_types.pv =
     T of string
   | S of string with sexp
--- a/lib/accept.ml
+++ b/lib/accept.ml
@@ -16,12 +16,18 @@
   PERFORMANCE OF THIS SOFTWARE.
 *)
 (* TODO: handle exceptions better *)
-(* TODO: sorting? *)
 
 open Printf
 include Accept_types
 module Parser = Accept_parser
 module Lexer = Accept_lexer
+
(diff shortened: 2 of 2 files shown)


rgrinberg:
* I'd merge the docs improvement immediately if it was in a different commit

* The implementation in this PR obviously looks good but since I know next to nothing about this "quality" stuff. Therefore I will leave this to @dsheets

lib/accept.mli
```diff
@@ -43,13 +56,14 @@ type encoding =
   | Identity
   | AnyEncoding with sexp
 
+(** Language tags, as a list of subtags.
+    ["en-gb"] is represented as [Language ["en"; "gb"]].
+    @see <http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4> The spec.
+*)
 type language = Accept_types.language =
```
dbuenzli: Note that this is formally a BCP47 *basic language range* tag and not a *language tag* (FWIW I do have a module [here](https://github.com/dbuenzli/remat/blob/master/src-common/nlang.mli) to handle these that I expect to release as an independent module at some point).  Also referring to https://tools.ietf.org/html/rfc7231#section-5.3.5 would be nicer, the latter obsoletes 2616 and makes a lot of things more clear notably in this case it links to the right RFC.

Drup: I know about BCP47, I was also considering doing such a module, happy to see you did it already.

Thanks for the link. It's updated. Is there a guideline in mirage libraries on how to reference RFCs and so on ? I guess it's waiting for ppx and [reqtrace](https://github.com/infidel/reqtrace) ?

Drup:
@rgrinberg done.

rgrinberg:
@Drup applied first patch, thanks.

dsheets:
LGTM.

rgrinberg:
@Drup please rebase

Drup:
Already did 2 minutes ago :D

PR #392, opened 2015-07-16
Small tweaks to the lwt backend

rgrinberg:
I hope this code wasn't inlined on purpose :D

### DIFF
--- a/async/cohttp_async_io.ml
+++ b/async/cohttp_async_io.ml
@@ -31,9 +31,6 @@ type ic = Reader.t
 type oc = Writer.t
 type conn = unit
 
-let iter fn x =
-  Deferred.List.iter x ~f:fn
-
 let read_line =
   check_debug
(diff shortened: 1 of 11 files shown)


rgrinberg:
The last commit can hypothetically break compatibility for some.

rgrinberg:
False alarm from travis.

rgrinberg:
The last commit also break compatibility in "theory" because IO is not part of our stable interface.

rgrinberg:
FYI @samoht @avsm 

* IO.iter will be gone
* Client.IO will be gone

avsm:
mainly need to check that mirage-http will be ok

rgrinberg:
@avsm I checked mirage-http and it's ok. Don't remove iter from mirage-http yet though (keep it compatible with older cohttp's as well)

avsm:
lgtm!

PR #395, opened 2015-07-19
Surround source copyright with fold markers

rgrinberg:
Vim and emacs can fold these by default and just display the first line
of the copyright.

learned this nice trick from the merlin source


rgrinberg:
If anyone finds this offensive or legally questionable feel free to revert.

Issue #396, opened 2015-07-19
Remove all IO related functions in Response/Request


rgrinberg:
Both the lwt and async backend expose the application of `{Response,Request}.Make` inside their respective `{Respones,Request}` modules. I don't think the functions contained in those are particularly useful to the user an they are definitely subject to change.

This will break user code that rely on this stuff but they can always work around it by manually apply `{Request,Response}.Make` (like our tests for example)

PR #397, opened 2015-07-19
Move all lwt signatures to Cohttp_lwt_s

rgrinberg:
To avoid duplicating the signatures in Cohttp_lwt.ml{,i} like `Cohttp`


rgrinberg:
This is likely to need mirage-http fixed but I'd prefer to address that along with #396

samoht:
Need to check that `ocaml-git+http` still compiles as well

rgrinberg:
@samoht I've added some aliases to preserve compatibility.

PR #401, opened 2015-07-26
Remove {Request,Response}.Make applications from interface

### LINKED ISSUE #396: Remove all IO related functions in Response/Request

rgrinberg:
Fix #396


rgrinberg:
`read_form` is really incorrect (the url decoding should be done to the Body.t) and from the looks of it isn't being used either [1]. I've removed it.

https://github.com/search?q=language%3AOCaml+read_form&type=Code&utf8=%E2%9C%93

seliopou:
This should really be in the interface. I've had to copy the form parsing code over to another project in the past, I've had people ask about how to do (I direct them to this PR), and I have another project where I'm going to have to copy the code over again.

rgrinberg:
What code are you copy pasting?

My problem isn't with the functionality that's being exposed but rather how it's being exposed. However, this revision isn't even hiding anything, it's just making it slightly more difficult to use.

PR #402, opened 2015-07-27
Do not return stacktrace in body on 500's

rgrinberg:
We need to think of another way to expose this kind of desired behavior when debugging but until then we should definitely err on the side of caution.

Also, respond_file no longer catches exceptions because they should just
be caught by the general 500 handler (mentioned above).

Any tips on how to test the respond_file change?

### DIFF
--- a/lib_test/test_sanity.ml
+++ b/lib_test/test_sanity.ml
@@ -27,6 +27,11 @@ let server =
     Server.respond_string ~status:`OK ~body:"three" ();
   ]
   |> List.map const
+  |> (fun tests ->
+    tests @ [
+      (fun _ body -> (* Returns 500 on bad file *)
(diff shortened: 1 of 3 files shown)


edwintorok:
How about using `Lwt_log.ign_warning ~exn` to log the exception? By default it'll go to stderr (if `lwt.unix` is linked, otherwise to null), but `Lwt_log.default` can be configured to log to null, a channel, a file, or syslog. The stacktrace will be logged only if `Printexc.backtrace_status` is on.

For testing the `respond_file` change can you try to remove read permission on the file? `|EACCES` doesn't seem to be caught explicitly.

rgrinberg:
Agreed, but we also need to change logging in `Cohttp_lwt_unix_io` then. I'll open up a separate ticket for that.

Good idea on the tests. Added a commit for that.

PR #406, opened 2015-07-29
Get rid extra functions Cohttp.S.Http_io

rgrinberg:
write_header make_body_writer write_footer

### DIFF
--- a/lib/s.mli
+++ b/lib/s.mli
@@ -74,10 +74,7 @@ module type Http_io = sig
   val make_body_reader : t -> IO.ic -> reader
   val read_body_chunk : reader -> Transfer.chunk IO.t
 
-  val write_header : t -> IO.oc -> unit IO.t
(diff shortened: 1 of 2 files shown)


samoht:
How are we supposed to do this now? [ocaml-git](https://github.com/mirage/ocaml-git/blob/master/lib/http/git_http.ml#L177) uses these function to encode the Git's smart HTTP protocol (basically an RPC over HTTP layer).

rgrinberg:
I have a sizable simplification coming up this week as to how cohttp handles request/response IO so it should actually make some things easier. So this particular bit will actually be easy to port in ocaml-git.

I'm a little curious about this part of ocaml-git though. It seems like you're reimplementing a bunch of cohttp inside git_http.ml. Is there anything we can do so that you can reuse cohttp better in ocaml-git?

In any case, if it doesn't work for ocaml-git we will definitely revert this.

Issue #242, opened 2015-01-26
IO.read should not gobble errors silently


avsm:
IO.read currently returns an option, with no ability to comment on the sort of error it's returning.  This is a fairly big revision, but it needs to move to a model where errors and end-of-files are distinguished.

See #207

Chris00:
What is the preferred scheme for error handling?  Using something like [Or_error](https://blogs.janestreet.com/how-to-fail-introducing-or-error-dot-t/), using `Lwt.fail` (nor really portable to async), polymorphic variants,…?

talex5:
Currently undecided, but I'm collecting comments and proposals here: https://github.com/mirage/mirage-www/pull/274

(also, there's lots of discussion on the mailing list at present)

avsm:
Gobbling errors like is is really, really harmful in larger scale uses of Cohttp.  Running https://github.com/avsm/opam-mirror is causing undebuggable failures when running a large number of threads.  This must be fixed before a 1.0 despite the interface breaking change I think /cc @rgrinberg

SGrondin:
Agreed. I'm running a busy cohttp server and managing its crashes and errors is a pain.

rgrinberg:
@avsm I'm +1231241 on this change but why do you consider this a breaking change? I don't consider `S.IO` part of our stable API.

@SGrondin Can you please describe in a bit more detail. If it's graphic enough, it would encourage me to crack my knuckles and finally tackle this.

yomimono:
`src/s.ml` now has a `module type Http_io` with

```
val read : IO.ic -> [ `Eof | `Invalid of string | `Ok of t ] IO.t
```

 , which seems to fulfill this request, so I'll close this issue.  If this is incomplete or incorrect, please feel free to reopen.

PR #410, opened 2015-07-30
Redo IO layer [WIP]

rgrinberg:
tl;dr: Separate IO from parsing.

The goals are:

* Make things like #396 easy
* Remove code duplication that we had in the previous `*_io` modules.
* Testing the parsers should be much easier than it currently is
* Address #242


avsm:
I like this, but it really will break existing consumers of Cohttp badly. Should we just cut a 1.0 from the 0.19.x base and dive straight into a Cohttp 2.0 with this?

rgrinberg:
Why should it break existing consumers badly? This is barely breaking mirage-http...

At this point the only thing it will break is ocaml-git which should be trivially fixable.

avsm:
> Why should it break existing consumers badly? This is barely breaking mirage-http

Was thinking of the fd-passing byte-by-byte reading version of Cohttp that @djs55 have in XenAPI. It might survive this reasonably actually...

The size limit is now the `string` size limit, isn't it? (just did a quick scan of the patch, will read properly tomorrow)

rgrinberg:
Yeah the size limit is 16k by default. The goal is to have some sort of rudimentary ddos protection from overly massive headers.

samoht:
> At this point the only thing it will break is ocaml-git which should be trivially fixable.

I'm happy to fix it if we have a nicer cohttp API ;-)

rgrinberg:
Cleaned up the commit history a little to make it easier to read.

PR #413, opened 2015-08-02
Test large headers and a large number of headers

avsm:


rgrinberg:
Need to bump the version in conf.ml as well.

PR #414, opened 2015-08-04
Add cancel API for Lwt unix.

toots:
I was writing code to handle HTTP audio streams so I needed a `abort` API. This is a quick and dirty implementation. Would you guys be interested by a clean one? If so, how do you think it should/could be done?

### DIFF
--- a/bin/cohttp_curl_lwt.ml
+++ b/bin/cohttp_curl_lwt.ml
@@ -26,7 +26,7 @@ let client uri ofile meth' =
   debug (fun d -> d "Client with URI %s\n" (Uri.to_string uri));
   let meth = Cohttp.Code.method_of_string meth' in
   debug (fun d -> d "Client %s issued\n" meth');
-  Client.call meth uri >>= fun (resp, body) ->
+  Client.call meth uri >>= fun (resp, body, _) ->
   let status = Response.status resp in
   debug (fun d ->
     d "Client %s returned: %s\n" meth' (Code.string_of_status status)
--- a/bin/cohttp_proxy_lwt.ml
+++ b/bin/cohttp_proxy_lwt.ml
@@ -40,7 +40,7 @@ let handler ~verbose (ch,conn) req body =
   in
   (* Fetch the remote URI *)
   let meth = Request.meth req in
-  Client.call ~headers ~body meth uri >>= fun (resp, body) ->
+  Client.call ~headers ~body meth uri >>= fun (resp, body, _) ->
   if verbose then
     eprintf "<-- %s %s\n%!"
       (Uri.to_string (Request.uri req))
--- a/examples/doc/client_lwt.ml
+++ b/examples/doc/client_lwt.ml
@@ -3,7 +3,7 @@ open Cohttp
 open Cohttp_lwt_unix
 
 let body =
-  Client.get (Uri.of_string "http://www.reddit.com/") >>= fun (resp, body) ->
+  Client.get (Uri.of_string "http://www.reddit.com/") >>= fun (resp, body, _) ->
   let code = resp |> Response.status |> Code.code_of_status in
   Printf.printf "Response code: %d\n" code;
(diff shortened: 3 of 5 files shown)


rgrinberg:
I've never actually tried this myself, but does `Lwt.cancel` not do what you want?

Haven't looked at your code however yet.

avsm:
Same question as @rgrinberg: the convention in Lwt is to raise `Lwt.Canceled` which ripples up through the cancelled threads until caught.  Is there a case where `Lwt.cancel` fails to cleanly shut down a Cohttp thread?  It's quite likely that something could go wrong since this is a very untested path (in most cooperatively threaded code sadly, not just Cohttp/Lwt)

toots:
Okay, I've done more testings. The problem with `Lwt.cancel` is that it indeed cancels the thread but none of the handlers for `on_terminate` are being called, which will result in `fd` leaks and all sort of confusions.. Will see if I can do another, cleaner pass with this in mind..

rgrinberg:
You're probably aware of this, but perhaps `Lwt.on_cancel` can be make sure cleanup occurs.

Which reminds me that we need to make cleanup exception safe as well.

toots:
Yes but for this to work, you need to be able to cancel a thread below the `body` stream. As far as I can see, canceling my processing thread does not propagate back to the stream, thus the stream's `fd`s aren't cleanup either..

I'm talking about this piece of code: https://github.com/mirage/ocaml-cohttp/blob/master/lwt-core/cohttp_lwt.ml#L69
I'm aware that the `Gc` should do the job eventually but I've never been quite keen on leaving `fd` close up to the `Gc`..

avsm:
Good point. The interface absolutely should not depend on the Gc cleaning up the file descriptors.

hcarty:
I'm think I'm running into this as well, with a Cohttp server ending up with thousands of `ESTABLISHED` connections to nowhere.  I'll try to come up with a minimal, reproducible test case next week.

avsm:
A small repro case would be appreciated.  The GC shouldn't have to clean up fds to prevent a leak indeed.

hcarty:
Sooner than I was hoping.  This seems to show the problem.  `lsof -i` shows the open connections.  I used cohttp.lwt + lwt.ppx.

```ocaml
open Cohttp
open Cohttp_lwt_unix

let server =
  let callback conn_id req body =
    Lwt_unix.sleep 0.05>>
    Server.respond_string ~status:`OK ~body:"helloworld" ()
  in
  let%lwt ctx = Conduit_lwt_unix.init ~src:"127.0.0.1" () in
  let ctx = Cohttp_lwt_unix_net.init ~ctx () in
  let mode = `TCP (`Port 8080) in
  let config = Server.make ~callback () in
  Server.create ~ctx ~mode config

let client =
  for%lwt _i = 1 to 50 do
    Lwt_io.printlf "%d" _i>>
    let getter =
      let%lwt _, body = Client.get @@ Uri.of_string "http://127.0.0.1:8080/" in
      let%lwt _ = Cohttp_lwt_body.to_string body in
      Lwt.return_unit
    in
    Lwt.pick [getter; Lwt_unix.sleep 0.01]
  done

let%lwt () =
  Lwt.join [server; client]
```

hcarty:
For slightly easier testing I created a fork with a ready to build/test binary: https://github.com/hcarty/ocaml-cohttp/tree/lwt-fd-leak

hcarty:
I removed my previous comments because I'm not sure the problem I'm seeing is related.

I have an example here which seems to never free client connections:
https://github.com/hcarty/ocaml-cohttp/blob/lwt-fd-leak/bin/cohttp_lwt_leak.ml

and a potential fix/patch here:
https://github.com/hcarty/ocaml-cohttp/commit/c4136c41cf90bf9b3e04cfa51568ef323201b974

If the patch looks appropriate I'll submit a pull request.

mseri:
This has become out of date, feel free to send a new PR if needed.

PR #419, opened 2015-08-10
Preserve header case WIP

rgrinberg:
Use a case ignoring comparison function rather than lowecasing header
keys in header map.

@SGrondin does this work for you?

@avsm Is this approach ok?

This is still WIP b/c LString remains useless and we shouldn't make
is_header_with_list_value inefficient like this.

### DIFF
--- a/lib/header.ml
+++ b/lib/header.ml
@@ -16,16 +16,30 @@
  *
   }}}*)
 
+exception Found_int of int
+
+let compare_caseless x y =
+  if String.(length x <> length y)
+  then String.compare x y
+  else
+    try
+      for i = 0 to String.length x - 1 do
+        if Char.(lowercase x.[i] <> lowercase y.[i]) then
+          raise (Found_int (Char.compare x.[i] y.[i]))
+      done;
+      0
+    with Found_int i -> i
+
 module LString : sig
-  type t
+  type t = string
   val of_string: string -> t
   val to_string: t -> string
   val compare: t -> t -> int
 end = struct
   type t = string
-  let of_string x = String.lowercase x
+  let of_string x = x
   let to_string x = x
-  let compare a b = String.compare a b
+  let compare a b = compare_caseless a b
 end
 
 module StringMap = Map.Make(LString)
@@ -45,7 +59,7 @@ let headers_with_list_values = Array.map LString.of_string [|
 let is_header_with_list_value =
   let tbl = Hashtbl.create (Array.length headers_with_list_values) in
   headers_with_list_values |> Array.iter (fun h -> Hashtbl.add tbl h ());
-  fun h -> Hashtbl.mem tbl h
+  fun h -> Hashtbl.mem tbl (String.lowercase h)
 
 let init () =
   StringMap.empty
--- a/lib_test/test_header.ml
+++ b/lib_test/test_header.ml
@@ -99,14 +99,14 @@ let large_header () =
   let sbuf = Cohttp.String_io.open_in ibuf in
   let header_printer h =
     Printf.sprintf "[length %d]%s"
-     (List.length (H.to_list h)) 
-     (String.concat "\n" (List.map (fun (k,v) -> Printf.sprintf "%s: %s" k v) (H.to_list h))) in 
+     (List.length (H.to_list h))
+     (String.concat "\n" (List.map (fun (k,v) -> Printf.sprintf "%s: %s" k v) (H.to_list h))) in
   assert_equal ~cmp:(fun a b -> H.compare a b = 0) ~printer:header_printer (HIO.parse sbuf) h
 
 let many_headers () =
   let size = 1000000 in
   let rec add_header num h =
-    match num with 
+    match num with
     | 0 -> h
     | n ->
        let k = sprintf "h%d" n in
@@ -117,7 +117,7 @@ let many_headers () =
   let h = add_header size (H.init ()) in
   assert_equal ~printer:string_of_int
     (List.length (H.to_list h)) size
-    
+
 module Content_range = struct
   let h1 = H.of_list [("Content-Length", "123")]
   let h2 = H.of_list [("Content-Range", "bytes 200-300/1000")]
--- a/lib_test/test_parser.ml
+++ b/lib_test/test_parser.ml
@@ -246,7 +246,7 @@ let write_req expected req =
 let make_simple_req () =
   let open Cohttp in
   let open Cohttp_lwt_unix in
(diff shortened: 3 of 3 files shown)


lib/header.ml
```diff
@@ -16,16 +16,30 @@
...
+exception Found_int of int
+
+let compare_caseless x y =
+  if String.(length x <> length y)
+  then String.compare x y
+  else
+    try
+      for i = 0 to String.length x - 1 do
+        if Char.(lowercase x.[i] <> lowercase y.[i]) then
+          raise (Found_int (Char.compare x.[i] y.[i]))
```
avsm: I don't think we should be leaking a new exception here...

avsm: This makes me wonder if we should just use one of the unicode libraries for the header comparison

rgrinberg: What do you mean leaking? I don't think a user can ever encounter this exception.

rgrinberg:
Are headers allowed to be unicode? I thought it was ascii only - need to check.

lib/header.ml
dbuenzli: > This makes me wonder if we should just use one of the unicode libraries for the header comparison

Certainly not, you are working on the ASCII set, no need to spend time decoding UTF-8.

dsheets:
Why?

lib/header.ml
```diff
@@ -16,16 +16,30 @@
...
+      done;
+      0
+    with Found_int i -> i
```
avsm: Sorry, this bit scrolled out of scope on my phone :) Ignore the previous "dont leak exception comment"

avsm:
RFC5987 , need to check browser adoption

dsheets:
I believe 5987 only applies to field *values*. [7230](http://tools.ietf.org/html/rfc7230#appendix-B) ABNF says header field *keys* are tokens which are sub-ASCII.

rgrinberg:
@dsheets you mean why do we need this change at all?

@SGrondin has a cohttp based proxy and his users are complaining that the header case isn't being preserved.

lib/header.ml
```diff
@@ -16,16 +16,30 @@
  *
   }}}*)
 
+exception Found_int of int
+
+let compare_caseless x y =
+  if String.(length x <> length y)
+  then String.compare x y
```
dsheets: This causes a weird ordering.

rgrinberg: Agreed. I should probably just get rid of that early check since it's just a premature optimization.

dsheets:
Are his users complaining for aesthetic or functional reasons? Is something broken due to lowercasing? What?

My gut reaction is that this change dramatically increases the difficulty of correctly using `Header.iter`, `Header.map`, and `Header.fold` as a previous invariant (matchable lowercase header names) is now broken. Without more information about what *specifically* is wrong with lowercasing header names, this seems like a bad trade-off. Perhaps functions which return header names on which to act programmatically could lowercase the names before returning? It is a possibility but increases complexity and confusion. Seeing as the spec makes header name case irrelevant, I don't see why users need to be exposed to this.

If there are problems with compatibility or similar in the proxy case, I'd love to hear about them and perhaps we can work out a design which satisfies everyone. As it stands, I think this change makes common, spec-obeying systems harder to write correctly.

SGrondin:
Technically, header keys are case insensitive. In practice, they are not. There are countless bash, js, php, etc, scripts out there that rely on it. It really sucks, but every week I have to close tickets opened by our users that turned out to be related to this. They complain that their headers disappear after going through our proxy. They didn't, their code just didn't check in a case insensitive way.

I don't want to impose something relatively minor like this on cohttp if it breaks other stuff left and right.

I'm not a fan of the approach in this PR, personally. I think a better way to do it would be to extend the Header type. Instead of the key being just a string, I think it should be `string * string`. The first is the lowercase string that will continue to be used everywhere. The second is the original casing as it was set when the Header was created, it is only used when writing to the socket.

Thoughts?

rgrinberg:
> The first is the lowercase string that will continue to be used everywhere. The second is the original casing as it was set when the Header was created, it is only used when writing to the socket.

It seems strange that you see this as an improvement over the PR. As to me your suggestion only looks like an optimization saving the costs of calling `String.lowercase` where the header key should be used.

SGrondin:
There's a couple reasons why I think it's an improvement.

- Quoting @dsheets "My gut reaction is that this change dramatically increases the difficulty of correctly using Header.iter, Header.map, and Header.fold as a previous invariant (matchable lowercase header names) is now broken."
- It's a safe refactoring. All we have to do is change the type and the compiler will tell us where to change it. There's virtually no risk of introducing new bugs.
- It involves zero new logic.
- It's more functional
- I'm not a fan of using exceptions for flow control, but I can see why in this case.
- I guess I just find it personally cleaner.

That's just my opinion, I'm sure you have plenty of reasons why you chose that approach. I'm glad you're looking into the issue I mentioned on Gitter and I thought you might find my suggestion helpful :+1:

dsheets:
I understand more why this functionality is necessary, thanks Simon!

I think we should have some kind of support for this for compatibility's sake. Perhaps have the header value contain the actual value *and* the original string and offer high fidelity accessors? The API consumer would never see a type change but new functions would be added to the interface to opt into this information.

On the unreasonable, utopian side of things, it seems that the current behavior is useful as it has helped your users find bugs in their/other software! :stuck_out_tongue:

avsm:
Preserving the original string and storing it alongside the normalised one is somewhat memory expensive, but quite simple compared to the other solutions.

One thing we could do for `Header.t` in general is to maintain a global hashcons of the string values, as Xenstore does with path names.  For a high bandwidth server, the majority of the header values will just be integer references into the global header name cache.

rgrinberg:
> Preserving the original string and storing it alongside the normalised one is somewhat memory expensive, but quite simple compared to the other solutions.

I agree with this and this is my main concern. Why should spec compliant users have to pay this cost?

> One thing we could do for Header.t in general is to maintain a global hashcons of the string values, as Xenstore does with path names. For a high bandwidth server, the majority of the header values will just be integer references into the global header name cache.

Agreed. This is a good optimization. Although storing the original header key seems to kill its usefulness.

IMO, if we go down the route of storing both the lowercase header + original header (although I still don't understand how that's better than my PR + lowercasing keys when passing them to map/fold/iter/etc.) then we should look for a way to make it optional because I can't justify punishing spec compliant users.

samoht:
I'm not sure to understand why keeping both strings is more expansive. https://github.com/mirage/ocaml-cohttp/blob/master/lib/header.ml#L26 already does a copy anyway... Just modify `LString` to keep a couple and nothing else will have to change.

avsm:
That code discards the original string. I meant more expensive in terms of memory pressure (which usually dominates HTTP connection bottlenecks over GC pressure, particularly for long lived connections)

PR #428, opened 2015-09-24
Support Async 113.00 by explicitly using the blocking Core printf

avsm:


avsm:
CI failures in cow are in mirage/ocaml-cow#84

PR #439, opened 2015-10-28
Small improvements to cohttp_server_async.ml

rgrinberg:
* Don't call `determine_mode` twice
* Log errors

### DIFF
--- a/bin/cohttp_server_async.ml
+++ b/bin/cohttp_server_async.ml
@@ -117,13 +117,15 @@ let start_server docroot port host index verbose cert_file key_file () =
   Unix.Inet_addr.of_string_or_getbyname host
   >>= fun host ->
   let listen_on = Tcp.Where_to_listen.create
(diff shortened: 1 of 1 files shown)


bin/cohttp_server_async.ml
```diff
@@ -117,13 +117,15 @@ let start_server docroot port host index verbose cert_file key_file () =
...
-      ~listening_on:(fun _ -> port)
+                    ~socket_type:Socket.Type.tcp
+                    ~address:(`Inet (host, port))
+                    ~listening_on:(fun _ -> port)
   in
   Server.create
-    ~on_handler_error:`Ignore
-    ~mode:(determine_mode cert_file key_file)
+    ~on_handler_error:(`Call (fun addr exn ->
+      Log.Global.error "Error from %s" (Socket.Address.to_string addr);
```
avsm: Is there a monitor-local Log in Async? That seems like the right place to dispatch a handler error.

PR #446, opened 2015-12-01
Fix descriptor leak on thread cancellation for the Lwt client

hcarty:
I've been testing this patch on an internal project and haven't seen any leaks since it was applied.

Possibly related to https://github.com/mirage/ocaml-cohttp/pull/414.

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -75,6 +75,10 @@ module Make_client
           closefn ();
           return (res, `Empty)
       end
+  let read_response ~closefn ic oc meth =
+    let t = read_response ~closefn ic oc meth in
+    Lwt.on_termination t closefn;
+    t
 
   let is_meth_chunked = function
     | `HEAD -> false


rgrinberg:
I'm convinced that such a change should go in.

However, what about adding this logic inside `read_response` itself?

Also, perhaps it makes sense to use `Lwt.{on_termination,on_failure}` rather than just `Lwt.on_cancel`?

hcarty:
The change could use `Lwt.on_termination` is already used [here](https://github.com/mirage/ocaml-cohttp/blob/master/lwt-core/cohttp_lwt.ml#L158) so that makes sense as an alternative.

Is it safe to call `closefn` multiple times? If so then it could reduce some code duplication to add `Lwt.on_failure` inside `read_response`.

rgrinberg:
I believe that `Lwt.on_termination` [1] only executes after the stream has been exhausted. You can cancel/raise without exhausting the stream which would prevent the termination callback from executing. I should verify this though.

I think `closefn` is idempotent, and if it's not we should consider making it like that.

[1] https://github.com/rgrinberg/lwt/blob/master/src/core/lwt_stream.mli#L227-L230

hcarty:
I've updated the change to wrap `read_response` instead.

rgrinberg:
Merged in #447

PR #447, opened 2015-12-13
Fix descriptor leak on thread cancellation for the Lwt client

rgrinberg:
@hcarty small tweak to your revision

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -52,7 +52,7 @@ module Make_client
   let default_ctx = Net.default_ctx
 
   let read_response ~closefn ic oc meth =
-    Response.read ic >>= function
+    Response.read ic >>= begin function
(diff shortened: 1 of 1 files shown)


hcarty:
Makes sense, thanks!

edwintorok:
I get `Lwt_io.Channel_closed("input")` when reading the body of a response now, reverting this commit makes the exception go away.
It would seem it calls closefn too eagerly now, before the body was fully read.

hcarty:
Ah @edwintorok I see why `Lwt.on_termination` is a problem now.

Maybe both `Lwt.on_cancel` and `Lwt.on_failure` should be called instead of `Lwt.on_termination`?  I can submit a new PR with that change.

hcarty:
Using `on_cancel` and `on_failure` seems to appease both test cases.  I'll submit a new PR.

Issue #448, opened 2015-12-15
proxying of chunked response broken after c863c670713588074081cf8a556d115ddf77e489


edwintorok:
Testcase for https://github.com/mirage/ocaml-cohttp/pull/447#issuecomment-164690639
```
$ ./cohttp_proxy_lwt.native &
$ http_proxy=127.0.0.1:8080 curl http://jigsaw.w3.org/HTTP/ChunkedScript
```

Before c863c670713588074081cf8a556d115ddf77e489 curl finishes, after c863c670713588074081cf8a556d115ddf77e489 curl hangs.

PR #449, opened 2015-12-15
Fix issue #448 after breakage from PR #447

hcarty:
Don't close ic/oc immediately as they may still be in use when reading body content.

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -75,7 +75,8 @@ module Make_client
       end
     end
     |> fun t ->
-    Lwt.on_termination t closefn;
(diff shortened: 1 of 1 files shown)


rgrinberg:
Looks good. Can we somehow incorporate @edwintorok 's report into a test? I would love it if we could test for this regression in the future.

hcarty:
Would you prefer a test that calls out to `curl` or see if the issue can be reproduced with Cohttp's client?

edwintorok:
Reproducing the issue with Cohttp's client and server would be useful.

rgrinberg:
Thanks @hcarty tests are in #451 Using cohttp's curl.

PR #455, opened 2016-01-24
Log uncaught exceptions from user callbacks

talex5:
Before, the actual error was thrown away, making debugging difficult.

I've run into this a few times myself, and @mattgray reminded me of the problem last week.

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -275,7 +275,9 @@ module Make_server(IO:IO) = struct
         (fun () ->
            Lwt.catch
              (fun () -> callback (io_id, conn_id) req body)
-             (fun _exn -> respond_error ~body:"Internal Server Error" ()))
+             (fun exn ->
+               Format.eprintf "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn);
+               respond_error ~body:"Internal Server Error" ()))
         (fun () -> Body.drain_body body)
     ) req_stream
 
--- a/opam
+++ b/opam
@@ -29,8 +29,8 @@ depends: [
   "cmdliner" {build & >= "0.9.4"}
   "re"
   "uri" {>= "1.9.0"}
-  "fieldslib" {>= "109.20.00"}
-  "sexplib" {>= "109.53.00"}
+  "fieldslib" {>= "109.20.00" & < "113.01.00"}
+  "sexplib" {>= "109.53.00" & < "113.01.00"}
   "conduit" {>= "0.7.0"}
   "stringext"
   "base64" {>= "2.0.0"}


edwintorok:
More logging is definetely good, although I would've preferred to see `Logs` (or `Lwt_log` if you don't want to add dependencies) used here.
Whats the agreed-upon logging for cohttp/conduit?

talex5:
I thought of using `Logs`, but it suppresses all logging by default, so doesn't help with beginners not seeing errors. `Lwt_log` could work (haven't used it before). Any objections to using that?

edwintorok:
Not an objection but I'd like to see whether Logs's default could be improved before discarding the idea. E.g. it would seem preferable to me to have Logs's default reporter print warnings/errors to stderr rather than loose those messages completely (perhaps together with a warning/error that logging itself isn't properly set up to guide beginners).

/cc @dbuenzli

dbuenzli:
This has been discussed numerous times and at that point is simply trolling from @talex5 who is the only one so far that sees a problem with this. I think that it is the application and not the libraries that needs to decide what it wants to do with its standard outputs. 

In practice libraries randomly vomiting on the standard outputs of your application without you necessarily being aware of is more annoying than seeing that nothing is being reported and ask yourself the natural question on how you could have messages be reported. In the worst case you won't realize that you have a library that writes to your standard outputs before being in production and it could for example corrupt a well-defined format your application is writing on a stream that happen to have been specified as one of its standard outputs. 

I'm more in favour of writing robust programs than helping beginners here. If you want to help beginners do actually take time to write some documentation...

rgrinberg:
@talex5 Could you comment on the fieldslib constraints change?

Otherwise I'm not opposed to including this although it's clearly not ideal. I don't have any preference towards logs or lwt_log. But if Logs will work well with async I'm in favor of it.

talex5:
@rgrinberg the fieldslib thing was just to work around the build failing with:

         $ make
        ./setup.bin -configure --enable-lwt  --enable-lwt-unix   --prefix /usr/local
        ocamlfind: Package `fieldslib.syntax' not found
        W: Field 'pkg_fieldslib_syntax' is not set: Command ''/home/user/.opam/4.02.3/bin/ocamlfind' query -format %d fieldslib.syntax > '/tmp/oasis-5a4e1e.txt'' terminated with error code 2
        E: Cannot find findlib package fieldslib.syntax
        E: Failure("1 configuration error")
        Makefile:17: recipe for target 'setup.data' failed
        make: *** [setup.data] Error 1

I don't want to get into another Logs discussion. I generally like it and have converted several libraries to use it, but it doesn't help here, as I explained above. Given that even OCaml itself writes to stderr by default on error, I don't see a problem with doing this.

rgrinberg:
Thanks @talex5 this is cherry picked!

PR #465, opened 2016-03-25
Switch to pa_*_conv packages

rgrinberg:
Like https://github.com/mirage/ocaml-conduit/pull/124

cc @dsheets @avsm


rgrinberg:
Looks like oasis is making this impossible somehow. Passing in `pa_{fields,sex}_conv` also doesn't work for some reason. Oasis doesn't make ocamlbuild enable `-syntax camlp4o`

rgrinberg:
@edwintorok @Drup Any idea why oasis is acting up here?

edwintorok:
On lack of `-syntax camlp4o` from the oasis manual: "automatic inclusion of -syntax camlp4o for packages that matches the internal heuristic (if a dependency ends with a .syntax or is a well known syntax)."
However adding a package with `.syntax` still doesn't work:
```
ocamlopt.opt: don't know what to do with /home/edwin/.opam/async/lib/pa_fields_conv/-ignore.
```

rgrinberg:
Yup. And you can't add `true: syntax(camlp4o)` to the `_tags` file either. Stuck in a bind here.

PR #467, opened 2016-03-25
fix leak on head client requests

edwintorok:
I had this bugfix on a branch of mine (the branch itself is not yet ready for integration), but this bugfix is useful on its own.

otherwise Client requests fail with a resolver error or time out due
to running out of file descriptors

test with ./test_net_lwt_client_and_server.native

### DIFF
--- a/lib_test/test_sanity.ml
+++ b/lib_test/test_sanity.ml
@@ -10,6 +10,8 @@ let message = "Hello sanity!"
 
 let chunk_body = ["one"; ""; " "; "bar"; ""]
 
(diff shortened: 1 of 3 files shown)


rgrinberg:
Would it be possible to convert this to a `test_sanity` test? This way it will run automatically.

edwintorok:
I'll look into it

edwintorok:
Updated to use `test_sanity`, and fixed `cohttp_lwt_unix_test` to exit when the test server fails.

rgrinberg:
:100:

PR #470, opened 2016-03-29
Make Uri "optional"

rgrinberg:
This is my personal fork that allows you to set request paths directly as a
string. The main motivation being that I need a special way of encoding query
params to access amazon cloudsearch.

Not all of these commits are relevant but it's a good sketch of where I'm
taking this

@avsm @dsheets What do you think?

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -20,9 +20,7 @@ PKG re
 PKG re.emacs
 PKG stringext
 PKG fieldslib
-PKG pa_fields_conv
 PKG sexplib
-PKG pa_sexp_conv
 PKG ipaddr
 PKG ipaddr.unix
 PKG conduit
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -165,13 +165,13 @@ module Client = struct
       let pipe = pipe_of_body Response.read_body_chunk reader in
       (res, pipe)
 
-  let request ?interrupt ?ssl_config ?target ?(body=`Empty) req =
+  let request ?interrupt ?ssl_config ?host ?(body=`Empty) req =
     (* Connect to the remote side *)
-    let target =
-      match target with
+    let host =
+      match host with
       | Some t -> t
       | None -> Request.uri req in
-    Net.connect_uri ?interrupt target
+    Net.connect_uri ?interrupt host
     >>= fun (ic,oc) ->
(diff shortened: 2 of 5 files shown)


avsm:
I am fine with this change. However I'm curious: when faced with a need to fork due to Uri not doing what you needed, why did you choose to modify Cohttp instead of forking Uri to match your requirements?  I am fine with Cohttp being more decoupled from Uri, if @dsheets is.

dsheets:
It's fine with me. Uri needs major interface update and clean-up. I suspect there is 50 hours of work in it. I would be willing to merge first-class character-encoding-set support into 1.x which is sitting in mirage/ocaml-uri#80 right now.

rgrinberg:
@avsm Because I was still using Uri's traditional uri encoding when accessing other web services. Yes I could have implemented something like `Uri.cloudsearch_encode` but then how would I tell cohttp which encoding scheme (when converting `Uri.t -> string`) to use depending on the service?

Therefore it seemed like it was much easier to set this at the request level where I can explicitly specify how to encode the uri.

avsm:
Makes perfect sense, thanks!

rgrinberg:
I've trimmed some fat from revision and also made `target` optional. This should mean that there will be almost no breakage for users apart from those who used the `uri` field label directly.

There's no way to specify `target` for the lwt client. This is because there's no `request` like function in the lwt client.

I also don't quite like the `target` parameter. Thinking of renaming it to `host`. Any suggestions?

rgrinberg:
The optional parameter is now named as `?host`. This isn't final, so alternative names are still welcome.

dsheets:
How about `uri` or `url` for the optional parameter name and `resource` for the field name? The optional parameter can contain more than just the host and the field can contain both path and query string.

rgrinberg:
I have a feeling that the REST inspired `resource` will not make things clearer to many users but this is still much better than my own scheme so I'll adopt it.

dsheets:
`endpoint`? `fetch`? `retrieve`?

rgrinberg:
Resource beats those. I've glanced over the RFC again and it does use the word "resource" to describe the request uri so it's probably OK.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1

PR #473, opened 2016-04-05
oasis: Update version number.

j0sh:
Currently opam reports 20.1, while findlib reports 19.3. This fixes findlib, which allows downstream packages to constrain cohttp to a ppx-supported version.


rgrinberg:
Whoops. Sorry for the blunder.

I'm not sure what you mean about downstream package constraints though. Constraints use opam versions no?

rgrinberg:
Actually I will just release 0.20.2 With an updated findlib version. Replacing 0.20.1 will cause pain for people until an updated version is accepted.

j0sh:
Thanks for the quick response. In this case, I had also set a cohttp version constraint in an oasis file. That failing is how I tracked down this problem.

PR #475, opened 2016-04-05
Add ocamlbuild as build dependency

rgrinberg:
Not very useful given the long list of dependencies that surely include
ocamlbuild already but good practice


dbuenzli:
Do not confuse good practice and correctness. If all the members of the long list of dependencies would switch to another build system you would have an incorrect description.

PR #484, opened 2016-05-22
Update oasis/useragent version to 0.21.0

rgrinberg:


lib/conf.ml
```diff
@@ -15,5 +15,5 @@
  *
  *)
 
-let user_agent = "ocaml-cohttp/0.20.2"
-let version = "0.20.2"
+let user_agent = "ocaml-cohttp/0.21.0"
+let version = "0.21.0"
```
dsheets: Seems odd that `user_agent` doesn't use `version` here...

rgrinberg: Agreed. I will fix this.

PR #486, opened 2016-06-01
Fix up the ChangeLog markdown to be more compliant

avsm:
This improves support for dsheets/ocaml-changes parsing


dsheets:
I think it's not quite the markdown we want, yet, because releases aren't marked as headers with `#` or similar but it's close. I'm also rather annoyed by file extensions but I guess GitHub demands them to do nice rendering...

dbuenzli:
@dsheets I vaguely remember that you can get the good rendering by using `.gitattributes` but then it's one more file to add and to maintain.

avsm:
Yeah the file extension is annoying but not a huge deal.  The layout of the Markdown is more aggravating to get the trial-and-error right...

PR #488, opened 2016-06-07
Expose [Reader.t] and [Writer.t] via [`Switching_protocols] for websockets

msaffer:
This allows CoHTTP to hand off the raw reader and writer to a handler.

### DIFF
--- a/cohttp-async/src/server.ml
+++ b/cohttp-async/src/server.ml
@@ -17,9 +17,22 @@ type ('address, 'listening_on) t = {
 
 type response = Response.t * Body.t [@@deriving sexp_of]
 
+type response_action =
+  [ `Expert of Cohttp.Response.t * (Io.ic -> Io.oc -> unit Deferred.t)
+  | `Response of response
+  ]
+
+type 'r respond_t =
+  ?flush      : bool
+  -> ?headers : Cohttp.Header.t
+  -> ?body    : Body.t
+  -> Cohttp.Code.status_code
+  -> 'r Deferred.t
+
 let close t = Tcp.Server.close t.server
 let close_finished t = Tcp.Server.close_finished t.server
 let is_closed t = Tcp.Server.is_closed t.server
+let listening_on t = Tcp.Server.listening_on t.server
 
 let read_body req rd =
   match Request.has_body req with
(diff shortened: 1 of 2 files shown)


seliopou:
It may be worth considering exposing the underlying socket rather than the reader/writers. The constructors for `Reader.t` and `Writer.t`'s take a lot of arguments that one may want to tweak for performance purposes, not the least of which is the buffer size.

msaffer:
That's an interesting point @seliopou, I haven't needed to modify anything with this patch to get reasonable performance out of expert mode.

I could thread [Tcp.Server.create_socket] through [Conduit_async.serve], but since I haven't run into that, I figured this was the minimal change.

msaffer:
I'm fairly certain the continuous integration failure is unrelated to my change as well, should I be basing this off of v0.21.0 explicitly?

msaffer:
I see now that the build is busted because of ocaml/opam-repository#6699. Will resubmit after that goes in.

msaffer:
I've rebased the changes on top of master.

msaffer:
Hmm. The build is failing with errors that do not appear to be related to my change, e.g.

> =-=- Synchronising pinned packages =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
>[cohttp-async: local] Command started
>[cohttp-async] /repo/ already up-to-date
>The following dependencies couldn't be met:
>  - cohttp-async -> cohttp -> ppx_type_conv >= v0.9.1
>Your request can't be satisfied:
>  - ppx_type_conv>=v0.9.1 is not available because your system doesn't comply with ocaml-version >= "4.04.1".
>No solution found, exiting
>'unset TESTS; OPAMYES=1 export OPAMYES; set -ue; opam install cohttp-async --deps-only' >exited 3. Terminating with 3

Any guidance on how to make the build pass would be appreciated.

avsm:
Hm, #591 may have removed support for OCaml 4.03.0 by mistake.  We may need to add a manual conflict for one version rather than just a lower bound. That's unrelated to this PR, sorry about that :)

msaffer:
Hey @avsm it's been almost 2 weeks, any idea how long it'll take to get this fixed? This change is blocking my ability to open source some other code.

rgrinberg:
@msaffer sorry for the long delay, I don't think there's a point to wait any longer as nothing better is going to come up.

msaffer:
Thanks all!

avsm:
Sorry about the delay @msaffer -- I'm cutting a release now.

PR #492, opened 2016-06-11
"modern" oasis

rgrinberg:
Generates better ocamlbuild boilerplate


dinosaure:
May be it's more useful to use `oasis` with the `dynamic mode` (like [`Syndic`](https://github.com/Cumulus/Syndic/issues/34)) ?

rgrinberg:
Interesting. How does it work with pinning then? In that case, oasis would  be a dependency of cohttp in development but not of any releases.

I'm OK with moving towards that but we would have to tweak the release process for this.

dinosaure:
Yes, if you pin `cohttp` we need to add the `oasis` dependency in `opam`. @Drup explain a good process to release in the issue (with a shell script) and you can remove the `oasis` dependency just before the release in the opam repository.

PR #477, opened 2016-05-01
Fd safety for Cohttp_async.Client.request

rgrinberg:
If `request` raises, we should still take care to close ic and oc. We
don't close the fd's unconditionally at the end of request using
something like Monitor.protect because the success case should delay
closing the fd's until the body pipe is closed.

Fix #444

PR #506, opened 2016-09-17
Fd safety for Cohttp_async.Client.request

vbmithr:
If `request` raises, we should still take care to close ic and oc. We
don't close the fd's unconditionally at the end of request using
something like Monitor.protect because the success case should delay
closing the fd's until the body pipe is closed.

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -167,47 +167,59 @@ module Client = struct
 
   let request ?interrupt ?ssl_config ?uri ?(body=`Empty) req =
     (* Connect to the remote side *)
-    let uri =
-      match uri with
+    let uri = match uri with
       | Some t -> t
-      | None -> Request.uri req in
-    Net.connect_uri ?interrupt uri
-    >>= fun (ic,oc) ->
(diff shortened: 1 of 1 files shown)


vbmithr:
@rgrinberg 

I've rebased #477 on master and tweaked it a bit.
Could you please review?

rgrinberg:
I'd like to revert the change you've made to my original diff, the one where you're using `Deferred.all_unit` instead of `don't_wait_for`. Reason being that closing itself can raise and then mask the original exception. This is probably unwanted.

mseri:
Already modified and merged as https://github.com/mirage/ocaml-cohttp/pull/523/files

mseri:
Already modified and merged as https://github.com/mirage/ocaml-cohttp/pull/523/files

mseri:
Already modified and merged as https://github.com/mirage/ocaml-cohttp/pull/523/files

mseri:
Already modified and merged as https://github.com/mirage/ocaml-cohttp/pull/523/files

mseri:
Already modified and merged as https://github.com/mirage/ocaml-cohttp/pull/523/files

PR #508, opened 2016-10-03
use bigsubstring

vbmithr:


avsm:
This prevents one extra copy -- is that right?

@seliopou has a much more efficient rewrite in the works, but this ltgm to merge since it doesnt affect interfaces.

vbmithr:
On 03/10/2016 15:04, Anil Madhavapeddy wrote:
> This prevents one extra copy -- is that right?

Yes, one extra copy.

-- 
Vincent

rgrinberg:
Is it really worth it to add this global contention on 1 buffer just to save 1 allocation? Seems like a benchmark would be useful to convince us.

@seliopou do you think it will help at all?

vbmithr:
On 04/10/2016 02:57, Rudi Grinberg wrote:
> Is it really worth it to add this global contention on 1 buffer just to
> save 1 allocation? Seems like a benchmark would be useful to convince us.
> 
> @seliopou <https://github.com/seliopou> do you think it will help at all?
> 

`read` should take a buffer as argument. I don't know what optimizations
@selioupou did but if it superseede this patch then I can wait :) I was
on a quest to understand why my programs were eating tons on RAM. Not
sure this is a big piece of the puzzle anyway.

-- 
Vincent

avsm:
@seliopou's patches are a fairly epic performance improvement, according to his OCaml 2016 talk...

vbmithr:
On 04/10/2016 11:33, Anil Madhavapeddy wrote:
> @seliopou <https://github.com/seliopou>'s patches are a fairly epic
> performance improvement, according to his OCaml 2016 talk...

Great then I can definitely wait for them :)

-- 
Vincent

seliopou:
Async's `Reader` module is already doing buffering, so I don't think this is saving an extra copy. In addition, the mutex around the shared `Bigstring.t` is definitely going negatively affect scalability.
But in any case, the best way to demonstrate performance improvements is to run the benchmarks described in #328.

vbmithr:
I'm not convinced by my patch anymore anyway. But clearly, the `String.sub` in the code is a useless copy. No way to fix this without changing the IO interface though. When are we merging your patches, @seliopou ?

seliopou:
The copy's simply being obscured by `Bigbuffer.contents`, which just calls`Bigstring.to_string` on the underlying buffer.

It'll be a little while until both async and lwt runtimes are fully implemented and tested, but the parts that are being used&mdash;[angstrom](https://github.com/inhabitedtype/angstrom) and [faraday](https://github.com/inhabitedtype/faraday)&mdash;are already open-sourced.

PR #523, opened 2017-01-21
Fd hygiene

rgrinberg:
#506 again

@vbmithr I'm not sure this way of doing fd hygiene on callv is done correctly this way. But it can't hurt I guess


vbmithr:
Ok, go ahead. I think this kind of code should be thoroughly tested anyway, do we do that already?

rgrinberg:
I don't think so. Fd leaks are a bit hard to test since we somehow need to mock raising exceptions at the right time. I'd love to hear some suggestions here however.

Issue #524, opened 2017-01-29
API inconsistencies between lwt and async


superbobry:
Here's a single example from `Server`:

```ocaml
val respond_string :
  ?headers:Cohttp.Header.t ->
  status:Cohttp.Code.status_code ->
  body:string -> unit -> (Response.t * Cohttp_lwt_body.t) Lwt.t
```

and

```ocaml
val respond_with_string :
  ?flush:bool ->
  ?headers:Cohttp.Header.t -> ?code:Cohttp.Code.status_code ->
  string -> response Deferred.t
```

The two functions are (almost) functionally identical, but they have *slightly* different names and *slightly* different parameters. I wonder if this was intentional? i.e. to make the user read the code carefully after the switch to another library.

rgrinberg:
Some differences are indeed intentional, but others are just inconsistencies that you've correctly pointed out.

superbobry:
Would you accept a PR renaming/deprecating some of these?

rgrinberg:
Sure. Just please don't make breaking changes.

A few points:

* Use the `[@@deprecated ..]` to deprecate functions
* Prefer names without "filler" words such `with`
* Reordering labelled/optional arguments doesn't count as breakage. So feel free to just make those consistent on the spot.

gjaldon:
@rgrinberg @superbobry I just created a PR for this issue. I was careful not to introduce any breaking change but did add an optional argument `flush` to the `Lwt.Server.respond_string` function. If the changes are approved, I could work on renaming the other functions and their optional/labeled arguments.

PR #526, opened 2017-02-14
Rename Cohttp_Async.Server.respond_with_string

gjaldon:
We rename respond_with_string to respond_string and replace all occurences of
respond_with_string. We want to make the API for Server in Async and LWT as
similar as possible.

Refs: https://github.com/mirage/ocaml-cohttp/issues/524

The only diff now between `Cohttp_Async.Server.respond_string` and `Cohttp_Lwt.Server.respond_string` is that the LWT version requires an additional `unit` parameter. Why do a lot of the LWT Server functions require a unit parameter while the Async ones don't?


rgrinberg:
Thanks

gjaldon:
Thanks for the merge!

Issue #385, opened 2015-07-11
Streaming body is still iterated/generated after connection closes


wiml:
If the client closes the connection while cohttp is sending a response, cohttp will keep consuming the body stream even though it has nowhere to go.

In the attached example, if you make several requests to the server and cancel them all, you can see in the console that the server process is still generating body text for all of them, apparently forever. The `conn_closed` callback doesn't seem to be called, either.

In this case the body stream is actually infinite, but even if the stream is simply large-but-finite cohttp should avoid computing the entire body unnecessarily.

```
open Lwt
open Cohttp
open Cohttp_lwt_unix

let str_of_cn conn = Connection.to_string (snd conn)
       
let tick_stream conn =
  let num = ref 0 in
  let next_tick () =
    bind ( Lwt_unix.sleep 1. )
	 ( fun () ->
	   let t = (incr num; string_of_int !num) in
	   begin
	     print_endline ( "Returning tick: " ^ t ^ " for connection " ^ (str_of_cn conn) ) ;
	     return (Some ("\r\nevent: tick\r\ndata: " ^ t ^ "\r\n"))
	   end
	 )
  in
  Lwt_stream.from next_tick

let route_req conn req body =
  match Request.meth req with
  | `GET ->
     return (
	 Response.make ~status:`OK
		       ~encoding:Transfer.Unknown
		       ~headers:(Header.init_with "Content-Type" "text/event-stream")
		       (),
	 `Stream (tick_stream conn)
       )
  | _ -> Server.respond_string ~status:`Method_not_allowed ~body:"Unsupported method\n" ()
;;

let _ = Lwt_main.run (Server.create
			~mode:(`TCP (`Port 8000))
			(Server.make
			   ~conn_closed:(fun c -> print_endline ("Connection "^(str_of_cn c)^" closed"))
			   ~callback:route_req ()))

```

PR #528, opened 2017-02-27
Ensure conn_closed is called if client goes away.

mattjbray:
Currently Cohttp calls `conn_closed` when the entire response stream has been written.

However, if there is an exception writing an element of the stream (for example, if the client has closed the connection), `conn_closed` is never called.

I would expect `conn_closed` to be called in the case where the client closes the connection early.

See [this gist][1] for an example server that exhibits this behaviour.

This change would enable the use case of cancelling the producer of a streaming body when the client disconnects (see #385). The user would:

* Maintain a mutable `Map` (or an `Lwt_mvar`) of cleanup hooks (of type `unit -> unit Lwt.t`) keyed on `Cohttp.Connection.t`.
* In the `callback`, register a hook in their map that cancels the `Lwt_stream` producer.
* In the global `conn_closed` hook, lookup the closed connection in the `Map` and call the registered cancellation hook, removing it from the map.

This assumes that `conn_closed` is always called.

The following is possibly out-of-scope for this PR but I'd also like to discuss a nicer API to support this use-case:

Allow the user to return from the `callback` a `this_conn_closed` hook for this specific request, which would be called in the same place as the current global `conn_closed` hook is called. Then the user does not have to maintain the `Map` of cleanup hooks.

The return type of `callback` could be changed to

```ocaml
(Cohttp.Response.t * Cohttp_lwt_body.t * (unit -> unit Lwt.t) option) Lwt.t
```

but perhaps a less invasive change would be to add this optional cleanup hook as a field in `Response.t`.

[1]: https://gist.github.com/mattjbray/103c360337f83d038d8d1016e3c62f35

### DIFF
--- a/lwt-core/cohttp_lwt.ml
+++ b/lwt-core/cohttp_lwt.ml
@@ -289,14 +289,19 @@ module Make_server(IO:IO) = struct
        responses are then sent over the wire *)
     let req_stream = request_stream ic in
     let res_stream = response_stream spec.callback io_id conn_id req_stream in
-    (* Clean up resources when the response stream terminates and call
-     * the user callback *)
-    Lwt_stream.on_terminate res_stream conn_closed;
-    (* Transmit the responses *)
-    res_stream |> Lwt_stream.iter_s (fun (res,body) ->
-      let flush = Response.flush res in
-      Response.write ~flush (fun writer ->
-        Body.write_body (Response.write_body writer) body
-      ) res oc
-    )
+    Lwt.finalize
+      (fun () ->
+         (* Transmit the responses *)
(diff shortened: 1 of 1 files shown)


rgrinberg:
This change is good because it fixes a currently buggy callback. But I'm wondering, why do we need to make the connection closing event part of the api at all? The issue mentioned in the original ticket (the stream getting stuck when the user closes the connection) should just be handled transparently.

So first I'd like to understand why do we need to care from which connection the request is coming from in the first place. After looking at WAI in Haskell, I don't see a way to intercept a connection being closed  either but that doesn't seem to cause any issues.

mattjbray:
Thanks for merging this. I'll open a new issue to continue the discussion.

PR #532, opened 2017-03-01
Add 4.04 to CI matrix

rgrinberg:
Travis is so fast anyway


rgrinberg:
The new travis builds pass with flying colors.

Issue #529, opened 2017-02-28
Lwt streaming responses and flush exceptions


mattjbray:
Continued from #528, responding to @rgrinberg.

> But I'm wondering, why do we need to make the connection closing event part of the api at all?

This is the use-case:

When the client connects, I want to start some concurrent worker threads. As each thread completes, I want to stream its result to the client. The order in which the client receives the results does not matter. If the client disconnects, I want to cancel the remaining threads.

```ocaml
open Lwt.Infix
open Cohttp
open Cohttp_lwt_unix

let server =
  let callback _conn req body =
    (* Create a stream for the response and a push function. *)
    let result_stream, push_result = Lwt_stream.create () in

    (* A worker thread. Perform some work and then push the result onto the
       stream.
    *)
    let do_work i seconds =
      Lwt_unix.sleep seconds >>= fun () ->
      Lwt_log.notice_f "Pushing %i" i >>= fun () ->
      push_result (Some (string_of_int i ^ "\n")) |> Lwt.return
    in

    (* Create a thread that starts some workers, then waits for them to
       complete.
    *)
    let workers =
      [2.0; 1.0; 3.0]
      |> List.mapi do_work
      |> Lwt.join
    in

    (* When all the workers are done, end the stream by pushing None. *)
    let () =
      Lwt.async
        (fun () ->
           Lwt.finalize
             (fun () -> workers)
             (fun () ->
                Lwt_log.notice "Closing stream" >>= fun () ->
                push_result None |> Lwt.return)
        )
    in

    (* Respond with the stream. *)
    let body = Cohttp_lwt_body.of_stream result_stream in

    Server.respond ~status:`OK ~body ()
  in

  (* Catch exceptions while flushing results to the client.

     Without this, the server will crash if the client disconnects before the
     stream has finished.
  *)
  let on_exn exn = Lwt_log.ign_error ~exn "Unhandled exception" in

  Server.create ~mode:(`TCP (`Port 8000)) ~on_exn (Server.make ~callback ())

let () = ignore (Lwt_main.run server)
```

We can tell that the client has disconnected when Cohttp tries to flush the next result an encounters an exception.

So, the missing piece is to call `Lwt.cancel worker` if there is an exception while flushing a result to the client.

It is now possible to implement this using the process described in #528 - maintain a map of connections to `worker` threads, and cancel them in `conn_closed`. But this doesn't feel very nice, or safe.

[...]

PR #534, opened 2017-03-06
For discussion: Cohttp_lwt_body: Add `CustomStream type

mattjbray:
The motivation for this change is to give the user more control over how and when the response body is written. In particular, this allows the user to catch exceptions raised during the write operation.

See #528, #529 (sorry for spreading this out over multiple issues).

This API is similar to how Haskell's WAI implements [streaming responses][1].

Benefits:
* Currently, the user can catch write/flush exceptions in the `on_exn` hook, but there is no way for the user to know which connection caused the exception. The user can get finalizer-like functionality (since #528) by keeping track of connection IDs and matching them up in the `conn_closed` hook. The `CustomStream` API provides a much simpler way of doing this.
* The only API change is to add a constructor to `Cohttp_lwt_body.t`, so downstream libraries (e.g. Opium, https://github.com/rgrinberg/opium/issues/65) are not affected. It is not a breaking change, c.f. adding per-connection `conn_closed`/`on_exn` hooks.

Drawbacks:
* Complicates the API slightly; there are now two ways to create streaming responses.

Other issues:
* I am not sure about the name `CustomStream`.
* WAI also provides the user with a `flush` function. This seemed like a more invasive change, and I haven't got my head around the mechanics of flushing in Cohttp yet, so I haven't tried this. Do we need to?
* I am not sure about the implementations of `is_empty`, `to_stream` and `length`.

[1]: https://hackage.haskell.org/package/wai-3.2.1.1/docs/Network-Wai.html#v:responseStream


rgrinberg:
I think this approach is definitely the correct approach in general. Lwt_stream is quite fragile  and should be deprecated altogether. The other benefits of this approach is that it can be made uniform between Async and Lwt.

> WAI also provides the user with a flush function. This seemed like a more invasive change, and I haven't got my head around the mechanics of flushing in Cohttp yet, so I haven't tried this. Do we need to?

We need a flushing mechanism but it's probably wrong to copy how WAI does it. There are 2 saner ways  to signal flushing:

* Changing the string type input in the function to `string option` and signaling flushes with `None`.
* Signaling flushes by sending the empty string.

I believe WAI used to do the former until they wanted to save some allocations (for reasons that I don't think should affect us). I'm leaning towards the former as well, but I'd like to check how Snap handles this as well.

> I am not sure about the name CustomStream

This doesn't follow our naming convention. It should be `Custom_stream`. I think something like `Gen_body` or `Gen_stream` can also be considered because it's a generator like interface.

hcarty:
It would be nice to have support for cstruct/bigstring-like streams at some point in Cohttp to limit copying.  Or Astring-style substrings.  So perhaps the name could reflect that this is a stream of strings?

rgrinberg:
Most flexible here would be a stream of bodies :P. But for now perhaps we can keep it simple and accept chunks that are either substrings or cstructs

avsm:
> I think this approach is definitely the correct approach in general. Lwt_stream is quite fragile and should be deprecated altogether. 

Agreed!

andreas:
I think this PR can be closed (according to this [comment](https://github.com/mirage/ocaml-cohttp/pull/647#issuecomment-436979312) with #647 merged).

PR #535, opened 2017-03-07
Port library logging to use Logs module

michellemicallef:
Replace logging done in the async and lwt modules to use @dbuenzli's Logs module. The port maintains the behavior that when the COHTTP_DEBUG environment flag is set, debug logs will be printed to std error. A user can set their own reporter that redirects debug logs elsewhere but stderr is the default. Also, a few binaries, such as cohttp_curl_async.ml, have been converted to use Logs.


avsm:
thank you for this PR! It looks like some of the reverse dependencies are failing

```
#          Logs_lwt referenced from /home/opam/.opam/4.03.0/lib/cohttp/cohttp_lwt_unix.cmxa(Cohttp_lwt_unix_io)
#          Fmt referenced from /home/opam/.opam/4.03.0/lib/cohttp/cohttp_lwt_unix.cmxa(Cohttp_lwt_unix_debug)
```

So somewhere in the _oasis file, these dependencies need to be added and `oasis setup` run again.

michellemicallef:
The build failures look unrelated to the change.

avsm:
Confirmed -- this looks like an infrastructure failure of the cloud solver, so only distros that use that are failing. 

```
[WARNING] External solver failed:
[ERROR] External solver failure, please fix your installation and check
        /home/opam/.opam/config and variable $OPAMEXTERNALSOLVER.
        You may also retry with option --use-internal-solver
Cudf_parser.Parse_error("unexpected RFC 822 token", _)
```

Ive restarted the cloud service in the hope it'll fix it, and restarted the jobs. This is not a blocker to merge -- just an attempt to get all the ticks green to give us a nice feel good before merging your PR :)

lwt/cohttp_lwt_unix_io.ml
```diff
@@ -27,29 +27,38 @@ type ic = Lwt_io.input_channel
...
 let write oc buf =
-  if !CD.debug_active then
-    (CD.debug_print ">>> %s" buf; Lwt_io.write oc buf)
+  if CD.debug_active () then (
+    Log.debug (fun f -> f  ">>> %s" (String.trim buf)) >>= fun () ->
+    Lwt_io.write oc buf
+  )
   else
-    (Lwt_io.write oc buf)
+    ( Lwt_io.write oc buf )
```
rgrinberg:
```
else (
  ...
)
```

yomimono:
Late to the party, but thanks a ton for this change @michellemicallef :)

PR #538, opened 2017-03-13
Require at least conduit 0.14.0 for lwt-unix

### LINKED ISSUE #536: Cohttp requires conduit >= 0.14.0
As I reported in ocaml/opam-repository#8524 : Cohttp uses ~on_exn arg of Conduit_lwt_unix.serve that has been introduced in Conduit.0.14.0.

Don't ask me why but tracis-ci loves to use conduit.0.11 which make https://travis-ci.org/Kappa-Dev/KaSim/jobs/210574792 fail

gjaldon:
`~on_exn` argument for Conduit_lwt_unix.serve was added in 0.14.0. Closes #536. Does this fix it?

### DIFF
--- a/cohttp.opam
+++ b/cohttp.opam
@@ -32,7 +32,7 @@ depends: [
   "uri" {>= "1.9.0"}
   "fieldslib"
   "sexplib"
-  "conduit" {>= "0.11.0"}
+  "conduit" {>= "0.14.0"}
   "ppx_fields_conv"
   "ppx_sexp_conv"
   "stringext"


rgrinberg:
No, this constraint should be updated in the opam file instead. You don't have to edit the oasis file.

By the way, when you edit an _oasis file, you should remember to run `$ oasis setup` as well. Otherwise, the change will not affect anything. But this more trivia than information for cohttp, as oasis is on life support until jbuilder is stable enough.

gjaldon:
Made the change on the `cohttp.opam` file instead and replaced the changes in the previous commit. 

Thanks for the info on how to use `oasis`. Will keep that in mind! JBuilder looks promising. Seems a lot easier configuring it with its jbuild file that uses sexp.

PR #540, opened 2017-03-19
avoid reader/writer leaks

vbmithr:
To be merged on Core/Async's official release. I don't think this will compile with released versions.

### DIFF
--- a/async/cohttp_async.ml
+++ b/async/cohttp_async.ml
@@ -60,12 +60,13 @@ module Response = struct
   include (Make(IO) : module type of Make(IO) with type t := t)
 end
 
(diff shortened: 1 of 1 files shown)


rgrinberg:
Why is it necessary to use new async features just to avoid leaks? Can't you back port whatever fix you've made to the old API?

rgrinberg:
Status on this @vbmithr a few words on what the leak is would really help :)

vbmithr:
Let me go through this again.

mseri:
Closing, there have been multiple changes going on in that area.
Feel free to update and reopen it if the leak is still there and you have a fix, make sure to include an explanation of what happens and why please

Issue #544, opened 2017-04-11
Sending binary data in XHR


zoggy:
It seems that sending binary data in XHR is not possible right now, for example with a POST request. This is due to the use of `Js.string` to convert the body (ocaml string containing binary data) to JS string:
  https://github.com/mirage/ocaml-cohttp/blob/master/js/cohttp_lwt_xhr.ml#L169

Replacing `Js.string` with `Js.bytestring` does not solve the problem, since `Js.bytestring` does some character conversion too.

I have implemented a solution and can make a PR but first let me know if it could be ok.

I define the following function to create a typedarray containing the binary data from the ocaml string:

```ocaml
let binary_string str =
  let len = String.length str in
  let a = new%js Typed_array.uint8Array len in
  for i = 0 to len - 1 do
    Typed_array.set a i (Char.code (String.get str i))
  done;
  a
```

Then, I use this function instead of `Js.string` to convert the body to a JS object I pass to `xml ## send `:

```ocaml
           Lwt.return (xml##send (Js.Opt.return (Obj.magic (binary_string body))))
```

Note that the inverse function could be used (instead of `Js.of_bytestring` for the ` convert_body_string` field of functor parameter) when receiving data and this should fix the problem exposed in example  `lib_test/test_xhr.ml`:
```
Download a file from test_net_lwt_server.native
   * There is an issue here with _build/lib_test/test_xhr.byte
   * where the file is ~8MB but we get ~13MB.
```

zoggy:
Ping ?

samoht:
I don't know anything about XHR but if your fix works for you I don't see why we can't include it :-)

I think @andrewray was the one adding that piece of code in the first place, so he might be able to comment.

zoggy:
My fix works for upload but not for download. I'm now encoutenring problems with download too. I will propose soon a fix for both.

andrewray:
I think a typedarray based interface for xmlHttpRequest in js_of_ocaml would be the ideal.  Not sure if there are other issues making this hard, but remember doing a wrapper at some point in the distant past that did this.

Its not too different to the Obj.magic fun here, except it would be hidden behind a nicer typed ocaml interface.

zoggy:
[...]

PR #548, opened 2017-05-05
XHR: upload as binary; download as binary when xhr##response is avail…

zoggy:
…able

I did not test with an old enough browser not supporting xhr#responseType, though.
This PR is related to issue https://github.com/mirage/ocaml-cohttp/issues/544 .


rgrinberg:
@andrewray would you mind reviewing this?

andrewray:
Looks reasonable to me.  I would perhaps drop the console log for every call though.

zoggy:
@andrewray I removed the console log.

PR #549, opened 2017-05-25
Use module aliases in async backend

rgrinberg:
All module in Cohttp_async are exposed under a top level module. This is a
breaking change, but there little users of `Cohttp_async_io`.

I believe the only user of the IO module is @vbmithr


rgrinberg:
@vbmithr Do you want to make a PR? Although I'm not sure how early we want to drop support for pre v0.9 async.

rgrinberg:
Hmm, looks like the travis failures are due to the newly ran tests. Will fix those in another PR

PR #550, opened 2017-05-25
Import mirage-http

rgrinberg:
cc @avsm @samoht @yomimono

### DIFF
--- a/mirage-http-CHANGES.md
+++ b/mirage-http-CHANGES.md
@@ -1,0 +1,73 @@
+### 3.1.0
+
+* Add `Cohttp_mirage_static` module for serving static files from a
+  read-only key-value store.  Includes magic mime detection.
+* Improve the ocamldoc strings for the modules.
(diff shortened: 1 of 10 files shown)


mirage-http-CHANGES.md
```diff
@@ -0,0 +1,73 @@
+### 3.1.0
```
samoht: @rgrinberg Do you know if that file will be picked automatically by topkg for the releases? By default it picks CHANGES.md but maybe topkg-jbuilder does something clever? /cc @diml and @dbuenzli

dbuenzli: The file that is picked up is the first one mentioned in `change_logs` [here](http://erratique.ch/software/topkg/doc/Topkg.Pkg.html#VALdescribe). You should be able to specify an alternate one the cli in the `log` and `publish` options via the `--change-log` argument.

PR #551, opened 2017-05-25
add topkg-jbuilder

rgrinberg:
@vbmithr @avsm


pkg/pkg.ml
```diff
@@ -0,0 +1,7 @@
+#use "topfind"
+#require "topkg-jbuilder"
```
samoht: should probably be `topkg-jbuilder.auto` as described in the docs. This will make topkg automatically lint all the top-level opam files and compile the project and run the tests on release (and other nice things).

rgrinberg: Don't I need to use the manual one because jbuilder will not be able to pick up the project name automatically because of mirage-http?

rgrinberg:
@samoht @dbuenzli I configured topkg to use both change log files.

dbuenzli:
Note that this will publish only changes found in `CHANGES.md` on the gh release page and in the opam repository PR.

rgrinberg:
If that's the best that can be done then it's fine. That changes file will no longer be updated anymore anyway. I'm just not sure how to merge the two.

PR #552, opened 2017-05-26
remove warnings with new Async

vbmithr:
As you said, nobody uses Async with cohttp. And seriously, there is no reason to use older Janestreet libraries now that new and better ones are released. And also, such user can use older release of cohttp is he/she really wants to.


vbmithr:
Should be good now.

PR #553, opened 2017-06-21
fix jbuild for bleeding edge conduit

vbmithr:


avsm:
lgtm

i think we need to agree on versions and add constraints/pins to cohttp/conduit/dns, @rgrinberg

rgrinberg:
Yes please. We need some up to date constraints her.

avsm:
see #557

PR #555, opened 2017-06-23
Change folder structure for jbuilder

rgrinberg:
The new project structure is such that every package is self contained in its
own directory.

I've also added the cohttp-top package (which was missing in the original port)


rgrinberg:
I've also removed all the manual tests. Nobody runs these manually so they're really not helping much. If someone would like to bring them back, please turn them into a proper test suite first.

rgrinberg:
This revision also reintroduces the various async/lwt curl/static serve binaries. I've elected to put them in their own packages because they have their own dependencies (like cmdliner) for example.

cohttp-lwt-unix-bin/jbuild
```diff
@@ -0,0 +1,7 @@
+(jbuild_version 1)
+
+(executables
+ ((names (cohttp_curl_lwt cohttp_proxy_lwt cohttp_server_lwt))
+  (libraries (cohttp-lwt-unix cohttp_server cmdliner))
+  (package cohttp-lwt-unix-bin)
```
avsm: Is it really worth having a separate OPAM package for the binaries? The cmdliner dependency is pretty minor and doesnt have a runtime impact on the cohttp_lwt_unix library. I'd suggest folding cohttp-lwt-unix-bin back into cohttp-lwt-unix

cohttp-top.opam
```diff
@@ -0,0 +1,26 @@
...
+tags: ["org:mirage" "org:xapi-project"]
+dev-repo: "https://github.com/mirage/ocaml-cohttp.git"
+build: [
+  ["jbuilder" "subst" "-n" name] {pinned}
+  ["jbuilder" "build" "-p" name "-j" jobs]
+]
+build-test: [["jbuilder" "runtest" "-p" name "-j" jobs]]
+depends: [
+  "jbuilder" {build & >= "1.0+beta10"}
+  "cohttp"
```
avsm: might as well put a minimum version in here that is greater than the current released cohttp, to ensure that --dev picks up a useful version

cohttp/src/jbuild
```diff
@@ -2,7 +2,7 @@
 #require "unix"
 
 let version =
-  let ic = open_in "../VERSION" in
+  let ic = open_in "../../VERSION" in
```
avsm: there's got to be a better way to do version stamping, but not a blocker to merge :-)

rgrinberg: This is the workaround I got from jsoo ^_^

rgrinberg:
@avsm Halp. looks like the version constraints are making the CI fail.

cohttp/src/jbuild
avsm: @hhugo @diml -- is it worth an upstream jbuilder issue to track how to improve creating version file stamps?  probably good not to to propagate too many hacky rules throughout the ecosystem.

cohttp-async/bin/jbuild
```diff
@@ -0,0 +1,7 @@
+(jbuild_version 1)
+
+(executables
+ ((names (cohttp_curl_async cohttp_server_async))
+  (package cohttp-async-bin)
```
avsm: I think this needs to be cohttp-async, which is killing the build (not constraints).  Same with the cohttp-lwt-unix jbuild

cohttp/src/jbuild
samoht: I though it was why we had `%%VERSION%%` and `opam subst` :-)

rgrinberg:
@avsm that was indeed a bug, but it wasn't the cause of the CI failures

avsm:
Oh I see...

```
The following dependencies couldn't be met:
  - cohttp-async -> cohttp > 0.22.0
Your request can't be satisfied:
  - cohttp>0.22.0 is not available because the package is pinned to /repo, version 0.22.0.
```

rgrinberg:
I'll revert those constraints for now

cohttp/src/jbuild
diml: Now you can also do:

```scheme
(rule (with-stdout-to version.ml "let version = {|${version:foo}|}\n"))
```

PR #560, opened 2017-07-06
Officially deprecated Cohttp_lwt.{Client,Server,Net} aliases

rgrinberg:


rgrinberg:
I think the minimum is 4.03 already because of ppx_sexp_conv. But that
attribute was introduced in 4.02.3

avsm:
Thanks for the clarification on both counts -- I find the changelog impossible to search in these instances.

avsm:
This needs to be merged for the 0.99 release I think

PR #561, opened 2017-07-06
Add Cohttp_lwt_unix.Net alias

rgrinberg:
In preparation for when Cohttp_lwt_unix switches to module aliases

@avsm this is also required for 0.99


avsm:
I'm respinning 0.99 now

PR #562, opened 2017-07-11
0.99 respin

avsm:
- fix lots of warnings
- autobuild code.ml
- build the examples/ directory (not installing binaries)
- make all the jbuild files static text


avsm:
This also populates the changelog and makes 0.99 releasable.

PR #566, opened 2017-07-14
cohttp.opam: enforce a modern fieldslib

avsm:
otherwise downgrades result in the deriving fields annotations
not being picked up (e.g. rubytt) and build failures


avsm:
noticed in revdeps for https://github.com/ocaml/opam-repository/pull/9824

PR #568, opened 2017-07-15
wrap cohttp-lwt and cohttp-lwt-unix

rgrinberg:
So that both libraries export a single top level module as we'd like.

this is required for 1.0 but should not be merged until 0.99.0 is released on
opam


rgrinberg:
@avsm I'm merging. There should be nothing surprising here anyway.

avsm:
belated lgtm, was on vacation last week -- back now

PR #572, opened 2017-07-30
Rename mirage-http to cohttp-mirage

rgrinberg:
Consistency with other libraries


avsm:
Not yet -- we need to coordinate this with all the other renames (there are various PRs floating around that need to be combined and then a new release of mirage/mirage done to account for the new names)

rgrinberg:
OK. Feel free to take over this effort.

avsm:
I'll coordinate these...

djs55:
I think we need to release `cohttp` soon -- the released version doesn't build with OCaml 4.06 (fix is in master).

I've made a change to the `mirage` tool and will try to make the CI pass (probably require some PINs). Assuming I get the `mirage` tool building, is there anything else needed before merging this and tagging 1.0.0?

avsm:
nothing else except `mirage` tool I believe...

PR #574, opened 2017-08-05
Specify dependencies more precisely

rgrinberg:
fmt/logs/magic-mime are in fact not dependencies of cohttp and should be
moved to their appropriate subpackages or set as test deps.


avsm:
lgtm!

PR #576, opened 2017-08-18
Add the convenience function Header.is_empty

dinosaure:


cohttp/src/header.mli
```diff
@@ -22,6 +22,9 @@ type t [@@deriving sexp]
 (** Construct a fresh, empty map of HTTP headers *)
 val init : unit -> t
 
+(** Test whether a map is empty or not. *)
```
rgrinberg: map? you mean header here

dinosaure: Done 👍

cohttp/src/header.mli
```diff
@@ -22,6 +22,9 @@ type t [@@deriving sexp]
 (** Construct a fresh, empty map of HTTP headers *)
 val init : unit -> t
 
+(** Test whether a HTTP headers is empty or not. *)
+val is_empty : t -> unit
```
rgrinberg: `-> bool` I think

And it should `whether HTTP headers are empty...`

dinosaure: 😄

PR #577, opened 2017-09-15
Update README.md

Gregoirevda:
I couldn't run the example without this change. 

Adding `cohttp-lwt-unix` as package in my _tags says it's not defined.


rgrinberg:
Which version of cohttp are you using? All extraneous top level module names have been deprecated so the readme is correct in not using them.

PR #580, opened 2017-11-04
Fix build with OCaml 4.06 (and -safe-string)

### LINKED ISSUE #579: Does not build under 4.06.0 with safe-string enabled
I may have time to look at this in a week or two.  It would be nice to get cohttp usable under 4.06.0 as quickly as we can!

djs55:
Fixes #579

For OCaml 4.06 it requires
- https://github.com/mirage/ocaml-uri/pull/108
- `hex.1.2.0`

I've not tested this with the `mirage-http` package as this depends transitively on `ppx_deriving` which requires OCaml 4.05 or earlier.

### DIFF
--- a/cohttp/scripts/generate.ml
+++ b/cohttp/scripts/generate.ml
@@ -128,11 +128,13 @@ type section = {
 }
 
 let normalize s =
+  let b = Bytes.of_string s in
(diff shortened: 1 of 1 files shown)


djs55:
CI is green; merging to unblock fixing of other libraries.

PR #586, opened 2017-11-18
Set default user-agent

TheCBaH:
This will make cohttp-based clients always set the user-agent request field.
See https://tools.ietf.org/html/rfc2616#section-14.43


avsm:
thanks!

PR #518, opened 2016-12-15
add exn_hook to match conduit

vbmithr:
To be merged after https://github.com/mirage/ocaml-conduit/pull/181 is merged.

PR #589, opened 2017-12-07
Catch exceptions when closing channels in mirage client

ansiwen:
Lwt.ignore_result doesn't catch exceptions, which can result in
uncatchable exceptions, because these exceptions will be handled by
Lwt.async_exception_hook.  In race conditions and when using the
socket network, the closing of channels can result in a 'Unix.Unix_error
(Unix.EBADF, "check_descriptor", "")' exception. This change catches
these exceptions.

### DIFF
--- a/cohttp-mirage/src/client.ml
+++ b/cohttp-mirage/src/client.ml
@@ -45,7 +45,9 @@ module Net_IO = struct
 
   let close_in _ = ()
   let close_out _ = ()
-  let close ic _oc = Lwt.ignore_result (Channel.close ic)
+  let close ic _oc = Lwt.ignore_result @@ Lwt.catch
+    (fun () -> Channel.close ic)
+    (fun e  -> Lwt.return @@ Ok ())
 
 end
 let ctx resolver conduit = { Net_IO.resolver; conduit }


ansiwen:
Here is a reproducer for this race condition: https://gist.github.com/ansiwen/baece8a522d5974310f3001b8db8e015

rgrinberg:
I'm +1 on this PR. But is it possible that we're swallowing some useful exceptions when closing?

samoht:
Thank you for the patch!

As @rgrinberg I think we should display the error instead of just throwing it away (use `Logs`). Also, this seems to show that there is a race in the socket stack that should be fixed upstream. I will open an issue to keep track of that.

ansiwen:
Please note, that I adopted the "best practice" ;-) from the unix client here: https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-lwt-unix/src/net.ml#L45
So, if you guys prefer to have a log message, how about a second PR, where I add that Log to all other Lwt.ignore_result calls as well, to have it all consistent?

@rgrinberg If there are important exceptions, then we should not use Lwt.ignore_result at all I guess, because - as you probably know - it returns immediately and creates a "loose-end" thread. If that thread raises an exception, it's pretty undefined, which handler will catch it, and makes it basically uncatchable. In my case, I randomly had _three_ different cases: either my own handler around the Client.call function caught it, or the Cohttp callback caller caught it, or - worst case - Lwt.async_exception_hook caught it, which kills the whole process (and which - as a side note - I never managed to overwrite. If I set the reference to := ignore, it still called the original handler. Weird.) So, if we want to catch these exceptions in a consistent manner, Net.close has to return a proper Lwt.t value, I guess.

@samoht I agree, there there is probably another issue there, but it might be a tough one, I think you can never really guarantee, that a socket has not been closed already in the meantime. I rather think, that there might be another (third?) problem that is something like a stack-memory-leak with many http requests in one persistent connection to the http server. I got very long stack traces with repeating call paths during my experiments, which looks to me like a missing tail call optimization while processing the repeating HTTP requests. Here are examples:

Backtrace of Unix.Unix_error (Unix.EBADF, "check_descriptor", "") from Lwt.async_exception_hook:
https://pastebin.com/HbwUg7Lv

GDB backtrace with breakpoint in connection-closed-callback:
https://pastebin.com/NfacXSGU

ansiwen:
Do we maybe need something like #518 but for the client cases? (That is, an on_exn hook.)

rgrinberg:
I think that's not a bad idea. It's an async convention, but Lwt's existing mechanism of having a global mutable hook is really bad.

samoht:
@ansiwen your plan makes sense. Before merging that PR, could you fix the travis scripts by replacing `mirage-http` by `cohttp-mirage`? It seems that the change was not done properly where the renaming was done.

ansiwen:
@samoht I did, but I'm not sure that will fix this CI issue: https://github.com/mirage/mirage-dev.git

samoht:
yea the other error is an issue with `safe-string` support in the tests. Would be great to fix it but not in that PR. I'm fine to merge this one when the mirage tests are green.

ansiwen:
@samoht: the other alpine job, which I changed to cohttp-mirage, also fails with an unrelated error: https://travis-ci.org/mirage/ocaml-cohttp/jobs/313631823#L2020
The other cohttp-mirage tests are successful. So, should I revert the mirage-http->cohttp-mirage change and we fix it together with the other CI issue in a separate change?

rgrinberg:
> The other cohttp-mirage tests are successful. So, should I revert the mirage-http->cohttp-mirage change and we fix it together with the other CI issue in a separate change?

There's no need. That commit is correct so I will cherry pick it to master.

rgrinberg:
Wait, it looks like you only have 1 commit here. I will push the CI fix and you can remove it from your PR here :)

ansiwen:
@rgrinberg: or, rebased on master and pushed again... but the two CI jobs will still be failing (not related to this change)

PR #591, opened 2017-12-19
Add ppx_type_conv lower bound

rgrinberg:
Previous versions don't interact with ppx_deriving that well.

Best would be to somehow black list this version in opam-repository. But it doesn't hurt adding this here.

@hannesm

### DIFF
--- a/cohttp.opam
+++ b/cohttp.opam
@@ -26,6 +26,7 @@ depends: [
   "uri" {>= "1.9.0"}
   "fieldslib"
   "sexplib"
+  "ppx_type_conv" {build & >="v0.9.1"}
   "ppx_fields_conv" {build & >="v0.9.0"}
   "ppx_sexp_conv" {build & >="v0.9.0"}
   "stringext"


hannesm:
@rgrinberg if `ppx_type_conv` v0.9.0 does not work anywhere (does it work anywhere?), I'd be in favour of deleting it from opam-repository!?
as far as I understand @diml, `ppx_type_conv` v0.9.0 needs a conflict with `ppx_deriving.4.2`!?  once this is added, is this change here needed (well, if cohttp depends on ppx_type_conv, it should have a dependency in opam).  I'm not familiar with either ppx_type_conv, ppx_deriving, nor cohttp to judge which change(s) are needed.

rgrinberg:
> if ppx_type_conv v0.9.0 does not work anywhere (does it work anywhere?), I'd be in favour of deleting it from opam-repository!?

It does work for other converters that don't generate modules, since in this situation generating the same code doesn't hurt.

diml:
Adding the conflict seems fine. Hopefully it shouldn't make packages uninstallable

PR #592, opened 2017-12-19
Add log warnings for uncaught exceptions

ansiwen:
Lwt.ignore_result doesn't catch exceptions, which can result in
uncatchable exceptions, because these exceptions would be handled by
Lwt.async_exception_hook. As a follow up to PR #589 this change makes
sure all exceptions are caught and a warning log message is produced.


cohttp-lwt-unix/src/debug.ml
```diff
@@ -28,7 +28,13 @@ let default_reporter () =
     let k _ =
       let write () = Lwt_io.write Lwt_io.stderr (fmtr_flush ()) in
       let unblock () = over (); Lwt.return_unit in
-      Lwt.finalize write unblock |> Lwt.ignore_result;
```
rgrinberg: Can Lwt.finalize actually raise?

ansiwen: Yes, IIUC it would forward the exception of both functions:

> finalize f g runs f (), and then g (), whether f () fails or completes normally. Lwt.finalize returns the 
> result of f (), unless g () fails, in which case it returns the exception raised by g ().
(https://ocsigen.org/lwt/dev/api/Lwt)

rgrinberg: Gotcha

rgrinberg:
Why can't these exceptions be logged in the async_excpetion_hook again? With close, I think it's ok to make close non raising, but things like writes are things that a user should log/possibly do something about.

samoht:
> Why can't these exceptions be logged in the async_excpetion_hook again

I think that's a better practice to log the exception where they might be discareded instead of relying on a global top-level monitor (that people forget most of the time to set-up anyway). So I really like this patch and I'd like to see it merged, thanks!

ansiwen:
@rgrinberg:
> Why can't these exceptions be logged in the async_excpetion_hook again? With close, I think it's ok to make close non raising, but things like writes are things that a user should log/possibly do something about.

I see several issues with ignore_result and async_excpetion_hook:
1. by default it fatally exits your program, which is usually not what you want
2. ignore_result is by definition a fire-and-forget function. If you are interested in a possible exception, don't use it, instead of relying on a global hook.
3. it is a race condition if the exception actually hits async_excpetion_hook. In case the thread ended already, it would be raised locally and end up in the callback catch. So you can't programmatically catch these exceptions if you use ignore_result.
4. nobody knows about async_excpetion_hook, and it's not trivial to overwrite it. I failed in my few attempts. I saw my Irmin server "crash" several times, just because the client closed the connection while the server was sending something.

So my conclusion is: Exceptions raised in ignore_result threads can only be used for signalling programming errors. However, exceptions coming from close or write calls are not programming errors, but API errors, and can happen during normal operation. They have to be caught within the ignore_result thread.

rgrinberg:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

rgrinberg:
I'm OK with the change to ignore exceptions thrown when closing fd's. But I still don't see the improvement to writing to stderr. If we fail to write stderr once, it's likely the failure will occur again. Why should we just log a failure to write to stderr and then keep retrying to log anyway over and over? I don't see how that behavior is desirable. That doesn't meant the old behavior was good, but I don't think we're making it better here. I think that even making the logging failure fatal is better.

ansiwen:
I just noticed this stale PR. What is the conclusion here? @samoht says he likes it like that, @rgrinberg would like to see the failed write to stderr handled differently (fatal). I'm fine with both, don't have a strong opinion here. (_If_ this _really_ can't happen in code, that we would prefer to continue to run, like it was the case for the failed close.)

I only oppose to continuing using `async_excpetion_hook`, because that leads to undefined behavior because of a race condition (the exception might get caught there or raised locally).

rgrinberg:
@mseri can you take over this one?

mseri:
Sure!

mseri:
I agree with the approach of avoiding the `async_excpetion_hook`, we did have problem with that as well in the past. I also agree with @rgrinberg that maybe failures on write should propagate instead of being ignored, @ansiwen why do you think that failures in writes are only API errors?

mseri:
Overall I am actually inclined to merge this as it is now, and eventually change that by fixing also https://github.com/mirage/ocaml-cohttp/issues/645

ansiwen:
@mseri maybe "API error" is the wrong terminology. What I mean is, that it is not an unexpected error, but an expected programmatic error that the API legitimately returns, that is out of our control, and it doesn't necessarily mean, it's a fatal error. Maybe in the case of writing to stderr it can be interpreted as fatal, but writing to a file in general isn't.

mseri:
Let's merge it as it is now. Keeping in mind that the debug behaviour may change again once we address #645

PR #594, opened 2018-01-03
Expose the IO module in cohttp-mirage

samoht:
Not sure why this one hidden and I can't find the reason in the history either.


rgrinberg:
It was hidden because I don't think it was used anywhere. This module is an implementation detail of the library, what use case do you have for it?

samoht:
This is needed by `ocaml-git` for implementing the smart HTTP protocol (where this is quite a bit of low-level logic to implement).

PR #605, opened 2018-03-19
Add sexp_conv ppx

kevinqiu:
This adds a preprocessor statement to the async package's `jbuild`. Without the statement, the `[@@deriving sexp_of]` statement is ignored.

### DIFF
--- a/cohttp-async.opam
+++ b/cohttp-async.opam
@@ -26,6 +26,8 @@ depends: [
   "magic-mime"
   "logs"
   "fmt"
+  "sexplib"
+  "ppx_sexp_conv" {build & >="v0.9.0"}
   "ounit" {test}
 ]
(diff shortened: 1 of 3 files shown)


rgrinberg:
Thanks. Would you mind updating cohttp-async.opam appropriately as well? `ppx_sexp_conv` should be a `{build}` dependency while `sexplib` a runtime dependency.

kevinqiu:
Is the transitive dependency though `cohttp` not enough? I noticed that `cohttp-lwt.opam` doesn't include `ppx_sexp_conv` or  `sexplib` even though it invokes `ppx_sexp_conv` in it's `jbuild`. Is that an oversight and the dependencies should be added there also?

rgrinberg:
Transitive dependencies are not to be relied on. The dependency list in
cohttp-lwt should also be updated then.

PR #607, opened 2018-03-25
Update code base to be compatible with new async/core

rgrinberg:
@avsm after this I will get rid of Core's command module. It adds an extra dependency (Async_extra) and it just makes us maintain 2 command line interfaces. Cmdliner is superior in every, we should just use it.

### DIFF
--- a/cohttp-async/bin/cohttp_server_async.ml
+++ b/cohttp-async/bin/cohttp_server_async.ml
@@ -15,8 +15,9 @@
  *
   }}}*)
 
-open Core
-open Async
+open Base
+open Async_kernel
+open Async_unix
 
 open Cohttp_async
 open Cohttp_server
@@ -40,7 +41,7 @@ let serve ~info ~docroot ~index uri path =
     (* Get a list of current files and map to HTML *)
     | `Directory -> begin
       let path_len = String.length path in
-      if path_len <> 0 && path.[path_len - 1] <> '/'
+      if Int.(path_len <> 0) && Char.(path.[path_len - 1] <> '/')
       then Server.respond_with_redirect (Uri.with_path uri (path^"/"))
       (* Check if the index file exists *)
       else Sys.file_exists (file_name / index)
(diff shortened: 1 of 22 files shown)


avsm:
Thanks, the patchset looks good.  However, the intention of using Command here was to stay in the same territory as the rest of the Base/Core/Async world.  In particular, this means that the Async scheduler interactions are taken care of by using `Async.Command`.

@yminsky - depending on async_extra just to be able to build command-line Async applications does seems like dependency creep.  Would it be possible to get that shifted out into a smaller dependency?

cohttp-async/bin/cohttp_server_async.ml
```diff
@@ -40,7 +41,7 @@ let serve ~info ~docroot ~index uri path =
     (* Get a list of current files and map to HTML *)
     | `Directory -> begin
       let path_len = String.length path in
-      if path_len <> 0 && path.[path_len - 1] <> '/'
+      if Int.(path_len <> 0) && Char.(path.[path_len - 1] <> '/')
```
avsm: It's really nice to see polymorphic comparisons disappear!

yminsky:
Looping in @sweeks and @seliopou on the subject of breaking out into something less than async_extra.

seliopou:
The `Tcp` module's in `async_extra`, so moving `Command` into a separate library won't do anything to eliminate the dependency. But either way,  moving `Command` to Async proper seems like the right thing to do.

seliopou:
Sorry, scratch that last part. I got `async_extra` confused with `async_extended`. `async_extra` is a dependency of Async, so I'm not quite sure what breaking `Command` into a separate library or moving it around would accomplish.

yminsky:
I think that the suggestion is that cohttp_async can right now work with only async_kernel and async_unix, which is plausible, in the absence of command. I'm not sure that async_extra is big enough to matter, though. Is there a particular reason this is important?

As a side note, someone who wants to use Cmdliner instead of Command might want to think about breaking out the Async support from Command so it could be used separately. There is a bunch of useful logic there, if I remember correctly.

seliopou:
But it can't work with just `async_kernel` and `async_unix` because the `Tcp` module comes from `async_extra`. So even if `Command moves`, the the dependency on `async_extra` is still there, probably via `conduit-async` rather than directly.

I think it may make sense to move at least some of the logic from [`in_async`](https://github.com/janestreet/async_extra/blob/master/src/command.ml#L48-L75) to [`Writer.behave_nicely_in_pipeline`](https://github.com/janestreet/async_unix/blob/master/src/writer.mli#L586-L593). Seems like a good home for it.

PR #608, opened 2018-04-11
Remove all instance of Lwt_log

rgrinberg:
Use Logs exclusively. This makes us compatible with lwt 4.0 as well


raphael-proust:
One nitpick: It changes the synchronous/asynchronous behaviour. Specifically, it only uses the `Logs` module which return `unit` instead of the `Logs_lwt` module which returns `unit Lwt.t`. In some cases, this is in line with the previous use, in other cases it's not.
See #609 

One issue: In `bin/cohttp_server_lwt.ml` you do not set a reporter. According to Logs' documentation:

> If you are writing an application you must remember to set the reporter before any logging operation takes place otherwise no messages will be reported.

mseri:
This has been superseded by #609, already merged. I am going to close. Feel free to reopen it if I am wrong

PR #609, opened 2018-04-12
Remove deprecated Lwt_logs

raphael-proust:
With the release of Lwt 4.0.0, the module `Lwt_logs` is deprecated. This PR replaces uses of this deprecated module with uses of `Logs` and `Logs_lwt`.

This is similar to #608 except that it uses `Logs_lwt` where there used to be Lwt.

### DIFF
--- a/cohttp-lwt-unix/bin/cohttp_server_lwt.ml
+++ b/cohttp-lwt-unix/bin/cohttp_server_lwt.ml
@@ -21,6 +21,8 @@ open Cohttp_lwt_unix
 
 open Cohttp_server
 
+let log_src = Logs.Src.create "cohttp-lwt-unix.bin.server"
+
 let method_filter meth (res,body) = match meth with
   | `HEAD -> Lwt.return (res,`Empty)
   | _ -> Lwt.return (res,body)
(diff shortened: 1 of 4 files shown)


rgrinberg:
We can go along with this PR instead. But you haven't added `logs` as a dependency everywhere that you're using. We shouldn't rely on it being available via transitive deps.

raphael-proust:
Ok, I'll improve this PR.

- [ ] add Logs depepndencies
- [ ] install Lwt-aware reporter in binary

rgrinberg:
Btw, do we really need to use `logs.lwt`? I know it's required for a faithful port of the logging, but I kind of doubt the initial version really relied on making sure the logging was flushed before proceeding - it was just something that was done b/c it was easy. So:

* I really dislike binding on log messages in the main code path

* `logs.lwt` still relies on optional dependencies to be installed unfortunately. I worked pretty hard to get rid of most of them in cohttp, it would be a shame to reintroduce them.

Would it be possible to just use vanilla `Logs` here?

rgrinberg:
> It would change the behaviour of the binaries which I don't know if it's ok or not.

It's definitely OK. As I said, we're probably giving this a lot more thought than anyone has before :)

raphael-proust:
I hadn't notice preexistent occurrences of `Logs`.

`Io` uses Lwt-aware logging functions, and also works around the level system of Logs.
`Debug` sets its own reporter (which goes again usage guidelines of the Logs library).

Should I just leave those untouched? Or should I make it more Logs-compliant?

rgrinberg:
> Io uses Lwt-aware logging functions, and also works around the level system of Logs.

Yeah, that was probably a mistake. I ported this code originally with @michelka and we just ported things while staying as faithful as possible. On second glance, it's not really necessary.

> Debug sets its own reporter (which goes again usage guidelines of the Logs library).

Yes this unfortunate but should probably stay. We want to stay compatible with the `COHTTP_DEBUG` env var to turn on logging automatically.

> Should I just leave those untouched? Or should I make it more Logs-compliant?

Feel free to get rid of `Logs_lwt`.

raphael-proust:
Got rid of `Logs_lwt`, even in the tests.

It might be better to move the side-effect part of the `Debug` module into the binaries. But that's a separate issue.

PR #611, opened 2018-04-17
do not rely on locale for printing qvalues

vbmithr:
Some libraries manipulate the C library locale settings and therefore special care must be taken to print qvalues.


rgrinberg:
Thanks Vincent. Would you mind adding a unit test?

rgrinberg:
Thanks. It's not that I don't trust your implementation, I want to make sure that nobody changes it accidentally :)

PR #612, opened 2018-05-01
mirage: impose a 4k limit on header lengths

djs55:
Although the HTTP spec does not impose a maximum limit on header lengths, most implementations impose a limit to avoid a client from exhausting all available server memory.

According to https://stackoverflow.com/a/8623061 the typical limit is between 4k and 48k, but this usually applies to the sum of the request line and all the headers. It's more convenient for us to
apply the limit per header.

This requires [mirage/mirage-channel#25]


mseri:
Is this still needed?

mseri:
Feel free to reopen and ping me if it is

PR #618, opened 2018-05-17
A chunk can have more than one chunk extension

djs55:
According to RFC7230 section 4.1.1[1]

> The chunked encoding allows each chunk to include zero or more chunk
> extensions, immediately following the chunk-size
...
> chunk-ext      = *( ";" chunk-ext-name [ "=" chunk-ext-val ] )

This means that the initial line can be
```
  1234;foo=bar;bar=foo
```
Therefore we should use `String.index` to find the left-most `;` rather than `String.rindex` which finds the right-most `;` when parsing the chunk size (the first number)

[1] https://tools.ietf.org/html/rfc7230#section-4.1.1

### DIFF
--- a/cohttp/src/transfer_io.ml
+++ b/cohttp/src/transfer_io.ml
@@ -34,8 +34,11 @@ module Make(IO : S.IO) = struct
 
     let parse_chunksize chunk_size_hex =
       let hex =
-        (* chunk size is optionally delimited by ; *)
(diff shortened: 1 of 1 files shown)


mseri:
Thanks for the fix. Could you also add a test for this?

mseri:
I have checked, to add a test we need to introduce inline tests. This is postponed to a separate PR

PR #623, opened 2018-06-26
Case sensitive header fields

liyishuai:
https://httpwg.org/http-core/draft-ietf-httpbis-messaging-latest.html#header.fields
> Each header field consists of a **case-insensitive** field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.


avsm:
I'm afraid I don't understand what this patch is trying to solve :-) The RFC you quote confirms that header fields are *case-insensitive*, but then your PR is reverting that behaviour and making them case-sensitive... the opposite of the desired behaviour.

PR #632, opened 2018-09-04
Fix travis

mseri:
This should allow us to rely a bit more blindly on Travis, as there were some tests that were expected to fail and not all the recent compilers where tested.


mseri:
There are two failure remaining, that I don't know how to solve:
```
[ERROR] The sources of the following couldn't be obtained, aborting:
          - ounit.2.0.8: Curl failed
```

Rerunning the two tests does not help

samoht:
Use `alcotest`? :-)

More seriously, it's probably due to a network glitch..,

mseri:
I had rerun the tests a handful of times between yesterday and today, but I think you have the right solution... time to add a commit to move to alcotest

samoht:
Note also that you can remove the minor version, as the container have pre-build switches for the `4.0*` compilers (>= 4.03).

mseri:
The port to alcotest of the remaining tests is a bit less straightforward than I envisioned. I suggest we keep it in a separate PR. I’l Uodate the minor version from the tests then I think it’s all for this PR

mseri:
Should we merge this and/or prepare a new release? This should allow us to close some issues already

avsm:
sgtm

PR #637, opened 2018-10-11
Replace sexplib with sexplib0

mseri:
This is a thin layer that is already linked as part of
`ppx_sexp_conv.runtime-libs`

Note that with the recent analogous change in ocaml-uri (https://github.com/mirage/ocaml-uri/pull/124),
this should positively affect the binary size of executables using `cohttp`.


mseri:
in fact if we are willing to drop a few versions of `ppx_deriving_sexp`, we could just use `ppx_deriving_sexp.runtime-lib` instead of declaring `sexplib0` explicitly

avsm:
Oh, we should include this as well in the release.

PR #642, opened 2018-10-21
cohttp-async: require new uri (was used in client.ml)

mseri:
This should prevent breaks when some specific versions of cohttp,
cohttp-async and uri pinned are installed.


mseri:
Also relative PR to opam: https://github.com/ocaml/opam-repository/pull/12842

PR #644, opened 2018-10-27
Add a note about the need for SSL support

jbranchaud:
This was quite a gotcha for me as I was trying to work through the _Client Tutorial_ example. After a lot of digging and question-asking I was able to arrive at a proper configuration of Cohttp and its dependencies such that I was able to request an HTTPS page. Here is [a relevant commit](https://github.com/jbranchaud/reasonml-cohttp-example/commit/a31c72ee412d7ec3e626d25e597af8ded7d8a34a#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) for reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,13 @@ There's a few things to notice:
 * We must trigger lwt's event loop for the request to run. `Lwt_main.run` will
   run the event loop and return with final value of `body` which we then print.
 
+Note that in order to request an HTTPS page like in the above example,
+you'll need Cohttp to have been compiled with SSL or TLS. For SSL, you'll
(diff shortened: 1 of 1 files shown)


mseri:
Thanks, this is indeed important. Can you add that this means having installed `lwt_ssl` before installing cohttp?

hcarty:
> Can you add that this means having installed lwt_ssl before installing cohttp?

Or tls installed.

jbranchaud:
@mseri @hcarty Thanks for the feedback. I've made some updates. Let me know how it looks!

hcarty:
@jbranchaud Looks good to me! Thanks

mseri:
Thanks!

PR #647, opened 2018-11-05
Add expert mode for cohttp-lwt

andreas:
This PR adds "expert mode" to `cohttp-lwt` similar to what exists for `cohttp-async` (#488).

My primary use case is for websocket support like #488. The changes also relate to other issues regarding streaming though (#534, #529), and this approach could be used to solve those use cases.  An `Expert`-response does currently not allow omitting a body (a decision inherited from #488), but if that was permitted, then it could be used to for writing/streaming bigarrays efficiently ([mentioned here](https://github.com/mirage/ocaml-cohttp/pull/534#issuecomment-284478847)).

I've rewritten the response/response handling, such that it no longer uses `Lwt_stream`. Though the abstraction of mapping a stream of requests to a stream of responses sounds elegant, I think it's actually simpler without it. In particular the use of refs (`early_close`) and  mutexes (`read_m`) is no longer required. I'm happy to receive feedback on the choice though.

I've not added any tests, but I'd be happy to if the overall approach is approved.

/cc @mattjbray

### DIFF
--- a/cohttp-async/src/server.mli
+++ b/cohttp-async/src/server.mli
@@ -16,9 +16,16 @@ type 'r respond_t =
   ?body:Body.t ->
   Cohttp.Code.status_code -> 'r Async_kernel.Deferred.t
 
+(** A request handler can respond in two ways:
+    - Using [`Response], with a {!Response.t} and a {!Body.t}.
+    - Using [`Expert], with a {!Response.t} and an IO function that is expected
+      to write the response body. The IO function has access to the underlying
+      {!Async_unix.Reader.t} and {!Async_unix.Writer.t}, which allows writing a
+      response body more efficiently, stream a response or to switch protocols
+      entirely (e.g. websockets). Processing of pipelined requests continue
+      after the {!unit Async_kernel.Deferred.t} is resolved. The connection can
+      be closed by closing the {!Async_unix.Reader.t}. *)
 type response_action =
-  (* The connection is not closed in the [`Expert] case until the [unit
-     Async_kernel.Deferred.t] becomes determined. *)
   [ `Expert of Cohttp.Response.t
                * (Async_unix.Reader.t
                   -> Async_unix.Writer.t
--- a/cohttp-async/test/test_async_integration.ml
+++ b/cohttp-async/test/test_async_integration.ml
@@ -26,21 +26,33 @@ let server =
     (* large request *)
     (fun _ body ->
        body |> Body.to_string >>| String.length >>= fun len ->
-       Server.respond_string (Int.to_string len));
+       Server.respond_string (Int.to_string len) >>| response
+    )
   ] @ (* pipelined_chunk *)
   (response_bodies |> List.map ~f:(Fn.compose const ok))
-  @ [ (* large response chunked *)
-    (fun _ _ ->
-       let body =
-         let (r, w) = Pipe.create () in
-         let chunk = chunk chunk_size in
-         for _ = 0 to chunks - 1 do
-           Pipe.write_without_pushback w chunk
-         done;
-         Pipe.close w;
-         r
-       in
-       Server.respond_with_pipe ~code:`OK body
+  @
+  (* large response chunked *)
+  (fun _ _ ->
+     let body =
+       let (r, w) = Pipe.create () in
+       let chunk = chunk chunk_size in
+       for _ = 0 to chunks - 1 do
+         Pipe.write_without_pushback w chunk
+       done;
+       Pipe.close w;
+       r
+     in
+     Server.respond_with_pipe ~code:`OK body >>| response
+  )
+  :: [ (* pipelined_expert *)
+    expert (fun _ic oc ->
+      Async_unix.Writer.write oc "8\r\nexpert 1\r\n0\r\n\r\n";
+      Async_unix.Writer.flushed oc
+    );
+    expert (fun ic oc ->
+      Async_unix.Writer.write oc "8\r\nexpert 2\r\n0\r\n\r\n";
+      Async_unix.Writer.flushed oc >>= fun () ->
+      Async_unix.Reader.close ic
     )
   ]
   |> response_sequence
@@ -84,11 +96,21 @@ let ts =
       assert_equal Cohttp.Transfer.Chunked (Response.encoding resp);
       body |> Body.to_string >>| String.length >>| fun len ->
       assert_equal ~printer:(Int.to_string) (chunk_size * chunks) len in
+    let expert_pipelined () =
+      let printer x = x in
+      Client.get uri >>= fun (_rsp, body) ->
+      Body.to_string body >>= fun body ->
+      assert_equal ~printer "expert 1" body;
+      Client.get ~headers uri >>= fun (_rsp, body) ->
+      Body.to_string body >>| fun body ->
+      assert_equal ~printer "expert 2" body
+    in
     [ "empty chunk test", empty_chunk
     ; "large response", large_response
     ; "large request", large_request
     ; "pipelined chunk test", pipelined_chunk
     ; "large chunked response", large_chunked_response
+    ; "expert response", expert_pipelined
     ]
   end
 
--- a/cohttp-lwt-unix/src/io.ml
+++ b/cohttp-lwt-unix/src/io.ml
@@ -29,20 +29,35 @@ type conn = Conduit_lwt_unix.flow
 let src = Logs.Src.create "cohttp.lwt.io" ~doc:"Cohttp Lwt IO module"
 module Log = (val Logs.src_log src : Logs.LOG)
 
+let with_closed_handler f ~if_closed =
+  (* TODO Use [Lwt_io.is_closed] when available:
+     https://github.com/ocsigen/lwt/pull/635 *)
+  Lwt.catch f
+    (function
+    | Lwt_io.Channel_closed _ -> Lwt.return if_closed
+    | exn -> raise exn
+    )
+
 let read_line ic =
-  Lwt_io.read_line_opt ic >>= function
-  | None ->
-    Log.debug (fun f -> f  "<<< EOF");
-    Lwt.return_none
-  | Some l as x ->
-    Log.debug (fun f -> f  "<<< %s" l);
-    Lwt.return x
+  with_closed_handler ~if_closed:None
+    (fun () ->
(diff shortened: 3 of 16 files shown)


cohttp-lwt/src/s.ml
```diff
@@ -106,13 +106,27 @@ module type Server = sig
 
   type conn = IO.conn * Cohttp.Connection.t
 
+  type response_action =
+    (* The connection is not closed in the [`Expert] case until the [unit
+       Lwt.t] becomes determined. *)
+    [ `Expert of Cohttp.Response.t
+                 * (IO.ic
+                    -> IO.oc
+                    -> unit Lwt.t)
```
avsm: Should we also define what happens if this function leaks an exception?  Where will that be dispatched to?

cohttp-lwt/src/s.ml
```diff
@@ -106,13 +106,27 @@ module type Server = sig
...
+    | `Response of Cohttp.Response.t * Body.t ]
+
   type t
 
   val make : ?conn_closed:(conn -> unit)
     -> callback:(conn -> Cohttp.Request.t -> Body.t
                  -> (Cohttp.Response.t * Body.t) Lwt.t)
     -> unit -> t
 
+  val make_expert : ?conn_closed:(conn -> unit)
```
avsm: I'm wondering if this shouldn't eventually become the default function, so it shouldn't be called `make_expert`.  I suggest three functions:

`make_raw` which returns what `make_expert` does above
`make_expert` which wraps the response of `make_raw` in a `Expert` value
`make` which wraps the response of `make_raw` in a `Response` value

andreas: Is this what you mean?

```ocaml
val make_raw :
     ?conn_closed:(conn -> unit)
  -> callback:(conn -> Cohttp.Request.t -> Body.t -> response_action Lwt.t)
  -> unit
  -> t

val make_expert :
     ?conn_closed:(conn -> unit)
  -> callback:(   conn
               -> Cohttp.Request.t
               -> Body.t
               -> Cohttp.Response.t * (IO.ic -> IO.oc -> unit Lwt.t) Lwt.t)
  -> unit
  -> t

val make :
     ?conn_closed:(conn -> unit)
  -> callback:(   conn
               -> Cohttp.Request.t
               -> Body.t
               -> (Cohttp.Response.t * Body.t) Lwt.t)
  -> unit
  -> t
```

cohttp-lwt/src/s.ml
avsm: I agree - we should encourage the developer to treat leaking of exceptions as an internal server error, and not part of the published API.  We do need to be careful to generate an information-free 500 (i.e. not put the error in the return body of the 500, since that might leak sensitive information or be vulnerable to CRLF)

cohttp-lwt/src/s.ml
avsm: yep! Perhaps `make_raw` would be clearer as `make_response_action`?

cohttp-lwt/src/s.ml
andreas: Thinking about this again, it may not make sense write a 500 response to the client in case of an exception. If the function `IO.ic -> IO.oc -> unit Lwt.t` raises an exception, it's impossible to know what state the client/server are in. As an example, if the client and server have switched to websocket protocol, it would be incorrect to write a 500 HTTP response to the client. Bubbling the exception up to `on_exn` and closing the connection may be the best option.

Am I overlooking something?

cohttp-lwt/src/s.ml
andreas: Updated.

cohttp-lwt/src/s.ml
avsm: That's true. I'm fine with just logging and closing the connection. The more conservative the better in this situation :)

avsm:
This approach and interface lgtm!  Some tests to validate it stays working for websocket would be great before merging if possible.

mattjbray:
This solves my streaming use case (catch exceptions while writing the body) so :+1: from me. This will close #534 and #529.

avsm:
One option is to be explicit with the purpose of the response_action, and have:


```
type response_action =
  [ 
  `Switch_protocol of Cohttp.Response.t              
               * (IO.ic
                  -> IO.oc
                  -> unit Lwt.t)
 | `Stream of Cohttp.Response.t * Body.t 
              * (IO.ic
                  -> IO.oc
                  -> unit Lwt.t)
  | `Response of Cohttp.Response.t * Body.t ]
```

Although I'm a little confused as to the purpose of the empty body write in the Async case. What's that actually sending out to the wire -- an empty HTTP chunked encoding ?

andreas:
Anything I can do to help this move forward? If we can agree on the approach, I'd be happy to continue with implementation and tests (both Async and Lwt).

mseri:
I think that changing the semantics of ``` `Expert``` makes more sense and is cleaner and easier to reason about. We could ping @avsm to have also his opinion

avsm:
> Unless I'm missing something, I would suggest to change the semantics for `Expert to not write an empty response body.

Thank you for the comprehensive analysis! I agree with @andreas and @mseri that changing the semantics of Expert in makes more sense. We can merge and cut a major version bump with this change as soon as the PR is ready.

andreas:
Great, thanks! I'll proceed with the implementation and tests.

avsm:
@andreas when ready, if you merge with the master branch the Travis tests should now be green as well

cohttp-lwt/src/s.ml
```diff
@@ -106,12 +106,37 @@ module type Server = sig
 
   type conn = IO.conn * Cohttp.Connection.t
 
+  type response_action =
+    (* The connection is not closed in the [`Expert] case until the [unit
```
mseri: Can we add a proper docstring with the rationale for the response action and the content of this comment?

andreas: Good point -- added. Let me know what you think.

cohttp-async/src/server.mli
```diff
@@ -16,9 +16,16 @@ type 'r respond_t =
   ?body:Body.t ->
   Cohttp.Code.status_code -> 'r Async_kernel.Deferred.t
 
+(** A request handler can respond in two ways:
+    - Using [`Response], with a [Response.t] and a [Body.t].
```
avsm: Minor: these `[Response.t]` can be `{!Response.t}` and they will be hyperlinked by odoc (useful in the functor heavy world of cohttp :)

cohttp-async/test/test_async_integration.ml
```diff
@@ -40,7 +41,18 @@ let server =
          Pipe.close w;
          r
        in
-       Server.respond_with_pipe ~code:`OK body
+       Server.respond_with_pipe ~code:`OK body >>| response
+    )
+  ]
+  @ [ (* pipelined_expert *)
```
avsm: I think these `x @ y @ z` constructions are more efficient as `x :: y :: z` where possible, since you are constructing a single-element list and then using `@` (which isn't tail recursive)

cohttp-lwt/src/s.ml
mseri: Thanks. They look great

cohttp-async/src/server.mli
andreas: I updated all of the references now -- I don't know whether you meant only `[Response.t]`..

cohttp-async/test/test_async_integration.ml
andreas: Updated.

cohttp-lwt-unix/src/io.ml
```diff
@@ -29,20 +29,33 @@ type conn = Conduit_lwt_unix.flow
 let src = Logs.Src.create "cohttp.lwt.io" ~doc:"Cohttp Lwt IO module"
 module Log = (val Logs.src_log src : Logs.LOG)
 
+let with_closed_handler f ~if_closed =
+  Lwt.catch f
+    (function
+    | Lwt_io.Channel_closed _ -> Lwt.return if_closed
+    | exn -> raise exn
+    )
```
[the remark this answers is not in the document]
andreas: Added `TODO`.

cohttp-lwt/src/server.ml
```diff
@@ -70,74 +91,56 @@ module Make(IO:S.IO) = struct
...
+    | `No | `Unknown ->
+      `Empty
+
+  let handle_request callback conn req body =
+    Lwt.finalize
+      (fun () ->
+         Lwt.catch
+           (fun () -> callback conn req body)
+           (fun exn ->
+             Format.eprintf "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn);
```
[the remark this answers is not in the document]
andreas: `Logs` is currently not a dependency of `cohttp-lwt`, but I can add it if you want?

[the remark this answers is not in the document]
avsm: Yes I think that would be a good idea - we need to eliminate printf errors from embeddable libraries

cohttp-async/src/server.mli
avsm: All the references are great; it's useful for anything in the ocamldoc comment that is a linkable value, to give odoc more metadata to work with.

avsm:
When ready to merge, if you could push a CHANGES entry that describes the backwards compat breaking change that would be helpful.  I can roll several PRs into a major version bump.

andreas:
- Replaced `Format.eprintf` with `Log.err` in `Cohttp_lwt.Server`
- Added an entry `CHANGES.md`

avsm:
Thanks for the contribution! I'll prepare a release after a bit more testing -- in the meanwhile, ensuring that the master branch of cohttp works for your websocket/other libraries that need this feature would be most appreciated.

avsm:
That would be great! Having more examples to point to in the README is probably one of cohttp's bigger user requests :-)

avsm:
md2mld seems just fine to me -- especially if it had a dune rule that just worked :-)

The other place where the README shows up is in the `opam` description field.  Perhaps we should break up the README contents into a `doc/` directly and assemble it via dune promotion rules into the opam/README/mld files. I did a crude version of that in the [duniverse tool](https://github.com/avsm/duniverse/blob/master/dune).

Issue #624, opened 2018-06-27
Problem making a request to Zendesk that doesn't happen with curl


dwwoelfel:
I'm having a problem making a POST request to Zendesk.

If I use curl, everything works fine:

```
~ $ curl -X POST "https://dww-test.zendesk.com/oauth/tokens"
{"error":"invalid_request","error_description":"'client_id', 'grant_type' required."}
```

If I use cohttp, I get a 403 forbidden response from Zendesk:

```
~ $ CONDUIT_DEBUG=true COHTTP_DEBUG=true cohttp-curl-async -X POST "https://dww-test.zendesk.com/oauth/token
s"                                                                                                          
cohttp-curl-async: [DEBUG] 17946 >>> POST /oauth/tokens HTTP/1.1

cohttp-curl-async: [DEBUG] 17946 >>> connection: close
host: dww-test.zendesk.com
user-agent: ocaml-cohttp/1.1.0


cohttp-curl-async: [DEBUG] 17946 >>> 
cohttp-curl-async: [DEBUG] <<< HTTP/1.1 403 Forbidden
cohttp-curl-async: [DEBUG] <<< Server: cloudflare
cohttp-curl-async: [DEBUG] <<< Date: Wed, 27 Jun 2018 19:57:07 GMT
cohttp-curl-async: [DEBUG] <<< Content-Type: text/html
cohttp-curl-async: [DEBUG] <<< Content-Length: 167
cohttp-curl-async: [DEBUG] <<< Connection: close
cohttp-curl-async: [DEBUG] <<< CF-RAY: 431a7b1aae3a288e-SJC
cohttp-curl-async: [DEBUG] <<< 
cohttp-curl-async: [INFO] cf-ray: 431a7b1aae3a288e-SJC
cohttp-curl-async: [INFO] connection: close
cohttp-curl-async: [INFO] content-length: 167
cohttp-curl-async: [INFO] content-type: text/html
cohttp-curl-async: [INFO] date: Wed, 27 Jun 2018 19:57:07 GMT
cohttp-curl-async: [INFO] server: cloudflare
<html>
<head><title>403 Forbidden</title></head>
<body bgcolor="white">
<center><h1>403 Forbidden</h1></center>
<hr><center>cloudflare</center>
</body>
</html>
```

[...]

PR #648, opened 2018-11-14
Async: use Conduit V2 API

emillon:
This switches the Conduit API used by the Async backend. It would previously use the default V1 one.

In particular, the TLS configuration in V2 has access to the hostname, and when it is passed in, it will add a Server Name Indication (SNI) extension.

This has the following effects on a sample Client Hello:

```diff
- Extensions Length: 164
+ Extensions Length: 191
+ Extension: server_name (len=23)
  Extension: ec_point_formats (len=4)
  Extension: supported_groups (len=4)
  Extension: SessionTicket TLS (len=0)
```

This should take care of #624 (@dwwoelfel can you check that this works for you?).

That's technically a breaking change since `ssl_config` is using a V2 type, but that seems reasonable.

Thoughts?

Thanks!

### DIFF
--- a/cohttp-async.opam
+++ b/cohttp-async.opam
@@ -28,7 +28,7 @@ depends: [
   "base" {>= "v0.11.0"}
   "core" {test}
   "cohttp"
-  "conduit-async" {>="1.0.3"}
+  "conduit-async" {>="1.2.0"}
   "magic-mime"
   "logs"
   "fmt" {>="0.8.2"}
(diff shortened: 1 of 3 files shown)


avsm:
lgtm; we could batch this change with #647 for a major bump due to the interface changes

avsm:
I've pushed a merge against master to the branch to fix CI (#649)

emillon:
Thanks for the merge!

cc @sgrove

PR #652, opened 2019-01-11
log invalid client input and do not catch out of memory

hannesm:
- log invalid input from client in handle_client
 - in handle_request, do not catch Out_of_memory exception

the reasoning behind this change is that I'm eager to see when something goes wrong, and since `Request.read` already takes care of building a string with an error message, we should display this somewhere :)

for the second part, not catching out of memory, this is likely a matter of taste. there are situations where aborting/closing a single client connection may recover sufficient memory to get cohttp into a working state again.

but in reality, what I see with Canopy is that something in the stack (likely cohttp/conduit, don't see this in other unikernels) leaks memory, now a lot of connections are running into the above catch, and get an 500 back, cohttp reclaims some more memory... this goes on for _days_ until some other piece of code tries to allocate just enough memory to be over the limit, and crashes the unikernel.

for several days, the state is that HTTP(S) is up, but nearly always responds with 500 -- this PR changes the behaviour and have cohttp crash hard if there's an out of memory. now, certainly this opens an attack vector: an attacker that can provoke `callback` to allocate lots of memory will crash the server. please keep in mind that when callback is called, cohttp already parsed the client input (which may be of arbitrary size(?)) into a Cohttp.Request.t structure.

depending on your point of view, I'd as well be glad if the desired behaviour would be a configuration option (for Server.create).

### DIFF
--- a/cohttp-lwt/src/server.ml
+++ b/cohttp-lwt/src/server.ml
@@ -109,16 +109,20 @@ module Make(IO:S.IO) = struct
       (fun () ->
          Lwt.catch
            (fun () -> callback conn req body)
-           (fun exn ->
-             Log.err (fun f -> f "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn));
-             respond_error ~body:"Internal Server Error" () >|= fun rsp ->
-             `Response rsp
+           (function
+             | Out_of_memory -> raise Out_of_memory
+             | exn ->
+               Log.err (fun f -> f "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn));
+               respond_error ~body:"Internal Server Error" () >|= fun rsp ->
+               `Response rsp
            ))
       (fun () -> Body.drain_body body)
 
   let rec handle_client ic oc conn callback =
     Request.read ic >>= function
-    | `Eof | `Invalid _ -> (* TODO: request logger for invalid req *)
+    | `Eof -> Lwt.return_unit
+    | `Invalid data ->
+      Log.err (fun m -> m "invalid input %s while handling client" data);
       Lwt.return_unit
     | `Ok req ->
       begin let body = read_body ic req in
--- a/cohttp-lwt/src/server.ml
+++ b/cohttp-lwt/src/server.ml
@@ -110,7 +110,7 @@ module Make(IO:S.IO) = struct
          Lwt.catch
            (fun () -> callback conn req body)
            (function
-             | Out_of_memory -> raise Out_of_memory
+             | Out_of_memory -> Lwt.fail Out_of_memory
              | exn ->
                Log.err (fun f -> f "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn));
                respond_error ~body:"Internal Server Error" () >|= fun rsp ->


cohttp-lwt/src/server.ml
```diff
@@ -109,16 +109,20 @@ module Make(IO:S.IO) = struct
       (fun () ->
          Lwt.catch
            (fun () -> callback conn req body)
-           (fun exn ->
-             Log.err (fun f -> f "Error handling %a: %s\n%!" Request.pp_hum req (Printexc.to_string exn));
-             respond_error ~body:"Internal Server Error" () >|= fun rsp ->
-             `Response rsp
+           (function
+             | Out_of_memory -> raise Out_of_memory
```
mseri: As much as I agree with your rationale, what if we make this change optional with a new optional parameter?

Also why have you decided to use `raise` instead of `Lwt.fail`, it makes less allocations in such pressure situation?

hannesm: @mseri as mentioned above, I'd be ok with this behaviour being optional. Oh, I wasn't aware whether `raise` or `Lwt.fail` would be preferred. If it is the latter, that's fine with me as well.

hannesm:
I'm still not sure when the current behaviour is preferable:
- let's consider a perfect universe where no memory leaks happen: if cohttp accepts too many connections so that it runs out of memory, a random one will receive a 500, memory will be reclaimed, and other connections are working fine
- in a slightly less perfect universe (something leaks), the behaviour described in my earlier comment is showing up. this delays the crash for quite some time, but in the medium the responses from cohttp are pretty unreliable - lots of 500 are served to random clients

NB: to avoid crashes in a non-leaking universe, another strategy would be to allow only x concurrent client connections, and not `accept` any more unless one client connection finished.  this is something configurable in other web servers as well (and esp. on UNIX where the amount of open file descriptors of each process is limited, a useful configuration setting).

mseri:
Maybe my main worry of maintaining backward behavioural compatibility in this case is misplaced.

I remember having had issues when conduit/cohttp exception handling changed and it took a long time to figure out what happened and figure out that we needed to modify manually the exception handler. In this case, a not too reliable service, may have resisted without issues for long (maybe because there were sort-of periodic reboots due to upgrades, maybe for other reasons) and then suddenly starts breaking in production. Even if I agree that it starts breaking for the right reason, I was worried about how it could affect some long-time users.

cohttp-lwt/src/server.ml
mseri: To be honest I don't know. `Lwt.catch` will catch both. I tend to use `Lwt.fail` only because wraps the failure in a `Lwt` promise and composes better.

hcarty:
I'd really like to have `Out_of_memory` escape either always or as an option.  I've been trying to track down a potentially similar memory leak @hannesm which also seems to be tied to cohttp/conduit but the internal catching of `Out_of_memory` has made it much more difficult to track what's happening.

hannesm:
I updated the code to use `Lwt.fail` instead of `raise` as suggested by @mseri.

hannesm:
i tried to look up the semantics of cohttp-async in this respect, but don't understand it entirely -- there's no exception handler involved afaict, so what happens if the user-provided callback raises an exception? (here, in the lwt server such exceptions are caught)

now, I wondered what the least surprising semantics would be ([POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)). IMHO, this would be to not install any exception handler in the cohttp-lwt server. if this in addition uniforms the cohttp-server semantics on both async and lwt, I'd be in favour.

avsm:
I think that not catching `Out_of_memory` is a good idea, but I don't think that calling `Lwt.fail` in the re-raise is wise.  `Lwt.fail` allocates in a low-memory situation...

avsm:
I've reviewed again and think my comment above can be dealt with separately. I'll merge this shortly.

mseri:
Oops, I thought we had merged this already. Sorry

avsm:
No problem! I'm sweeping through cutting some releases so ran across it :-)

PR #654, opened 2019-02-03
use namespaced jsoo interfaces

avsm:
also clean up the opam file to remove redundant conflicts


avsm:
This unbreaks CI in trunk as well

PR #655, opened 2019-02-03
use base64-3.1.0 interface and remove some opam file redundancy

avsm:
This will fail until ocaml/opam-repository#13381 is merged


mseri:
It is good now

PR #656, opened 2019-02-03
Remove compat shims in META file

avsm:


avsm:
I've bumped the major to 2.0 to reflect the compat break and make it easier to add constraints

PR #661, opened 2019-03-27
Fix for async v0.12.0

copy:
If I'm not mistaken, this is the only bit necessary to compile with async v0.12.0.


avsm:
Thanks! This should be backwards compatible too, handily

mseri:
Thanks!

PR #664, opened 2019-04-09
fix warning with re.1.9.0

vbmithr:


avsm:
Does this also impose a minimum restriction on the `re` opam version?

vbmithr:
Yes, re >= 1.9.0

avsm:
Thanks; would you mind adding the corresponding constraint into cohttp.opam, and we can merge this then

avsm:
thanks!

PR #667, opened 2019-07-05
Add uri-sexp as dependency (DON'T MERGE)

dinosaure:


avsm:
Shouldnt be needed any more now that uri-sexp is upstream in opam-repo

Issue #645, opened 2018-10-31
Expose custom Logs module(s) and/or source


hcarty:
The move to using Logs for logging is wonderful!  However, it makes it almost impossible to use debug-level logging in an application using cohttp as cohttp does not expose its custom logging source.  That means that an application passing around large blob of data with cohttp ends up dumping all of those bytes as log messages when an application's log verbosity is set to `debug`.

It would be useful to expose enough of the internal logging configuration to allow an application to explicitly set cohttp's log level separately from the rest of the application.

osener:
Is there a way to silence Cohttp's log messages or redirect them somewhere else? I use Logs library in my application as well, and Cohttp's log messages make it very hard to follow what's going on during development in a terminal emulator.

mseri:
If you see cohttp logging you probably set the `debug` level for all sources.
In this case, you can mute all cohttp logging as follows (remember to run this each time you set the log level with `~all:true` and don't use those as your sources names).

```ocaml
List.iter (fun src ->
      let name = Logs.Src.name src in
      if List.mem name [ "cohttp.lwt.io"; "cohttp.lwt.server" ] then
        Logs.Src.set_level src None)
  @@ Logs.Src.list ();
```

mseri:
I think we should document this better, and provide those labels somewhere, but I have experimented with it for a bit and I wouldn't really know how else to change the API.

mseri:
I am going to close this issue, if there are further ideas for improvements feel free to reopen it or start a discussion in https://github.com/mirage/ocaml-cohttp/discussions

PR #669, opened 2019-07-18
Better handling of IO errors

### LINKED ISSUE #511: cohttp server crash when client dies mid-transfer
The following server will crash (usually) if the `wrk` process is canceled with a `ctrl-c`.  I've seen at least `EPIPE`, `ECONNRESET`, `EPROTOTYPE` errors while testing this.

This may be related to #503

Using OCaml 4.02.3, cohttp 0.21.0.  Tested on 64bit OSX and Linux with `libev`/`conf-libev` installed.

Server code:

```ocaml
module Server = Cohttp_lwt_unix.Server

let callback _conn _req body =
  let%lwt body = Cohttp_lwt_body.to_string body in
  Server.respond_string ~status:`OK ~body ()

let server port =
  Server.create ~mode:(`TCP (`Port port)) (Server.make ~callback ())

let () =
  Lwt_engine.set (new Lwt_engine.libev);
  Lwt_main.run (server (int_of_string Sys.argv.(1)))
```

Client using `wrk`:
```
wrk -s put.lua 'http://127.0.0.1:8080/'
```

The `put.lua` script creates a blob of bytes to send as a `PUT` request to the server.

`put.lua`:
```lua
local Chars = {}
for Loop = 0, 255 do
   Chars[Loop+1] = string.char(Loop)
end
local String = table.concat(Chars)

local Built = {['.'] = Chars}

local AddLookup = function(CharSet)
   local Substitute = string.gsub(String, '[^'..CharSet..']', '')
   local Lookup = {}
   for Loop = 1, string.len(Substitute) do
       Lookup[Loop] = string.sub(Substitute, Loop, Loop)
   end
   Built[CharSet] = Lookup

   return Lookup
end

function string.random(Length, CharSet)
   -- Length (number)
   -- CharSet (string, optional); e.g. %l%d for lower case letters and digits

   local CharSet = CharSet or '.'

   if CharSet == '' then
      return ''
   else
      local Result = {}
      local Lookup = Built[CharSet] or AddLookup(CharSet)
      local Range = table.getn(Lookup)

      for Loop = 1,Length do
         Result[Loop] = Lookup[math.random(1, Range)]
      end

      return table.concat(Result)
   end
end

wrk.method = "PUT"
wrk.body   = string.random(1000000, "%l%d")
wrk.headers["Content-Type"] = "text/plain"
```

talex5:
Before, if the client closed the connection while we were handling a request, the server would crash (by default, unless the user overrode that using `on_exn` or changing `Lwt`'s async exception handler). Now, we just log this at `info` level and continue. Exceptions produced by user code are logged as errors, while other exceptions generated by cohttp call back to the conduit exception handler, as before.

Fixes #511.

I first tried getting the IO operations to return a `result` and plumbing that through everywhere, but it was very painful and caused lots of API changes. So instead, each IO monad is now expected to include handling of IO errors, and provides a `catch` operation to expose it. This avoids changes to the core of cohttp. The lwt-unix and mirage implementations both make use of `Lwt.fail` to carry IO errors.

This PR adds a unit-test to check that the server doesn't crash if the client closes the connection. It also adds checks to some other tests to ensure that no errors or warnings were logged during the test.

### DIFF
--- a/cohttp-lwt-unix/src/io.ml
+++ b/cohttp-lwt-unix/src/io.ml
@@ -14,7 +14,13 @@
  *
   }}}*)
 
+exception IO_error of exn
+
 let () =
+  Printexc.register_printer (function
+      | IO_error e -> Some ("IO error: " ^ Printexc.to_string e)
+      | _ -> None
+    );
   if Sys.os_type <> "Win32" then
     Sys.(set_signal sigpipe Signal_ignore);
 
@@ -29,17 +35,25 @@ type conn = Conduit_lwt_unix.flow
 let src = Logs.Src.create "cohttp.lwt.io" ~doc:"Cohttp Lwt IO module"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let with_closed_handler f ~if_closed =
+let wrap_read f ~if_closed =
   (* TODO Use [Lwt_io.is_closed] when available:
      https://github.com/ocsigen/lwt/pull/635 *)
   Lwt.catch f
     (function
-    | Lwt_io.Channel_closed _ -> Lwt.return if_closed
-    | exn -> raise exn
+      | Lwt_io.Channel_closed _ -> Lwt.return if_closed
+      | Unix.Unix_error _ as e -> Lwt.fail (IO_error e)
+      | exn -> raise exn
+    )
+
+let wrap_write f =
+  Lwt.catch f
+    (function
+      | Unix.Unix_error _ as e -> Lwt.fail (IO_error e)
+      | exn -> raise exn
     )
 
 let read_line ic =
-  with_closed_handler ~if_closed:None
+  wrap_read ~if_closed:None
     (fun () ->
       Lwt_io.read_line_opt ic >>= function
       | None ->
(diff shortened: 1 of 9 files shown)


cohttp-lwt-unix/src/io.ml
```diff
@@ -29,17 +35,25 @@ type conn = Conduit_lwt_unix.flow
 let src = Logs.Src.create "cohttp.lwt.io" ~doc:"Cohttp Lwt IO module"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let with_closed_handler f ~if_closed =
+let wrap_read f ~if_closed =
   (* TODO Use [Lwt_io.is_closed] when available:
      https://github.com/ocsigen/lwt/pull/635 *)
```
avsm: Does this TODO make sense to implement (or delete) here?  Lwt_io.closed has been available since lwt.4.2.0

talex5: Probably, but I think that should be a separate PR.

avsm:
Tested on github and rwo's regression tests also -- thanks for this fix!

hcarty:
Thank you for this! I think this makes #645 even more important, so these log messages can be controlled separately by the application.

PR #671, opened 2019-07-24
Async client: do not read body if there is none

### LINKED ISSUE #666: 204 causes cohtt-async to hang
It looks like all of the informational responses mentioned here https://github.com/mirage/ocaml-cohttp/pull/355 cause cohttp-async client to hang until the server times out the connection:

```reason
let testCall = (code: int) => {
  let start = Unix.time();

  Cohttp_async.Client.get(
    Uri.of_string(Format.sprintf("https://httpstat.us/%d", code)),
  )
  >>= (
    ((resp, body)) => {
      let status =
        resp |> Cohttp.Response.status |> Cohttp.Code.string_of_status;
      let debug =
        Format.asprintf(
          "Status: %s\nFull response:\n%a\n",
          status,
          Cohttp.Response.pp_hum,
          resp,
        );

      Format.printf("Cohttp result:\n%s\n", debug);
      Cohttp_async.Body.to_string(body);
    }
  )
  >>| (
    _body => {
      let duration = Unix.time() -. start;
      Format.printf("Request took: %fs\n", duration);
    }
  );
};

/* All these should complete instantly */
let testWorking = () => {
  testCall(200)
  >>= (_ => testCall(201))
  >>= (_ => testCall(300))
  >>= (_ => testCall(301))
  >>= (_ => testCall(400))
  >>= (_ => testCall(401))
  >>= (_ => testCall(500));
};

/* Each of these will hang */
let testFailing = () => {
  testCall(204);
  testCall(304);
  testCall(100);
  testCall(101);
  testCall(102);
};

let command =
  Async.Command.async_spec(
    ~summary="",
    Command.Spec.empty,
    () => {
      testWorking();
      testFailing();
    },
  );

Command.run(command);
```

In the cases where responses include a `content-length` header, the client works as expected:
```
// Has a content-length: 0 header so the client is fine:
testCall(200);
Cohttp result:
Status: 200 OK
Full response:
((encoding (Fixed 0))
 (headers
  ((access-control-allow-origin *) (cache-control private) (content-length 0)
   (date "Sun, 05 May 2019 22:05:48 GMT") (server Microsoft-IIS/10.0)
   (set-cookie
    "ARRAffinity=580ecdc04582a8c27680686941822edec8921fe0a4244b99f898ac1d3a4de89a;Path=/;HttpOnly;Domain=httpstat.us")
   (x-aspnet-version 4.0.30319) (x-aspnetmvc-version 5.1)
   (x-powered-by ASP.NET)))
 (version HTTP_1_1) (status OK) (flush false))

Request took: 0.000000s
- : unit = ()
```

But it seems all of the informational responses (correctly) do not include a `content-length` header, and this causes the client to hang:

```
// Note the lack of content-length header, and that the request took 129 seconds
testCall(204);
Cohttp result:
Status: 204 No Content
Full response:
((encoding Unknown)
 (headers
  ((access-control-allow-origin *) (cache-control private)
   (date "Sun, 05 May 2019 22:07:18 GMT") (server Microsoft-IIS/10.0)
   (set-cookie
    "ARRAffinity=580ecdc04582a8c27680686941822edec8921fe0a4244b99f898ac1d3a4de89a;Path=/;HttpOnly;Domain=httpstat.us")
   (x-aspnet-version 4.0.30319) (x-aspnetmvc-version 5.1)
   (x-powered-by ASP.NET)))
 (version HTTP_1_1) (status No_content) (flush false))

Request took: 129.000000s
- : unit = ()
```

Presumably the client is waiting to read more input from the server, until the server times out the connection.

I think this has been the case for awhile - one of the services we were interacting with returns lots of 204 responses, but previously they included a `content-length` header. Once they removed that (presumably to be "more correct") we started hitting this issue, and it seems to be reproducible across several services.

emillon:
Responses with code 204 (No content) and 304 (Not modified) do not have
bodies. They can come with no Content-Length header, in which case the
transfer encoding is inferred as `Transfer.Unknown`. When trying to read
the nonexistent body, it would do so using `Transfer_io.Unknown.read`,
which attempts to read until EOF, which can cause hangs.

There is some logic in `Response.has_body` to determine if we should
attempt to read a body at all. This was hooked in the `Lwt` backend but
not in the `Async` one.

Closes #666

### DIFF
--- a/cohttp-async/src/client.ml
+++ b/cohttp-async/src/client.ml
@@ -50,10 +50,17 @@ let read_request ic =
   | `Eof -> failwith "Connection closed by remote host"
   | `Invalid reason -> failwith reason
   | `Ok res ->
-    (* Build a response pipe for the body *)
-    let reader = Response.make_body_reader res ic in
-    let pipe = Body_raw.pipe_of_body Response.read_body_chunk reader in
-    (res, pipe)
+    begin
+      match Response.has_body res with
+      | `Yes | `Unknown ->
+        (* Build a response pipe for the body *)
+        let reader = Response.make_body_reader res ic in
+        let pipe = Body_raw.pipe_of_body Response.read_body_chunk reader in
+        (res, pipe)
+      | `No ->
+        let pipe = Pipe.of_list [] in
+        (res, pipe)
+    end
 
 let request ?interrupt ?ssl_config ?uri ?(body=`Empty) req =
   (* Connect to the remote side *)
--- a/cohttp-async/src/client.ml
+++ b/cohttp-async/src/client.ml
@@ -45,7 +45,7 @@ module Net = struct
     Conduit_async.V2.connect ?interrupt mode
 end
 
-let read_request ic =
+let read_response ic =
   Response.read ic >>| function
   | `Eof -> failwith "Connection closed by remote host"
   | `Invalid reason -> failwith reason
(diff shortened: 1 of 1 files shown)


emillon:
Not sure how to write an automated test for this, but this fixes the 204 & 304 cases in #666.

avsm:
This lgtm, thanks!

PR #680, opened 2019-12-09
Support async and base >= v0.13

copy:
Addresses the following changes in async and base v0.13:

- `Async_extra.Tcp` has been moved to `Async_unix.Tcp`
- `Async_extra.Command` has been moved to `Async_command` (part of the `async` opam package)
- `Base.sexp_opaque` has been deprecated


mseri:
Looks like cohttp-async needs a lower bound on ocaml 4.08 (and the other tests in travis matrix need to be dropped). The lwt jsoo module stopped compiling, is it a consequence of these changes (may be due to changes in other libraries)?

copy:
I have updated the test matrix to use 4.08 on all distributions.

I didn't add a lower bound on OCaml 4.08 since cohttp-async itself works fine below 4.08, only its dependencies don't. Slightly unfortunate, but at least this only affects cohttp-async.

The jsoo failure seems related to this change: https://github.com/ocsigen/js_of_ocaml/pull/863

mseri:
Thanks

mseri:
@avsm do we need to keep supporting old version of ocaml with cohttp-async for some reason or this can be merged?

PR #684, opened 2019-12-31
Async: Use Pipe.empty over of_list for empty pipes

msaffer:
The current implementation of [Pipe.empty] is to call [of_list] on an
empty list, however this may change in the future.


mseri:
It is failing with
```
# File "cohttp-async/src/body_raw.ml", line 39, characters 14-24:
3191# 39 |   | `Empty -> Pipe.empty ()
3192#                    ^^^^^^^^^^
3193# Error: Unbound value Pipe.empty
```

PR #685, opened 2020-01-06
Add ppx_compare to provide comparison functions for Response.t

msaffer-js:


msaffer-js:
I'll fix this and reopen.

PR #692, opened 2020-04-22
Conduit 3.0.0

dinosaure:
A draft about a move to `conduit.3.0.0`. For some details, see mirage/ocaml-conduit#311.

Currently, the new version of `conduit` comes with 2 major updates (just about implementation):
- the global loop is not defined by `conduit` anymore. Implementation must write it - and take care about how to finish the loop, how to close connections, etc.
- the `ctx` is, conceptually, replaced by `resolvers` (available into the `conduit` package) and must be filled by the user to be able to resolve something. The previous version of `conduit` has a side-effect on the `ctx` and can be filled in many ways - specially by application of _functor_ in the context of `mirage`. Now `conduit` gives only an _empty_ `resolvers` and it must be filled by the user at the end - as I do in this PR.

Of course, 2 others updates appear:
- the ability to destruct a flow to a concrete type from a given protocol implementation (which is define by the user only)
- the ability to compose with `conduit-tls` (which is outside the scope of `conduit-lwt` or `conduit-lwt-unix`, it depends only on `conduit) a protocol with TLS (and `ocaml-tls`). Examples are done into binaries provided by `cohttp-lwt-unix`

It still a draft and I'm currently on:
- [x] `mirage` support
- [x] `async` support
- [x] `js_of_ocaml` support


dinosaure:
`mirage` support is done. However, an orthogonal issue exists where `conduit` decides to update the `mirage-flow` interface to let the user to allocate the input buffer. `conduit-mirage` provides a `mirage-flow` interface as expected by `mirage-channel` - however it seems that `mirage-channel` is used only by `cohttp`.

We should delete the use of `mirage-channel` (or update it according the new given interface of the flow provided by `conduit`).

dinosaure:
Trust on `ocaml-ci` instead Travis CI (much more faster).

dinosaure:
Ok, `ocaml-ci` is happy! It failed before due to an overlap of cohttp servers on the same port `9193` but it fixed now with the last commit. I think, the diff is small as I can. Please someone to review this PR (@mseri, @avsm) I'm not really confident with `cohttp-async` when I don't use it.

It still is a draft when we need to add a submodule to be able to compile correctly `cohttp`. Once `conduit` is released, I will clean this PR.

mseri:
I am going to review it next week, as soon as I can.

Note for myself: before merge

- [ ] update the examples in the README and any other snippet around if needed (we should really test these...)
- [ ] make a new release of cohttp before 3.0, if not already done

mseri:
I will likely merge a few PRs before the next release, including https://github.com/mirage/ocaml-cohttp/pull/698 or a variation of it, sorry for the potential merge conflicts

dinosaure:
Yes, of course, I can make a pass then to avoid any conflicts.

dinosaure:
The PR was rebased with the new trunk, however @avsm wants to look something about `conduit`/`cohttp`.

mseri:
Thanks. Can you point me to the documentation and design of conduit 3, so that I can study a bit before embarking in the review?

dinosaure:
Currently, the `README.md` of `conduit` is not well updated to the new interface but a global view is available [here](https://github.com/dinosaure/blog.x25519.net/blob/master/html/posts/2020-02-27--tuyau.org) if you want - note that I will make something more accurate with the new interface of `conduit`.

cohttp-async/src/client.ml
```diff
@@ -28,24 +28,44 @@ module Net = struct
         Or_error.return (host, Ipaddr_unix.of_inet_addr addr, port)
       | _ -> Or_error.error "Failed to resolve Uri" uri Uri_sexp.sexp_of_t
 
-  let connect_uri ?interrupt ?ssl_config uri =
-    (match Uri.scheme uri with
-    | Some "httpunix" ->
+  let connect_uri ?interrupt:_ ?ssl_ctx uri =
```
mseri, line 31: What was `interrupt` for? Is there a reason for keep it there unused?

cohttp-async/src/server.ml
```diff
@@ -152,36 +143,50 @@ let respond_with_file ?flush ?headers ?(error_body=error_body_default) filename
...
+        Conduit_async.cast flow Conduit_async_ssl.TCP.protocol with
+  | Some flow, None ->
+    Async.return (Conduit_async.TCP.Protocol.reader flow,
+                  Conduit_async.TCP.Protocol.writer flow)
+  | None, Some flow ->
+    let { Conduit_async_ssl.reader; writer; _ } = flow in
+    Async.return (reader, writer)
+  | _ -> Conduit_async.reader_and_writer_of_flow flow
+
+let create_raw ?max_connections:_ ?backlog:_ ?buffer_age_limit:_ ~protocol ~service
```
mseri, line 157: Similarly as in the other case, the new conduit is dropping `max_connections`. `backlog` and `buffer_age_limit`. Is this something to worry about? Is there a reason for keep it here if not?

cohttp-lwt-unix/src/net.mli
```diff
@@ -18,19 +18,21 @@
...
 
-val init : ?ctx:Conduit_lwt_unix.ctx -> ?resolver:Resolver_lwt.t -> unit -> ctx
+val empty : resolvers
 
-val default_ctx : ctx
+(** Exceptions from [conduit].
+
+    [conduit] can returns error about [recv] and [send] processes.
+    These exceptions are raised when one of these {i syscall} returns
+    an error. *)
```
mseri, line 29: I would reword this as follows, or something similar.

```
When the [recv] or the [send] {i syscalls} return an error,
[conduit] will reraise it.
```

cohttp-lwt-unix/src/server.ml
```diff
@@ -67,7 +67,29 @@ let log_on_exn =
...
+  : type cfg t flow.
+     ?timeout:int
+  -> ?backlog:int
+  -> ?stop:unit Lwt.t
+  -> ?on_exn:(exn -> unit)
+  -> cfg
+  -> (_, flow) Conduit_lwt.protocol
+  -> (cfg, t, flow) Conduit_lwt.Service.service
+  -> _ -> unit Lwt.t
+  = fun ?timeout:_ ?backlog:_ ?(stop= fst (Lwt.wait ())) ?(on_exn=log_on_exn)
```
mseri, line 85: As before, we are ditching `timeout` and` backlog`. Is this something to worry about? Should we just remove them?

examples/async/hello_world.ml
```diff
@@ -22,9 +21,10 @@ let handler ~body:_ _sock req =
 let start_server port () =
   Caml.Printf.eprintf "Listening for HTTP on port %d\n" port;
   Caml.Printf.eprintf "Try 'curl http://localhost:%d/test?hello=xyz'\n%!" port;
-  Cohttp_async.Server.create ~on_handler_error:`Raise
-    (Async.Tcp.Where_to_listen.of_port port) handler
-  >>= fun _ -> Deferred.never ()
+  let _never, server = Cohttp_async.Server.create ~on_handler_error:`Raise
+    ~protocol:Conduit_async.TCP.protocol ~service:Conduit_async.TCP.service
+    (Conduit_async.TCP.Listen (Async.Tcp.Where_to_listen.of_port port)) handler in
+  server
```
mseri, line 27: How is this thing of `_never, server`, returning `server, instead of throwing away everything with `fun _ -> Deferred.never ()` working?

mseri (commented):
I have done a first superficial review. Sorry it took so long. A blog post on the server side of conduit would be very useful!

dinosaure:
Thanks for your report.

> What was interrupt for? Is there a reason for keep it there unused?

> Similarly as in the other case, the new conduit is dropping max_connections. backlog and buffer_age_limit. Is this something to worry about? Is there a reason for keep it here if not?

> As before, we are ditching timeout and backlog. Is this something to worry about? Should we just remove them?

So all of these questions point to a _regression_ about what it should be done by `conduit` and `cohttp`.

Come back to the rewrite of `conduit`, I discovered an hard effort to provide [a `server_with_handler`](https://github.com/mirage/ocaml-conduit/blob/master/lwt-unix/conduit_lwt_unix.mli#L184) close to what [Async](https://ocaml.janestreet.com/ocaml-core/latest/doc/async_unix/Async_unix/Tcp/Server/index.html) (see `create`) provides. . Conduit tried to provide the same logic on both sides (`conduit-lwt` and `conduit-async`).

On other side, it seems clear that we can not /mimic/ (or reproduce) the same behavior than `async` on `conduit-lwt` with such complexity (many arguments and all of them slightly change our _loop_). A main difference between `lwt` and `async` is highlighted by [this required `wait`](https://github.com/mirage/ocaml-conduit/blob/9815aa091e9c285e236e1deffb0dc48723ac72a0/tests/ping-pong/with_async.ml#L67) on conduit's tests.

From the perspective of `conduit`, I would not provide a common and complex way to start a server loop (even if [I added](https://github.com/mirage/ocaml-conduit/blob/9815aa091e9c285e236e1deffb0dc48723ac72a0/src/lwt/conduit_lwt.ml#L49) [it!](https://github.com/mirage/ocaml-conduit/blob/9815aa091e9c285e236e1deffb0dc48723ac72a0/src/async/conduit_async.ml#L18)) to let the user to really do what he wants (and take the responsibility on it).

> A blog post on the server side of conduit would be very useful!

So more concretely, `conduit` provides what it should be expected by the user to create its own server loop ([`bind` & `listen`, `accept` and `close`](https://github.com/mirage/ocaml-conduit/blob/9815aa091e9c285e236e1deffb0dc48723ac72a0/src/core/conduit.mli#L316)). `Service` uses the abstraction as `Protocol` but all types are well-known (this is the main difference from the old version of `conduit`) and can be concretely constructed/destructed.

Finally, I would like to not provide more than this to still be homogeneous over `lwt`, `async` and `mirage` - and [some tricks](https://github.com/mirage/ocaml-conduit/blob/9815aa091e9c285e236e1deffb0dc48723ac72a0/src/mirage/conduit_mirage_tcp.ml#L230) are required to ensure this homogenization between these back-ends.

Recently, some others issues were discovered in mirage/irmin#1062 and mirage/irmin#1066 where it's really hard to figure out at which layer these bugs appear. This recent misadventure gives me a more strong opinion about this question. HTTP/AF gives an other opinion where the server loop can be more complex than what we can naively provide - and it depends on which protocol we want to implement.

So about `cohttp`, I think the best is to shift the complexity of the server loop into `cohttp`. More concretely:
- `backport` is the `capacity` or in other words, the number given to `listen`
- `timeout` is a `sleep` and if the `accept` take more than `timeout`, we shutdown the server
- `max_connections` appears only on the `async` side
  However, the current old version of `conduit-lwt-unix` can limit the number of clients with `set_max_active` (which is not used by `cohttp-lwt-unix`).
- `on_exn` (`cohttp-lwt`)/`on_handler_error` (`conduit-async`) could be add into `conduit`
- `buffer_age_limit` exists only on the `async` side
- `interrupt` exists only on the `async` side (only the client part)

So I can try to solve these issues and provide something on `cohttp` (on this PR) to still continue to handle these options. However, we should decide what it should be common and deleted between `cohttp-async` and `cohttp-lwt` (@avsm?) - or assert a difference between `cohttp-async` and `cohttp-lwt` (in that case, `cohttp-async` should directly use `Async_unix.Tcp` instead to use `conduit-async`).

dinosaure:
So I updated my PR such as:
- `?backlog` is used __only__ when the user wants to start a TCP connection. The integer is given to `listen` properly as before. However, we need to clarify a point: `?backlog` leads the configuration.
- `?timeout` is added too into `conduit-{lwt,async}`. I took the opportunity to /homogenize/ the API between `cohttp-async` and `cohttp-lwt-unix`, both have the `?timeout` argument
- `?interrupts`/`?max_connections`/`?buffer_age_limit` are too close to `Async`, so I removed them (only into `cohttp-async` of course). As I said, we can consider this update as a _regression_ however: if it's really required, I can propose an other PR which will implement such arguments for both _back-ends_

Finally, an error occurs on 32-bits architectures but it's unrelated to this PR. I will clean the PR (about history) and I think it will be ready to merge. If you have any others feedbacks @mseri, tell me :+1:.

- [ ] clean the PR

mseri:
Thanks for the update and the clarifications. I agree with your comments and look forward hearing the opinion of other users and maintainers.

I have just a small request and two questions.

- [ ] do the examples (https://github.com/mirage/ocaml-cohttp#client-tutorial) in the tutorial need fixing to work in the new cohttp word? (I cannot create the switch to check it properly at the moment, otherwise I would have looked at them myself)
- [ ] do you think that the new conduit and this revamp of cohttp can solve the issue pointed out here: https://github.com/mirage/ocaml-cohttp/pull/704?
- [ ] do think that with you the new conduit and this revamp of cohttp we can also address the issue pointed out here: https://github.com/mirage/ocaml-cohttp/pull/696?

dinosaure:
> do the examples (https://github.com/mirage/ocaml-cohttp#client-tutorial) in the tutorial need fixing to work in the new cohttp word? (I cannot create the switch to check it properly at the moment --and for a while unfortunately--, otherwise I would have looked at them myself)

Surely, yes. If I take the example as is, `conduit` will be not able to resolve anything (due to `Conduit.empty`). A solution, and I think @avsm agree with that, is to define `Net.empty` with, at least, the host-system DNS resolver to be able, at least, to create a TCP/IP (`*:80`) connection. I don't have a strong opinion with that so I can add it into `cohttp-async` and `cohttp-lwt`.

About `cohttp-async` and `cohttp-lwt-__unix__`, we can extend the resolution (`Conduit.resolvers`) with an SSL support (`*:443`) with `conduit-{async,lwt}-ssl` (in anyway, `conduit-async` and `conduit-lwt-unix` depend on these packages).

- [ ] add a default `resolvers` which include resolution of TCP connection and SSL connection

> do you think that the new conduit and this revamp of cohttp can solve the issue pointed out here: #704?

From my perspective, the ability to use something else (defined by the user) than a simple `socket` such as, in this case, a _pool_ of `socket`/`connection` is possible - but it's not well documented and it's not the part of `conduit` to propose such implementation. The initial issue should be fixed in a different path where the user should implement its own _protocol_, register it into `conduit` and re-extract/de-abstract it as well in `callv`.

So it will not fix the initial issue (where `conduit` provides something simpler/easier to maintain) but it unlocks the ability to inject an another (new) kind of connection (with another impl. of `connect`/`recv`/`send`/`close`) into `cohttp` independently than `conduit` and `cohttp`.

So for me, we should keep #704 and may be rethink it in the scope of `conduit3.0.0`.

> do think that with you the new conduit and this revamp of cohttp we can also address the issue pointed out here: #696?

It seems that this specific issue is out of the scope of the purpose of `cohttp + conduit.3.0.0` and the issue is more about the usage of `Lwt_io.channel` (already provided by `conduit`).

mseri:
Thank you very much for the fast response! Let's keep #704 open and see if we can rethink it then.

I agree that having a default DNS resolver would be incredibly useful for http(s) connections. I think we should add a paragraph in the documentation referring to the appropriate conduit section for people that want to implement their own custom resolver.

examples/async/hello_world.ml
dinosaure: We can imagine to replace my update by `let _, server = ... in server >>= fun _ -> Deferred.never ()`. `Cohttp_async.Server.create` does not returns an `Async` value but a pair of an `Async` condition and an `Async` value. The user must destruct it and use `server` to _launch_ the server (however, in reality, the server is already launched). `_never` is the condition to stop the server and in this special case, we don't want to stop the server - so I added `_` in front of `never`.

mseri: Cool, thanks!

cohttp-lwt-unix/src/server.mli
```diff
@@ -34,5 +34,6 @@ val create :
   ?backlog:int ->
   ?stop:unit Lwt.t ->
   ?on_exn:(exn -> unit) ->
-  ?ctx:Net.ctx ->
-  ?mode:Conduit_lwt_unix.server -> t -> unit Lwt.t
+  'cfg ->
+  (_, 'flow) Conduit_lwt.protocol ->
+  ('cfg, 't, 'flow) Conduit_lwt.Service.service -> t -> unit Lwt.t
```
CraigFe, line 39: Is it possible with this signature to block on the server being ready to receive messages? (I was expecting to see a `unit Lwt.t Lwt.t` or similar somewhere.)

dinosaure: I can provide something such as `unit -> unit Lwt_condition.t * unit Lwt.t` if you want to be able to delayed the initialization of the server but I would like to know the opinion of others peoples. Note that such request will fallback to an update of `ocaml-conduit`.

samoht: I think it's a good idea to do it ; it's a bit painful right-now to write reliable tests which are using cohttp because of this issue (e.g. we need to add an extra retry-logic to make sure the server is ready before starting the tests, see for instance https://github.com/mirage/irmin/blob/master/test/irmin-http/test_http.ml#L61-L82)

mseri: I agree, this sounds like a useful thing to have

dinosaure:
So I finish to implement the request of @mseri about resolvers. I notice that the current implementation of `cohttp-async`,  even if it use `conduit-async` does not rely on `Conduit_async.resolve` and it uses directly `Conduit_async.connect` with resolved values. In other words, `cohttp-async` does by itself the introspection of the given `Uri.t` and _create_ the value needed to start a connection by itself (without the mechanism of `conduit`).

The current behavior (which can be replaced by the user) is to be able to resolve an HTTP connection and an HTTPS connection with `SSLv23` (see `Ssl.protocol`). A comment was added into the code to avoid any surprise on this part.

The request from @CraigFe was added too and fits well then into executables.

Let me clean the PR and it's ready to merge.

dinosaure:
READY TO MERGE 🎉 🌮 !

CraigFe:
The CI is broken because Dune 2.0+ enables formatting of OCaml files by default.

You should add something like:

```
(formatting (enabled_for dune))
```

to disable formatting on this repository. Adding an `.ocamlformat` file with `disable = true` would also work.

mseri:
@hcarty wrote me the following:

> It looks like `cohttp-lwt-unix` now requires `lwt_ssl` which, in turn, requires one of the openssl-like libraries.  Do you know if that's an intentional and permanent change?  Will `ocaml-tls` be supported outside of mirage builds?

dinosaure:
> @hcarty wrote me the following:
> 
> > It looks like `cohttp-lwt-unix` now requires `lwt_ssl` which, in turn, requires one of the openssl-like libraries.  Do you know if that's an intentional and permanent change?  Will `ocaml-tls` be supported outside of mirage builds?

About details, `cohttp-lwt-unix` requires `conduit-lwt-ssl` for:
1) provided binaries such as `cohttp_proxy_lwt` and `cohttp_server_lwt` which are not the part of the library
2) be able to produce a TCP/IP + SSL connection to handle, by default, HTTPS connection (see `net.ml`) - which is a part of the library

The second point came since f9b2ff8 to:
> having a default DNS resolver would be incredibly useful for http(__s__) connections

At this stage, the user is able to use `conduit-lwt-tls` (which uses `ocaml-tls`) and extend `resolvers` such as:
```ocaml
let tls_config = Tls.Config.client ...

let resolvers =
  Conduit_lwt.add Conduit_lwt_tls.TCP.protocol
    (Conduit_lwt_tls.resolve ~port:443 ~config:tls_config) Cohttp_lwt_unix.Net.empty
```

To start an HTTPS connection with `ocaml-tls` - note that you can add a `priority` argument to enforce `conduit` to try `ocaml-tls` before `ssl`. The user can use its `resolvers` to let `conduit` to dispatch to the right implementation:

```ocaml
Cohttp_lwt_unix.Client.get ~resolvers (Uri.of_string "https://google.com")
```

### About the old version of `conduit`

The old version of `conduit` choose the TLS implementation according what it is available into the _build_ scope (OPAM or `dune`) at the compile time. So if you have `lwt_ssl`, `conduit` will be compile with this implementation of TLS. Otherwise, it will try to choose `ocaml-tls`. For my perspective, this _feature_ is wrong and leads several errors and issues (see [conduit-3.0.0](https://github.com/mirage/ocaml-conduit/pull/311#issuecomment-629342906) for more details).

So about the new version of `conduit`, we explicit the choice by two packages:
- `conduit-{lwt,async}-ssl` and use OpenSSL
- `conduit-{lwt,async}-tls` and use `ocaml-tls`

The user is able to choose one of them or both, the choice should be made by an extension of the `resolvers` which these protocols. So `ocaml-tls` can be used outside a MirageOS project.

### About f9b2ff8

The first task was to handle HTTP and HTTPS connection for `cohttp-lwt-unix` (this approach does not concern `cohttp-async` which implements [its own dispatch](https://github.com/mirage/ocaml-cohttp/blob/63e6026178b8a8750837f343d386ad5445aa236d/cohttp-async/src/client.ml#L31)). To be able to resolve an HTTPS connection, you must need a TLS implementation. My arbitrary choice is to use `lwt_ssl` to be able to handle HTTPS connection without any required update from the user.

Of course, we can discuss about this choice - but it will lead the default behavior of `cohttp-lwt-unix` when you want an HTTPS connection. We have several possibilities:
1) handle only `*:80` and let the user to choose which implementation he wants
2) choose `conduit-lwt-tls`/`ocaml-tls` as the default way to initiate an HTTPS connection
   It implies a dependence to `ocaml-tls` (and remove `lwt_ssl`)
3) use `conduit-lwt-tls` and `conduit-lwt-ssl` (so OpenSSL and `ocaml-tls`) with a certain order of priorities
   It implies a dependence to `ocaml-tls` and `lwt_ssl`
4) stay with this behavior (default implementation is `lwt_ssl`)

I don't have strong opinion with all of these choices, so it's hope to you :+1:.

hcarty (commented):
So far just a question about `cohttp-lwt-unix` and `ocaml-tls`.  I haven't tested this yet but hope to soon!  Thanks for all your conduit and cohttp 3.x work @dinosaure!

hcarty:
Oops, I submitted my comment before seeing that @mseri had already raised the question!  Apologies for the noise.

dinosaure:
If anyone wants to test `cohttp`, I just made an `opam-repository` with an update of all the stack here: https://github.com/dinosaure/conduit-dev

mseri:
Thanks a lot for all this work @dinosaure

mseri:
Sorry @dinosaure I merged a PR that created some small merge conflicts. I will not be able to have a look before a couple of days, if they are not fixed by then I can take care of it

CraigFe:
I've pushed a fix; just two new stanzas that needed to be converted to Dune 2.0. Hope you don't mind @dinosaure :-)

mseri:
Thanks @CraigFe!

> Of course, we can discuss about this choice - but it will lead the default behavior of `cohttp-lwt-unix` when you want an HTTPS connection. We have several possibilities:
>     1. handle only `*:80` and let the user to choose which implementation he wants
>     2. choose `conduit-lwt-tls`/`ocaml-tls` as the default way to initiate an HTTPS connection
>        It implies a dependence to `ocaml-tls` (and remove `lwt_ssl`)
>     3. use `conduit-lwt-tls` and `conduit-lwt-ssl` (so OpenSSL and `ocaml-tls`) with a certain order of priorities
>        It implies a dependence to `ocaml-tls` and `lwt_ssl`
>     4. stay with this behavior (default implementation is `lwt_ssl`)

I don't have a strong preference. I think 4 is better since it requires way fewer dependencies than using `ocaml-tls` (which on the other hand could already transitively be there).

I would not be against 1 in principle, but history tells us that then we would be flooded by requests on how to use tls. In that case, would it be possible, and **would it make any sense**, to mock what conduit is doing and introduce `cohttp-lwt-unix-ssl`, which is practically `cohttp-lwt-unix` + the https resolver, and then go with 1 for `cohttp-lwt-unix`? Then in the README and in the documentation we could show how to extend it with `tls` for users that prefer it, or we could add `cohttp-lwt-unix-tls`. The world is moving towards https first, but I worry that it could be an issue in cases in which you cannot install/update openssl. In that case a hard dependency on ssl would prevent you to use `cohttp-lwt-unix`.

@hcarty what do you think?

hcarty:
@mseri I like the idea of moving to `cohttp-lwt-unix-ssl` and `cohttp-lwt-unix-tls` for exactly the reasons you specified.  It's really easy, even when you know what you're doing!, to accidentally end up with a build of cohttp 2.x that doesn't know how to speak https.  This is particularly problematic for the client interface since the first thing a lot of users will end up talking to is likely listening for https requests.

Could we use the name `cohttp-lwt-unix-notls` or `-nossl` for the library that doesn't support https?  Naming a something according to a missing feature is not a good idea in general but given the prevalence of https the name seems worthwhile here.

`cohttp-lwt-unix` in 3.x could be an alias for `cohttp-lwt-unix-ssl` to limit confusion when users move from 2.x to 3.x.

@dinosaure Thank you for all your time and effort on this adaptation!

dinosaure:
If you agree with that @mseri, I can do the job on my side :+1:.

mseri:
Yes, I agree with @hcarty suggestion. Namingwise, I don't have a preference. Feel free to pick a side :)

dinosaure:
> Yes, I agree with @hcarty suggestion. Namingwise, I don't have a preference. Feel free to pick a side :)

I took `cohttp-lwt-unix-nossl` as a new OPAM package which it seems enough explicit. However, on the OCaml side, I kept `Cohttp_lwt_unix` - this diphasing can disturb a bit the end-user where, by an installation of `cohttp-lwt-unix-nossl`, the module `Cohttp_lwt_unix` will be available (instead of `Cohttp_lwt_unix_nossl`). This is my opinion where an append of `_nossl` for a module used by `cohttp-lwt-unix-{tls,ssl}` is a bit annoying.

Again, I don't have strong opinion about renaming `Cohttp_lwt_unix` to `Cohttp_lwt_unix_nossl` if you are afraid by a cascade of issues about that.

Then, as @hcarty asked, `cohttp-lwt-unix-{tls,ssl}` are available where it's a just an other specialization of the module `Net` with a default context. Note that `ocaml-tls` does not come with root anchors - which is an other problematic - and retook the same behavior than the old version of `conduit`, a `Tls.Config.client` which accepts any certificates. It's certainly an issue, however, it does not concern `conduit` or `ocaml-tls` (but the glue between them) - [mirage/ca-certs](https://github.com/mirage/ca-certs) was done in this way (but it was not released).

So, let me know about `Cohttp_lwt_unix` or `Cohttp_lwt_unix_nossl`.

PS: tests are a bit flacky but it seems an error which come from the orchestration of test executables by `dune`. If I have time, I will figure out an other solution to fix it.

cohttp-lwt/src/client.mli
```diff
@@ -4,4 +4,4 @@
     module.  The resulting module satisfies the {! Client } module type. *)
 
 module Make (IO:S.IO) (Net:S.Net with module IO = IO) : S.Client
-  with type ctx = Net.ctx
+  with type resolvers = Net.resolvers
```
avsm, line 7: This renaming causes interface churn -- could we just keep the `ctx` name in this release?  Most people will just pass it through unmodified, but the type renaming will break libraries that use the Cohttp functor.

cohttp-lwt/src/s.ml
```diff
@@ -21,9 +21,11 @@ end
     and close the resulting channels to clean up. *)
 module type Net = sig
   module IO : IO
-  type ctx [@@deriving sexp_of]
-  val default_ctx: ctx
-  val connect_uri : ctx:ctx -> Uri.t -> (IO.conn * IO.ic * IO.oc) Lwt.t
+
+  type resolvers
```
avsm, line 25: This does need to be sexp-serialisable as well.

cohttp-lwt-unix-tls/src/net.ml
```diff
@@ -0,0 +1,60 @@
...
+let empty =
+  Conduit_lwt.empty
+  |> Conduit_lwt.add Conduit_lwt.TCP.protocol
+    (Conduit_lwt.TCP.resolve ~port:80)
+  |> Conduit_lwt.add ~priority:10 Conduit_lwt_tls.TCP.protocol
+    (Conduit_lwt_tls.TCP.resolve ~port:443 ~config)
+
+let failwith fmt = Format.kasprintf (fun err -> Lwt.fail (Failure err)) fmt
+
+let connect_uri ?host:(default= "localhost") ~resolvers uri =
```
avsm, line 40: `?host` doesn't actually seem to be used here. What is it for?

avsm (changes requested):
With respect to the opam changes, let's go with `cohttp-lwt-unix.opam` for the TLS-enabled version. The default should be TLS on the modern Internet.  We can retain the `cohttp-lwt-unix-notls.opam` for the bare version, but I expect that not many users will need that.

It would be good to minimise some of the interface changes here, notably `ctx`->`resolve` as downstream libraries use the functor signature and would break unnecessarily.   Other than that, it looks good to me. I'm going to give it a bigger round of tests through this week but wanted to get you these review comments now; thanks for bearing with me on the long delay :-)

cohttp-lwt/src/client.mli
dinosaure: I'm ok with that :+1:.

cohttp-lwt/src/s.ml
dinosaure: It will be an opaque sexp-serialisable value where `resolvers` is a map of functions. It's fine for you?

cohttp-lwt-unix-tls/src/net.ml
dinosaure: It is used by `Uri.host_with_default ~default` where we want to extract the `domain-name` from the `Uri.t`. `Uri.t` is super-versatile in this case and the cast can lead to an error when you give something like: `https://0.0.0.0/` - throught, I started to discuss about this specific point [here](https://github.com/mirage/ocaml-conduit/pull/311#issuecomment-694970864).

dinosaure:
> With respect to the opam changes, let's go with cohttp-lwt-unix.opam for the TLS-enabled version. The default should be TLS on the modern Internet. We can retain the cohttp-lwt-unix-notls.opam for the bare version, but I expect that not many users will need that.

I will revert so my last commit to provide `cohttp-lwt-unix` with TLS.

mseri:
This looks good for what concerns me. Once conduit 3.0.0 has been released, if you can clean it up and make the tests green, and if @avsm agrees, I think we could merge and release.

Before releasing I will double-check the README and the examples there, to update the stale content

cohttp-mirage/src/client.ml
```diff
@@ -19,29 +19,28 @@
...
 
-  let connect_uri ~ctx uri =
-    Resolver_lwt.resolve_uri ~uri ctx.resolver >>= fun endp ->
-    Conduit_mirage.client endp >>= fun client ->
-    Conduit_mirage.connect ctx.conduit client >>= fun flow ->
-    let ch = Channel.create flow in
-    Lwt.return (flow, ch, ch)
+  let connect_uri ?host:(default= "localhost") ~ctx uri =
+    let domain_name = Domain_name.(host_exn (of_string_exn (Uri.host_with_default ~default uri))) in
+    Conduit_mirage.resolve ctx domain_name >>= function
```
hannesm, line 38: here we loose information which makes the resolve process tedious (undoable):
- consider the uri being http://localhost, all is fine and good
- consider the uri being http://localhost:8080, now the port information is lost (not transmitted to Conduit_mirage.resolve, which constructs an endpoint) -- this can be fixed in here `match Uri.port uri with None -> endp | Some port -> {endp with port}`
- consider the uri being https://localhost:1234, now Conduit again sees only "localhost", but neither "https" nor the port "1234" - the latter can be fixed as above, but it is not clear to me how conduit should be able to derive from "localhost" whether it should setup a tcp or tls endpoint.

dinosaure: After a discussion, it seems that this _error_ is mostly about _design_. With a big retrospective of what `conduit.2.0.0` and what `conduit.3.0.0` is, it seems that we miss something between an `Uri.t` and a `[ host ] Domain_name.t` (or any less/more complex concrete type required by `conduit`).

In one side, `ocaml-git` decides to handle by itself with a `Smart_git.endpoint` which connection we should start depending on what the user have (what is accessible) and the Git endpoint (or GRI). This is on the responsibility of the user to generate a `Conduit.resolvers` regarding to the `Smart_git.endpoint` (the user is able to introspect it) __and__ in which protocols it can communicate.

From my point of view, for `cohttp`, it should be the same. An `Uri.t` can point to anything and it does not concern _only_ HTTP(s). The user should, as `ocaml-git`, able to make a proper `resolvers`/`ctx` depending on his context and the given `Uri.t` but it should not be on the responsibility of `cohttp` to care about that.

In other words, the main difference between `ocaml-git` and `cohttp` is:
1) `ocaml-git` does not define a default `resolvers`
2) `cohttp` defines a default `ctx`/`resolvers` which can not take the opportunity to fully understand an `Uri.t`

The second point is really true for MirageOS. For example about the port information: try to pass an `Uri.t` with a port to `Conduit.resolve` assumes that __any__ protocols registered into `conduit` can do something with the `port` value (a systemic information) but in reality, it's not true. Of course, in most of the cases, we want to use a TCP/IP protocol (plus TLS) which requires the `port` - but what about UNIX domain, VCHAN or any protocol available on a device? This information will be useless. That mostly shows to us that the interpretation of an complex _endpoint_ depends on what is available about protocols.

An other example is about scheme. First, [the scheme __is not__ the protocol](https://tools.ietf.org/html/rfc7595#section-3.8). And from my point-of-view, it appears that we initiated a non-legacy (regarding to IANA) arbitrary dispatch over scheme with `conduit.2.0.0` (where we decided into the impl. of `conduit` that some schemes will produce some "protocols").

The point is, the choice of `[ `host ] Domain_name.t`/`Conduit.Endpoint.t` is to be, certainly, less expressive about required information to start a `Conduit` connection, but it sanitizes on the other hand the dispatch of protocols over less information - by this way, it gives a better opportunity to _infer_ which protocol `conduit` chosen.

More globally and independently of `conduit.2.0.0`/`conduit.3.0.0`, we miss a middle-step/layer between `cohttp`, `uri` and `conduit`. On `http/af` (see `paf`), the choice was made to let (as `ocaml-git`) the user to define the dispatch/the `resolvers`.  Of course, it will be good to have `Uri.t -> Conduit.flow` which can handle every information available into the given `Uri.t` - but we should assume that this step should be a bit far from the dispatch of _protocols_ (and this is the case from the UNIX babylon tower).

dinosaure:
Flaky tests was fixed, last update about the API of `conduit`. So I think we are ready for a merge. I will be available next week for any minor updates - but it seems that it will be about documentation/lint/etc. I hope that next week we will be able to cut a release.

cohttp-mirage/src/client.ml
hannesm: thanks for your detailed reply. To keep (most of) the cohttp client API, and preserve the semantics in the common case, I'd propose the following:
- Make `Net.connect_uri` `ctx:ctx` argument optional (used by cohttp-lwt-unix and cohttp-mirage to pass in the `default_ctx` / `empty`)
- In `connect_uri`, do the above mentioned dispatch on Uri (i.e. schema = https -> use tls resolver, schema = http -> use tcp resolver) and port (no `:abcd` specified, use default port, otherwise use port `abcd`).

The selection of the resolver in `connect_uri` shoud only be done if no custom resolver is provided (here, unfortunately the lwt-unix and mirage implementations differ -- the lwt-unix has a default resolver with tcp and tls, the mirage implementation uses an empty resolver) --> I'd suggest to unify the default ctx and have some defaults for mirage as well :)

For users that manually pass a custom resolver through, their code will need to change since resolvers have changed as well. For users (of cohttp-lwt-unix) who do not pass a custom ctx through, with the above suggested change, the behaviour will be the same, test cases:
http://example.com -> use tcp resolver and port 80
https://example.com -> use tls resolver and port 443
http://example.com:1234 -> use tcp resolver and port 1234
https://example.com:1234 -> use tls resolver and port 1234

dinosaure:
So, I integrated last few fixes from @hannesm about what he expects (and delete some diffs about the API which is better). The `cohttp-mirage` should be updated too but I suggest to post-pone that to another PR. Several issues still exist but it's mostly about the new use of `conduit` into `cohttp` (comparing to `ocaml-git`). I can reclean the PR (@mseri?) but it seems that it should fine.

Documentation and `README.md` was not done. About that, I prefer to post-pone these updates to another PR. I still believe that a release can be done at the end of this week if nobody argue about some details. Of course, anybody can share his opinion about all of this PR :+1:.

mseri:
If @avsm and @hannesm are also happy, I am fine with your suggestion. I think we need to update documentation and READMe before a release though (it can be done as a separate pr fwiw). I am happy to help with it but I don't know if I can do it bedoel the end of the week

mseri:
I will wait until the late afternoon to see if there are more comments (or requests to wait before merge). If not, I will merge

hannesm:
I tried (using cohttp-lwt-unix/bin/cohttp_lwt_curl.exe), and figured the following:
- `cohttp_curl_lwt.exe http://retreat.mirage.io` leads to an expected positive result
- `cohttp_curl_lwt.exe https://retreat.mirage.io` leads to some unexpected positive result - the host only serves plain http:// on port 80
- `cohttp_curl_lwt.exe https://github.com` leads to an unexpected error: `cohttp-curl: internal error, uncaught exception: (Failure "Client connection was closed")`
- `cohttp_curl_lwt.exe http://github.com` leads to the same unexpected error as above (and connects to https / port 443)
- `cohttp_curl_lwt.exe http://github.com:80` leads to the expected positive result "301 Moved permanently"

I have not checked with cohttp on the latest release, but I think there are some unexpected behaviour changes -- as outlines in earlier comments -- if someone uses http://yyy, I'd expect a plain text http connection on port 80, and the behaviour of http://yyy and http://yyy:80 should be identical.

A separate issue is that it looks like client https connections are not working at all - always receiving Eof -- any hints where to look?

mseri:
Indeed, I was able to replicate the issue with the latest version of the code also on different projects. Well spotted!
I think the curl example used to work in a previous version of the PR, but I am no longer sure I did try https calls with it!

mseri:
Now that I think about it, I tested `cohttp-server-lwt` but not with `--tls`. @dinosaure have you tried that?

dinosaure:
I'm currently on @hannesm's issues, hard to find at which layers we have an error.

cohttp-lwt-unix/src/net.ml
```diff
@@ -21,26 +21,58 @@ open Lwt.Infix
...
+   *
+   * We can see an another use of [Conduit] with [Cohttp_async] which does not
+   * rely on the given [ctx] but it uses [Conduit_async.connect] according to
+   * the introspection of the given [uri]. We have 3 incompatible choices here:
+   * - expect an user's [ctx] with a [Conduit.Endpoint.t]
+   * - make a [ctx] from the given [uri]
+   * - *)
+  let ctx = match Uri.scheme uri, Uri.port uri with
+    | Some "https", Some port ->
+      Conduit_lwt.add ~priority:0 Conduit_lwt_tls.TCP.protocol (Conduit_lwt_tls.TCP.resolve ~port ~config:tls_config) ctx
```
mseri, line 66: What if `ctx = init ~tls_config:some_other_tls_config ()`? Why here we resolve always using the default config? Shouldn't we pass the same one as the context instead?

dinosaure:
So we have 2 kinds of errors:
1) According to https://github.com/mirage/ocaml-cohttp/pull/692#discussion_r498884883, the `default_ctx` used then to start a connection __must__ be empty. If we consider a default behavior from a given `Uri.t`, we must construct a `ctx`/`Conduit.resolvers` from this `Uri.t` - the user still will be able to _inject_ its own way (with priority) to initiate a connection (like example with Docker in the `README.md`)
2) `conduit-tls` + `Lwt_io` (used by `cohttp`)  is broken and it's hard to really see a good way to fix that - I'm currently on it

cohttp-lwt-unix/src/net.ml
dinosaure: The best for us (and to have an understandable behavior of `conduit`) is to construct a _special_ `ctx`/`Conduit.resolvers` from the given `Uri.t` - like, an introspection of the scheme, the port, etc. to produce the right way to initiate a connection. The `ctx` still will be on the control of the user - and can prioritize an other initialization (depending on his context) - but we should not try to use what `init ()` gives to us or `default_ctx` - I mean, such `ctx` does not correspond in anyway to the given `Uri.t` and it's why the current status of `cohttp` has some strange behaviors (which don't reflect the given `Uri.t`)

dinosaure:
With https://github.com/mirage/ocaml-conduit/pull/328 and the last patch, the behavior explained by @hannesm works perfectly:
- `cohttp_curl_lwt.exe http://retreat.mirage.io` works (simple HTTP request)
- `cohttp_curl_lwt.exe http://retreat.mirage.io` **does not** work (remove of `Cohttp_lwt_unix.Net.init ()`)
- `cohttp_curl_lwt.exe https://github.com` works (mutex on `Conduit_lwt.io_of_flow`
- `cohttp_curl_lwt.exe http://github.com` works (fix first-recv behavior of `Lwt_io` on `Conduit_lwt.TCP`)
- `cohttp_curl_lwt.exe https://google.com` works (fix first-recv behavior of `Lwt_io` on `conduit-tls`)
- `cohttp_curl_lwt.exe https://www.google.com` works (simple HTTPS request)

I will try to see if we have any issues about `cohttp-server-lwt` @mseri.

mseri:
Awesome, thanks!

mseri:
Do we still plan to re-add a `cohttp-lwt-unix-notls` package?

dinosaure:
> Do we still plan to re-add a cohttp-lwt-unix-notls package?

So, according to the last commit, the change will be on `Net.connect_uri` - and raise `invalid_arg` if the user give an https URL - so, I can do that now, yes.

hannesm:
I added trust anchor verification using the ca-certs package in https://github.com/dinosaure/ocaml-cohttp/pull/2

dinosaure:
https://github.com/mirage/ocaml-conduit/pull/328 was updated properly to fix the server-side. The main problem is the use of `Lwt_io` which wants to start to read in any context - even if the underlying socket is not `Lwt_unix.readable`. So I did some tweak on `conduit-tls` and `conduit-lwt` (`TCP` modules) to properly start to read if we are the server and properly start to send if we are the client. The TLS layer gives the opportunity to re-schedule the `read`/`write` to the `io_of_flow` function used by `cohttp-lwt` then.

- Behaviors described by @hannesm work (as I wrote before).
- A TLS server or a simple HTTP server work too from my tests (on my computer and on my server).

However, I would like a double check when the problem is about scheduling between the `read` and the `write` used by `Lwt_io`. Then, this issue highlights that we should not use `Lwt_io` finally - I mean, if we want to properly handle an HTTP/1.1, we should have a full control of when we want to read and when we want to write.

mseri:
The extra test that I am making is the following:
```
git clone https://github.com/mseri/doi2bib/
cd doi2bib
opam install . --deps-only
dune runtest
```

Most test cases (but not all!) are now failing:
```
$ dune exec bin/doi2bib.exe -- 1902.00436        
Fatal error: exception (Failure "Client connection was closed")
```

The only call that is not failing is
```
$ dune exec bin/doi2bib.exe -- arXiv:math/0606217
@misc{Marklof2006Distribution,
      title={Distribution modulo one and Ratner's theorem}, 
      author={Jens Marklof},
      year={2006},
      eprint={math/0606217},
      archivePrefix={arXiv},
      primaryClass={math.NT}
}
```
which is the only one not calling the DOI registry with a custom header.

dinosaure:
@mseri according to mirage/ocaml-conduit#329 and the last patch, your tool works perfectly (with `ocaml-tls`). However, it seems more trickier about OpenSSL support currently, I think we should provide `cohttp-lwt-unix-ssl` and `cohttp-lwt-unix-notls` (as you suggested) to fix a global and systemic error: the use of `Lwt_io`.

The error did not appear before because we were able to make `Lwt_io` values from an `Lwt_ssl.socket` and mostly because the old version of `conduit` was not really abstracted (for this point, due to the choice at the compile time of which implementation of TLS we wanted, the old version of `conduit` was able to directly provide `Lwt_io` values from the well-known `Lwt_ssl.socket` - however, with the new version of `conduit` which is fully-abstract, we must rewrap `Lwt_io` with `Conduit.send` and `Conduit.recv`).

About the idea to make `cohttp-lwt-unix-ssl`, it's just to have the ability (like before) to get `Lwt_io` values in an other way than `Conduit_lwt.io_of_flow` (which does not work in anyway for a `Conduit_lwt_ssl.Protocol.flow`) - and, if we have a `Conduit_lwt_ssl` flow, we are able to _destruct_ it and make `Lwt_io` values from `Lwt_ssl.{out,in}_channel_of_descr`.

About the error with `ocaml-tls`, it's mostly due to that doi.org requires at the handshake the peer name (see 7d56281). Otherwise, it fails.

With these patches and the idea, I can use `doi2bib` with `ocaml-tls` and OpenSSL as before :tada:. I will make these packages tomorrow.

mseri:
Awesome. Thank you very much for looking into it so fast.
Indeed, now it works perfectly

I like the plan! Look forward for the final changes so that we can finally merge this

mseri:
@dinosaure if you remove the pin-depends, this is ready for merge

mseri:
Thanks. I will wait for the CI before merging, just to check that all is good with the new opam-repository version

mseri:
Before making a new release there are two small breaking changes from previous PRs that I want to bring back

mseri:
Thanks again for such big effort

PR #694, opened 2020-05-27
Correction of [Header.replace] function and better encoding management in [Request.make] and [Response.make] functions.

lyrm:
1.  `Header` module: 
The `replace` function was not working properly (it should do nothing if the key is not already in the headers). I have also corrected some minor typos in documentation.

2. `Response.make` function:
This function create a `Response.t` record from the inputs. In particular, it fills its `encoding` field with the optional argument `~encoding`. However it does not take into account the headers and that can cause a bug when these values are different. Read the detailed example below for more explanation. In brief, you can have a chunked body that is read as not chunked by the client. 

3. `Request.make` function:
As requested by this [RFC](https://tools.ietf.org/html/rfc7230#section-3.3.2), the `guess_encoding` function should first check for the presence of the header *transfer-encoding* before checking for *content-length* or *content-range* headers to determine the body encoding.

**Example**: 
Let's take a response defined by: 
```ocaml 
let resp = Response.make ~headers:(Header.of_list ["content-length", "100L"]) ()
````
The value of the field `resp.encoding` is therefore `Chunked` (default value).  When the response is written (see `Cohttp-lwt.Server.handle_client` which calls `Cohttp.Response.write`): 
-  as the headers contain a `content-length` value, they are written unchanged by `Response.write_header` whatever the value of `response.encoding`.  
-  the value of the field `encoding` is used to determine how to write the response (see `Response.make_body_writer`) without considering the headers. 

As a result, the body is written as a chunked body whereas the client will read it as not chunked.

### DIFF
--- a/cohttp/src/request.ml
+++ b/cohttp/src/request.ml
@@ -27,9 +27,9 @@ type t = {
 let fixed_zero = Transfer.Fixed Int64.zero
 
 let guess_encoding ?(encoding=fixed_zero) headers =
-  match Header.get_content_range headers with
-  | Some clen -> Transfer.Fixed clen
-  | None -> encoding
+  match Header.get_transfer_encoding headers with
+  | Transfer.(Chunked | Fixed _) as enc -> enc
+  | Unknown -> encoding
 
 let make ?(meth=`GET) ?(version=`HTTP_1_1) ?encoding ?headers uri =
   let headers =
--- a/cohttp/src/response.ml
+++ b/cohttp/src/response.ml
@@ -24,8 +24,12 @@ type t = {
   flush: bool;
 } [@@deriving fields, sexp]
 
-let make ?(version=`HTTP_1_1) ?(status=`OK) ?(flush=false) ?(encoding=Transfer.Chunked) ?headers () =
-  let headers = match headers with None -> Header.init () |Some h -> h in
+let make ?(version=`HTTP_1_1) ?(status=`OK) ?(flush=false) ?(encoding=Transfer.Fixed Int64.zero) ?(headers=Header.init ()) () =
+  let encoding =
+    match Header.get_transfer_encoding headers with
+    | Transfer.(Chunked | Fixed _)  as enc -> enc
+    | Unknown -> encoding
+  in
   { encoding; headers; version; flush; status }
 
 let pp_hum ppf r =
--- a/cohttp/src/response.ml
+++ b/cohttp/src/response.ml
@@ -24,7 +24,7 @@ type t = {
   flush: bool;
 } [@@deriving fields, sexp]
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: cohttp/src/response.ml [hunk 1]
```diff
@@ -24,8 +24,12 @@ type t = {
   flush: bool;
 } [@@deriving fields, sexp]
 
-let make ?(version=`HTTP_1_1) ?(status=`OK) ?(flush=false) ?(encoding=Transfer.Chunked) ?headers () =
-  let headers = match headers with None -> Header.init () |Some h -> h in
+let make ?(version=`HTTP_1_1) ?(status=`OK) ?(flush=false) ?(encoding=Transfer.Chunked) ?(headers=Header.init ()) () =
+  let encoding =
+    match Header.get_transfer_encoding headers with
+    | Transfer.(Chunked | Fixed _)  as enc -> enc
+    | Unknown -> encoding
```
avsm, line 31: Thanks, this looks like a strict improvement to me! Would it be possible to make it clear in the ocamldoc header for Response that the `headers` field has priority over the encoding field?  This would also be good to note in the CHANGES entry for this so that developers are aware of the change in case they happened to depend on the old behaviour.

The new one is obviously better though.

cohttp/src/request.ml
```diff
@@ -27,9 +27,9 @@ type t = {
 let fixed_zero = Transfer.Fixed Int64.zero
 
 let guess_encoding ?(encoding=fixed_zero) headers =
-  match Header.get_content_range headers with
-  | Some clen -> Transfer.Fixed clen
-  | None -> encoding
+  match Header.get_transfer_encoding headers with
+  | Transfer.(Chunked | Fixed _) as enc -> enc
+  | Unknown -> encoding
```
avsm, line 32: This looks good. Would it be possible to get a test case added to exercise this part of the RFC? I.e. one with the transfer encoding headers, and another with content range, and another with all three.  There are already some similar tests in the tests/ directory.

avsm:
Many thanks for the well-motivated PR and links to the relevant RFCs. Just left a few minor comments -- mainly, test additions to cover the changes would be appreciated. Otherwise good to merge from me.

cohttp/src/response.ml
lyrm: Hi ! Thanks for the review. Just a quick question before pushing: where should I write in the CHANGES entry ?

lyrm:
I have just added all the requested changes.

mseri:
Looks good, thanks!
@avam the ocaml-ci failure seems to be an excessively aggressive log analysis (and partial, it seems to miss the second unix error). I am asking because I don’t know how that works

mseri:
I am going to merge this, thanks!

Issue #690, opened 2020-02-23
callv: graceful handling of remote close


madroach:
Hi,

now my problem is about a connection shutdown initiated by the remote server during a pipelined connection (for example because it imposes a limit on requests per connection). This shutdown is only noticed after the next request is already sent to the wire and ``read_response`` tries to read from the in channel. It will then fail with ``Failure "Client connection was closed"``. This is especially annoying for non-idempotent requests, because the user won't know whether the request has been processed by the server.

It would be better to detect the shutdown as soon as the tcp FIN packet is received and then either attempt a reconnect or fail when the next request is queued.

rgrinberg:
@madroach is there a particular reason why you need pipelining? Pipelining is mostly a misfeature avoided by most http clients. I'd suggest you to do the same here.

madroach:
Hi, thanks for your reply.
I chose the wrong wording. The problem is not about pipelining, but only about connection keep-alive. During pipelining, which is currently not supported, it won't occur.

What is missing is a reader that can handle `EOF` while there is _no_ request on the wire.

I do already have a working pipelining implementation with connection cache sitting around since 2020, which solves this and supports pipelining, which in my mind has legitimate applications.
It just needs a bit polishing, then I'll submit a pull request.

madroach:
fixed by #853

PR #696, opened 2020-06-05
partly inline read_response, fix body stream leak

madroach:
I'm currently working on an improved pipelining implementation that will gracefully handle timeouts and simplify the interface. (See #690)
The changes in this pull request are just preparations.
But I believe they would benefit the existing code, too.

* to handle for example remote timeouts it is necessary to always listen on the incoming stream to read the EOF. But when starting to listen one may not yet know the request method of the next response. That's why I inlined the `Response.read` handling, which does not need knowledge of the request method.
* to me it seemed the finalising in the body handling is wrong. closefn is called twice ~~and there is a stream leak if the body is not fully consumed by the user~~.

### DIFF
--- a/cohttp-lwt/src/client.ml
+++ b/cohttp-lwt/src/client.ml
@@ -19,6 +19,8 @@ module Make
       let stream = Body.create_stream Response.read_body_chunk reader in
       let body = Body.of_stream stream in
       let closed = ref false in
+      (* Lwt.on_success registers a callback in the stream.
+       * The GC will still be able to collect stream. *)
       Lwt.on_success (Lwt_stream.closed stream)
         (fun () -> closed := true; closefn ());
       (* finalise could run in a thread different from the lwt main thread.
--- a/cohttp-lwt/src/client.ml
+++ b/cohttp-lwt/src/client.ml
@@ -64,12 +64,15 @@ module Make
     sent >>= fun () ->
     Response.read ic >>= function
     | `Invalid reason ->
+      closefn () ;
       Lwt.fail (Failure ("Failed to read response: " ^ reason))
     | `Eof ->
+      closefn () ;
       Lwt.fail (Failure "Server closed connection prematurely.")
     | `Ok res ->
       match meth with
       | `HEAD ->
+        closefn () ;
         Lwt.return (res, `Empty)
       | _ ->
         let body = read_body ~closefn ic res in
@@ -114,12 +117,15 @@ module Make
       Lwt_mutex.with_lock read_m begin fun () ->
         Response.read ic >>= function
         | `Invalid reason ->
+          closefn () ;
           Lwt.fail (Failure ("Failed to read response: " ^ reason))
         | `Eof ->
+          closefn () ;
           Lwt.fail (Failure "Server closed connection prematurely.")
         | `Ok res ->
           match meth with
           | `HEAD ->
+            closefn () ;
             Lwt.return (res, `Empty)
           | _ ->
             let body = read_body ~closefn ic res in
--- a/cohttp-lwt/src/client.ml
+++ b/cohttp-lwt/src/client.ml
@@ -12,28 +12,18 @@ module Make
 
   type ctx = Net.ctx
 
-  let read_response ~closefn ic _oc meth =
-    Response.read ic >>= begin function
-    | `Invalid reason ->
-      Lwt.fail (Failure ("Failed to read response: " ^ reason))
-    | `Eof -> Lwt.fail (Failure "Client connection was closed")
-    | `Ok res -> begin
-        let has_body = match meth with
-          | `HEAD -> `No
-          | _ -> Response.has_body res
-        in
-        match has_body with
-        | `Yes | `Unknown ->
-          let reader = Response.make_body_reader res ic in
-          let stream = Body.create_stream Response.read_body_chunk reader in
-          let closefn = closefn in
-          Lwt.on_success (Lwt_stream.closed stream) closefn;
-          let gcfn _st = closefn () in
-          Gc.finalise gcfn stream;
-          let body = Body.of_stream stream in
-          Lwt.return (res, body)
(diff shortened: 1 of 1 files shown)


cohttp-lwt/src/client.ml
```diff
@@ -13,33 +13,18 @@ module Make
...
-    t
+  let read_body ~closefn ic res =
+    match Response.has_body res with
+    | `Yes | `Unknown ->
+      let reader = Response.make_body_reader res ic in
+      let stream = Body.create_stream Response.read_body_chunk reader in
+      let body = Body.of_stream stream in
+      Lwt.on_success (Lwt_stream.closed stream) closefn;
+      Gc.finalise
+        (fun body -> Lwt.async (fun () -> Body.drain_body body))
```
madroach, line 24: The finaliser may run in any (system) thread. `Body.drain_body` calls into Lwt and therefore must be run in the Lwt main thread. How can this be accomplished? I suspect `Lwt_preemptive.run_in_main` is no option because not available in mirage !? Any suggestions how to deal with this?

mseri:
This needs to be reviewed also by somebody more acquainted with Mirage. @avsm do ou have time or suggestions?

cohttp-lwt/src/client.ml
mseri: To me it seems that 589464a is the most reasonable thing to do, given the constraints. But I'll wait for a more mirage-knowledgeable reviewer before taking premature decisions

mseri:
Hi @madroach can you rebase on the current master? I would like to get this into the next release. 

 I have asked for help to see if Lwt_preemptive.run_in_main is a potential problem for mirage and if there are other options to resolve this issue. Hopefully I will be able to tell you soon

mseri:
@madroach this is the discussion about `closefn` and why it should not be an issue if it is called twice: https://github.com/mirage/ocaml-cohttp/pull/446

cohttp-lwt/src/client.ml
mseri: I can confirm that `Lwt_preemptive.run_in_main` is only available in Unix and would be a problem for mirage

hannesm: Lwt_preemptive is not available in MirageOS, sorry about that. I'm not entirely sure what is the best to do here.

mseri: No problem, thanks for looking into it

cohttp-lwt/src/client.ml
```diff
@@ -13,33 +13,23 @@ module Make
...
-    Lwt.on_failure t (fun _exn -> closefn ());
-    t
+  let read_body ~closefn ic res =
+    match Response.has_body res with
+    | `Yes | `Unknown ->
+      let reader = Response.make_body_reader res ic in
+      let stream = Body.create_stream Response.read_body_chunk reader in
+      let body = Body.of_stream stream in
+      let closed = ref false in
+      Lwt.on_success (Lwt_stream.closed stream)
```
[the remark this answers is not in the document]
madroach: That's what I initially suspected because I thougt `Lwt.on_success` will hold a reference on `stream`. But this is not true. It is the other way round. `stream` holds a callback to the `Lwt.on_success` machinery. So the GC will be able to collect `stream`.

madroach:
> Hi @madroach can you rebase on the current master? I would like to get this into the next release.
> 
> I have asked for help to see ~if Lwt_preemptive.run_in_main is a potential problem for mirage and~ if there are other options to resolve this issue. Hopefully I will be able to tell you soon

So is `Lwt_preemptive.run_in_main` is no option ?

cohttp-lwt/src/client.ml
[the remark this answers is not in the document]
mseri: Oh, I see, thanks. This explains why while testing I was nit noticing any regression.
Can you add a comment about it on this line?

mseri:
> So is `Lwt_preemptive.run_in_main` is no option ?

That’s right, we cannot use it. You could override the function in `cohttp-lwt-unix` and use it there if you want. So that both the lwt module and lwt-unix module are correct, but in unix we use the better mechanism

mseri:
The test are failing with
```
OK
 test_sanity alias cohttp-lwt-unix/test/runtest (exit 2)
Fatal error: exception (Failure "Limit of file-descriptors reached")
```

I need to make a new, clean, environment locally and test but the master branch is building properly

cohttp-lwt/src/client.ml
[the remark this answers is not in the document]
mseri: It is no longer calling closefn in case of failure or cancellation now though, correct?

mseri:
Locally I am also getting
```
FF
==============================================================================
Failure: Cohttp_Lwt.Body:3:Stream with empty strings

not equal
------------------------------------------------------------------------------
==============================================================================
Failure: Cohttp_Lwt.Body:2:Stream with leading empty strings

is_empty should consume leading spaces
not equal
------------------------------------------------------------------------------
```
and stress-testing some clients I have around, I get also lots of instances of
```
Cohttp_lwt: body not consumed - leaking stream!
```
being printed on screen.

I think we need to investigate this more carefully before being able to merge the PR

dinosaure:
I added a patch to fix the test suite, due to the remove of `Lwt.on_failure`/`Lwt.on_cancel` and the special case with `HEAD`, we must take care by ourselves the close operation.

mseri:
So I wasn't completely off when I thought that `on_success` was not enough in https://github.com/mirage/ocaml-cohttp/pull/696/files#r510746387
Wouldn't it be better just to attach the `on_cancel` and `on_failure` callbacks instead?

dinosaure:
It's done, I re-use `Lwt.on_failure`/`Lwt.on_cancel` to be sure to keep the old behavior :+1:.

mseri:
Fantastic, thank you very much!

PR #698, opened 2020-06-25
Implement Unix domain socket support for Cohttp-async

Leonidas-from-XIV:
Conduit already supported it and Cohttp-async nearly did as well, defining a `httpunix` schema, but there were some snags in supporting it fully:

1. It tried to get the port of a Unix domain socket. They usually don't have ports. In theory creating such can be done with Uri, but it would be rather unexpected
2. When using Unix sockets on HTTP 1.1 it is required to send the `Host` header but [RFC2616, Section 14.26](https://tools.ietf.org/html/rfc2616#section-14.23) states that for Unix domain socket said header should be empty

The latter change was required to talk to the Docker daemon which otherwise returns HTTP 400, but now responds fine.

### DIFF
--- a/cohttp-async/src/client.ml
+++ b/cohttp-async/src/client.ml
@@ -15,7 +15,7 @@ end
 module Net = struct
 
   let lookup uri =
-    let host = Option.value (Uri.host uri) ~default:"localhost" in
+    let host = Uri.host_with_default ~default:"localhost" uri in
     match Uri_services.tcp_port_of_uri ~default:"http" uri with
     | None -> Deferred.Or_error.error_string
                 "Net.lookup: failed to get TCP port form Uri"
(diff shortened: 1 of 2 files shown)


mseri:
This was not a problem in cohttp-lwt-unix because there you can pass a custom resolver?
I wonder if we should fix that or if we should do here something akin to https://github.com/mirage/ocaml-cohttp/commit/a0c12408b4005da3fbb951662f2969be58519cdd

Leonidas-from-XIV:
Thanks for looking into this!

Personally I think the Lwt Docker example in the README is not a good user-facing API, because it does a lot of unintuitive things that no user would attempt to do on their own to connect to a Unix domain socket: create a resolver, add a magical hostname to it that points to a Unix domain socket, then create a `Uri.t` that is not pointing to the actual domain socket but rather pretend to speak HTTP to this magical host and remember to use the custom resolver.

I have not tried it, but I am not sure it would even work, since Cohttp-lwt doesn't know it is talking through an Unix domain socket, so it probably will add `Host: docker` as a header which is wrong and might actually be rejected by the Docker daemon.

The only advantage of the resolver approach I can see there is that it is probably pretty easy to use HTTPS over Unix domain sockets, but there are remain a number of questions like "how to verify the certificate" or "does anyone do HTTPS over Unix domain sockets".

I do see a point in having a similar approach both for Cohttp-lwt and Cohttp-async but I think it should be guided by ease of use/discoverability.

mseri:
This is more correct than before, so I am going to merge it. Can you add a line to the CHANGES file before I merge?

At a certain point I would like to make sure this is properly fixed also in the lwt code, thanks for this fix

PR #699, opened 2020-06-25
Adapt to async >= v0.14

copy:


mseri:
Thanks. I’ll merge this in few days, after releasing one last bugfix version with support to v0.13.

Issue #701, opened 2020-07-15
Headers: missing function to set a header, possibly overwriting an existing header


vouillon:
It would be convenient to have a function in module `Header` to add a header possibly replacing an existing header with the same key (similarly to `Hashtbl.replace`). The function `Header.replace` was doing just that, but its semantics was changed by #694

mseri:
There is `Header.add` to add a header. We could add `replace_if_exists` and `add_or_replace` functions which I am more in favour of and deprecate `replace` (maybe also revert to the old `replace` behaviour but update its docstring).

Since the conduit 3 release will be breaking anyways, we can fix the API if needed.

avsm:
I hadn't noticed that the `Headers.replace` changed in a minor release -- it was indeed a documentation error, but it would perhaps be better to keep its behaviour matching `Hashtbl.replace` (that is, add-or-update).

How about:

- `Header.replace` goes back to add-or-update with a corrected docstring
- Add `Header.update` which fails if the key doesn't exist.

We could cut this as a minor release (as it's restoring old functionality and adding a new function), and then merge conduit 3 for a 3.0 release subsequently to that.

mseri:
> We could cut this as a minor release (as it's restoring old functionality and adding a new function), and then merge conduit 3 for a 3.0 release subsequently to that.

Yes, this is what I was planning to do. But I wasn't sure if it is worth to do a renaming for 3.0.

> How about:
>
>-    Header.replace goes back to add-or-update with a corrected docstring
>-   Add Header.update which fails if the key doesn't exist.

If we add `update`, then there is no need to change the API in the 3.0 version, which is even better!
You think it should fail (say with `Not_found`) if the key is not there instead of doing nothing?

avsm:
I am a little reluctant to add more exception raising functions. But we could do `update` (return an option) and `update_exn` to raise a Not_found.

pm-mck:
Maybe a bit late in the game, but could `update` work similarly to [Map.S.update](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Map.S.html)?

For quick reference:
```
ocaml
val update : key -> ('a option -> 'a option) -> 'a t -> 'a t
```

[...]

PR #702, opened 2020-07-20
cohttp: revert change to Header.replace, add Header.update

mseri:
A change in #694 modified the semantics of Header.replace.
The semantics change is reverted, and a new Header.update function is introduced which changes a header value only when present.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 ## Unreleased
 
 - cohttp: a change in #694 modified the semantics of Header.replace.
-  The semantics change is reverted, and a new Header.update function
-  is introduced which changes a header value only when present. (@mseri)
+  The semantics change is reverted, and a new Header.update_existing function
+  is introduced which changes a header value only when present. (#702 @mseri)
 
 ## v2.5.3 (2020-06-27)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 ## Unreleased
 
 - cohttp: a change in #694 modified the semantics of Header.replace.
-  The semantics change is reverted, and a new Header.update_existing function
-  is introduced which changes a header value only when present. (#702 @mseri)
+  The semantics change is reverted, and a new Header.update function
+  is introduced, following the semantics of Map.update. (#702 @mseri)
 
 ## v2.5.3 (2020-06-27)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+- cohttp: a change in #694 modified the semantics of Header.replace.
+  The semantics change is reverted, and a new Header.update function
+  is introduced which changes a header value only when present. (@mseri)
+
 ## v2.5.3 (2020-06-27)
 
 - cohttp-async: adapt to async >= v0.14 (#699 @copy)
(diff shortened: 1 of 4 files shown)


dinosaure:
May be we should add some unit tests to avoid such breakage.

mseri:
Thanks @dinosaure. Before merging, I wanted to ask if it is better to have `update` do nothing when the key is absent (like here) or fail like suggested at the issue page: https://github.com/mirage/ocaml-cohttp/issues/701

mseri:
> May be we should add some unit tests to avoid such breakage.

I agree. Once we settle on a semantic for `update`, I'll add all the tests

dinosaure:
According #694:
> The replace function was not working properly (it should do nothing if the key is not already in the headers). I have also corrected some minor typos in documentation.

I think a case such as a use of `Header.replace` with a `fold` (eg. `List.fold_left`) should not fail and it should ensure that we inserted all required fields (even if they already exist at the beginning).

PS: more deeply, I think the use of a `StringMap` is not so good and a simple associative list should be better - but it's my opinion.

mseri:
I don't want to change the underlying representation at this stage, but I would not mind an issue to discuss it for cohttp 3.0.

Before discussing a new API or a new internal description, I would like to restore the semantics and add an `update` option. I think the suggestion by @anil on the issue page of returning a `'header option` instead of silently doing nothing is better than what I do in the PR. I would actually avoid producing the `_exn` version, interested users can easily add their exception raising on None.

mseri:
I have changed the function name to `update_existing`, added (I hope clear) descriptions and added unit tests for the expected behaviours. The CI failures seem due to the usual extra pedantic check of the logs if I am not mistaken.

dinosaure:
The error is about `async` and it is unrelated to `cohttp`:
```
#18 9.217 (cd _build/default/cohttp-async/test && ./test_async_integration.exe)
#18 9.217 Uncaught exception:
#18 9.217   
#18 9.217   (Sys_error "/etc/localtime: No such file or directory")
#18 9.217 
#18 9.217 Raised by primitive operation at file "stdlib.ml", line 398, characters 28-54
#18 9.217 Called from file "src/zone.ml", line 348, characters 21-49
#18 9.217 Called from file "src/timezone.ml", line 142, characters 8-75
#18 9.217 Called from file "camlinternalLazy.ml", line 31, characters 17-27
#18 9.217 Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
#18 9.217 Called from file "src/core_time_ns.ml", line 462, characters 38-56
#18 9.217 Called from file "src/raw_scheduler.ml", line 895, characters 41-69
#18 9.217 Called from file "cohttp-async/test/test_async_integration.ml", line 122, characters 21-49
```

So, CI should be happy :+1:.

mseri:
I modified `Header.update` to follow more closely the semantics of `Map.update` as suggested by @pm-mck. Instead of using the more general `f: string list option -> string list option`, I followed the `get` function and limited the type to `f: string option -> string option`, with some documented concatenation and splitting under the table.

If you feel that using a more general type, say `f: string list option -> string list option` or `f: string list -> string list` and reduce the magic in the function, I'm happy to modify it, otherwise I think this is ready to merge.

The version I wrote only accepting a `string list` was messier in my opinion (I was associating `[]` and `None` in that case), the one accepting `string list option` allows users to associate list of values to headers that should not accept them (note that you can still inject lists of values using `map`).

mseri:
@dinosaure can you have a look at the latest changes? If you are happy with it, I'd merge and draft a new minor release

mseri:
Thanks!

Issue #18, opened 2013-01-21
Is there a way to properly install cohttp with lwt support in opam?


rgrinberg:
Asking the question on SF: http://stackoverflow.com/questions/14410213/how-to-install-cohttp-with-lwt-support I get Thomas hinting that it must be done manually. 

Is it possible to make this process easier?

PR #706, opened 2020-08-17
Fix body not being uploaded with unchunked Async.Pipe.

mefyl:
`Body_raw.disable_chunked_encoding` would drain the pipe to get the content-length, but discard the read body, leaving it empty.

### DIFF
--- a/cohttp-async/src/client.ml
+++ b/cohttp-async/src/client.ml
@@ -128,20 +128,19 @@ let callv ?interrupt ?ssl_config uri reqs =
 let call ?interrupt ?ssl_config ?headers ?(chunked=false) ?(body=`Empty) meth uri =
   (* Create a request, then make the request. Figure out an appropriate
      transfer encoding *)
-  let req =
+  begin
     match chunked with
     | false ->
-      Body_raw.disable_chunked_encoding body >>| fun (_body, body_length) ->
-      Request.make_for_client ?headers ~chunked ~body_length meth uri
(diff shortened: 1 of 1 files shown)


mefyl:
The CI is red but I'm under the impression it's about the setups themselves ? For instance I'm seeing ```#18 19.28 Uncaught exception:
#18 19.28   
#18 19.28   (Sys_error "/etc/localtime: No such file or directory")```

cohttp-async/src/client.ml
```diff
@@ -128,20 +128,19 @@ let callv ?interrupt ?ssl_config uri reqs =
 let call ?interrupt ?ssl_config ?headers ?(chunked=false) ?(body=`Empty) meth uri =
   (* Create a request, then make the request. Figure out an appropriate
      transfer encoding *)
-  let req =
```
mseri, line 131: I think it is easier to read if you don't pipe such a large block. I'd suggest
```
let (req, body') = match chunked with
[...]
in
req >>= request ?interrupt ?ssl_config ~body:body' ~uri
```

mseri:
I agree that the CI failures seem all unrelated. If you make the little stylistic change above, I'll be happy to merge

cohttp-async/src/client.ml
mseri: I had another look at the code, similar blocks appear in other places. I am going to merge as is

PR #711, opened 2020-09-07
Update codes/*.json

### LINKED ISSUE #710: HTTP 308: Resume incomplete or Permanent redirect?
Hi,

Cohttp lists 308 as a code for "Resume Incomplete", but some hosts use it for "Permanent Redirect".
It seems that the earlier is nonstandard, but I'm not too sure.

References: [httpstatuses](https://httpstatuses.com/308), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308), [RFC7538](https://tools.ietf.org/html/rfc7538)

emillon:
This updates the JSON files that describe the various HTTP codes.
They are sourced from the following URL (last updated 2014):

https://github.com/arunshejul/httpstatus.es/tree/master/codes

Note that since code is generated from this, this will break API.

Closes #710

### DIFF
--- a/cohttp/scripts/codes/3.json
+++ b/cohttp/scripts/codes/3.json
@@ -168,11 +168,11 @@
         },
         "308":{
             "code":"308",
-            "title":"Resume Incomplete",
-            "summary":"resumable HTTP requests",
+            "title":"Permanent Redirect",
+            "summary":"connect again to a different URI using the same method",
             "descriptions":{
                 "wikipedia":{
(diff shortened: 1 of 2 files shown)


mseri:
Permanent redirect is the standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308 (there is also a link to the specification later in the page). I think that this is the correct thing to do, even though google drive uses it in a non-standard custom way.

PR #712, opened 2020-09-09
Cohttp_async.Body.is_empty should return true iff pipe is empty

anuragsoni:
Before this change i was seeing the following:

```ocaml
Body.is_empty (`Pipe (Pipe.of_list [""; "foo"])) = true
Body.is_empty (`Pipe (Pipe.of_list [])) = false
```

### DIFF
--- a/cohttp-async/src/body_raw.ml
+++ b/cohttp-async/src/body_raw.ml
@@ -29,11 +29,11 @@ let is_empty (body:t) =
   | `Pipe s ->
     Pipe.values_available s
     >>| function
-    |`Eof -> false
+    |`Eof -> true
     |`Ok ->
       match Pipe.peek s with
-      | Some "" -> true
-      | Some _ | None -> false
+      | None -> true
+      | Some _ -> false
 
 let to_pipe = function
   | `Empty -> Pipe.of_list []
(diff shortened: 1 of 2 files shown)


anuragsoni:
This should align the behavior with Cohttp_lwt's Lwt_stream handling. 

On a related note, I also noticed that cohttp considers `String ""` as an empty body, but `Strings [""]`, `Stream (Lwt_stream.of_list [""])`, and now `Pipe (Pipe.of_list [""])` will not be considered as empty. I'm wondering if it'd make sense to treat the various forms of `list [""]` bodies as empty as they don't contain any bytes?

mseri:
I agree that it would make sense to consider also empty lists as empty. Would you mind adding the change?

mseri:
As an aside, please have a look at https://github.com/mirage/ocaml-cohttp/pull/692 Any feedback welcome.

anuragsoni:
> I agree that it would make sense to consider also empty lists as empty. Would you mind adding the change?

@mseri I will put up a PR for discussion, but i think it might be better to do that as a separate PR so this commit can be about fixing the bug in the async Body module. What do you think? I will be able to put up a PR later today or tomorrow.

Thanks!

mseri:
Sounds good to me

PR #713, opened 2020-09-10
filter empty strings in list/pipe/stream

anuragsoni:
This is more of a question about what behavior should be expected for body. Right now Body.is_empty returns false if the body is comprised of `[""], [""; ""]` etc. Should this be considered empty? If so, it'll be difficult to enforce this without making the types for body private and enforcing that bodies get created via the public api. 

This might be a scenario that only happens if someone misuses the api, but the fact that it can happen made me feel like at-least bringing this up as a question.

### DIFF
--- a/cohttp-lwt/src/body.ml
+++ b/cohttp-lwt/src/body.ml
@@ -75,7 +75,7 @@ let drain_body (body:t) =
 
 let of_string_list l = `Strings l
 
-let of_stream s = `Stream s
+let of_stream s = `Stream (Lwt_stream.filter (fun s -> s <> "") s)
 
 let transfer_encoding = function
   |#Body.t as t -> Body.transfer_encoding t
--- a/cohttp-lwt/src/body.ml
+++ b/cohttp-lwt/src/body.ml
@@ -42,7 +42,10 @@ let create_stream fn arg =
 let is_empty (body:t) =
   match body with
   | #Body.t as body -> return (Body.is_empty body)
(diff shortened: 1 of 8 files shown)


cohttp-lwt/src/body.ml
```diff
@@ -75,7 +75,7 @@ let drain_body (body:t) =
 
 let of_string_list l = `Strings l
 
-let of_stream s = `Stream s
+let of_stream s = `Stream (Lwt_stream.filter (fun s -> s <> "") s)
```
mseri, line 78: Can this, or the equivalent Pipe version, hang due to a broken infinite stream of empty lists?

mseri:
This is a good point. I think it is reasonable to consider it empty and I agree that it does not seem enforceable. Even if we llow the construction from a different set of APIs, the stream of data we receive may as well have the broken structure.

I think the fix you suggest is the most reasonable thing to do.

mseri:
I'd also like ping @brendanlong in this discussion

cohttp-lwt/src/body.ml
thedufer: Surely no more than it would hang if the first element in the pipe is indefinitely delayed, I would think.

anuragsoni:
I just pushed an update + tests for consuming leading empty strings for Lwt_stream and Pipe

mseri:
I think this is good. If we notice that it excessively affects performances, we can have another look later on

PR #714, opened 2020-09-14
don't change the pipe when creating async body

anuragsoni:
Leftover from #713. I missed this earlier, but now that we are
consuming empty strings when we call `is_empty`, we don't need
to change the pipe when constructing the async body.

The Lwt body's `of_stream` and cohttp's `of_string_list` were left untouched already so no change needed there.


mseri:
Well spotted! Thanks

PR #715, opened 2020-09-23
Fix chunked encoding of empty bodies.

mefyl:
An empty (zero length) chunk signals the end of an HTTP chunked body encoding. When uploading an empty body or a strings body with an empty string somewhere, that chunked would be sent as a zero length chunk, ending the body per the protocol, and the subsequent chunks (or the final 0 length chunk) would become trailing garbage.

While this is incorrect for one-shot requests but can be recovered from in the parser is permissive, it is absolutely breaking in keep-alive connections as the rest of the body is being parsed as the next HTTP header.

### DIFF
--- a/cohttp/src/transfer_io.ml
+++ b/cohttp/src/transfer_io.ml
@@ -80,9 +80,14 @@ module Make(IO : S.IO) = struct
 
     let write oc buf =
       let len = String.length buf in
-      write oc (Printf.sprintf "%x\r\n" len) >>= fun () ->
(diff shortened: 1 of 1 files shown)


samoht (approved):

mseri:
If you find a nice way to test it, I would be happy to get tests for i

mefyl:
Doing any chunked request with `Cohttp_async.Body.of_string ""` as body should trigger it.

PR #717, opened 2020-10-17
Simplify cohttp-lwt-jsoo

mseri:
I was trying to use the package and found rather confusing that one does not use Cohttp_lwt_jsoo and instead uses Cohttp_lwt_xhr.

The documentation already makes it clear that the module uses an XHR request, so I don't really see the need for this discrepancy.

With the upcoming cohttp 3.0 breakage, I think it is worth adding this other small breaking change.


mseri:
I don't really know who is using this package, so I am not sure whom to pin...

dinosaure (approved):
LGTM for me.

avsm (approved):
Good idea! The only other user I'm aware of is the github-jsoo package, but it's easy to adapt the code of course.   A note in the CHANGES file about this would be great.

mseri:
Done. I have updated CHANGES with all the changes since the last release. Thanks

PR #721, opened 2020-10-22
Fix transfer-encoding in reverse order

### LINKED ISSUE #695: Values of transfer-encoding header  are written in reverse order
Just a small note here (that follows one of the points of the issue #501): the duplicate headers are in reversed order (for a given header, the last added value is the first one written). That is an issue for the `transfer-encoding` header as this order is the one in which the corresponding transformations are applied to the body.

**Example**:
```ocaml
open Header
let h = of_list ["transfer-encoding", "gzip"; "transfer-encoding", "chunked"];;

print_endline (to_string h);;
(* transfer-encoding: chunked
   transfer-encoding: gzip *)
print_endline (match get h "transfer-encoding" with None -> "" | Some v -> v);;
(* chunked,gzip *)
````

Note that this order is right for other headers.

mseri:
This should fix #695


lyrm:
Hi ! Everything seems alright ! Thanks!

Issue #719, opened 2020-10-22
COHTTP_DEBUG CONDUIT_DEBUG do no longer work


mseri:
It used to be that to have extra debug information, it was enough to do the following

```
You can activate some runtime debugging by setting `COHTTP_DEBUG` to any
value, and all requests and responses will be written to stderr.  Further
debugging of the connection layer can be obtained by setting `CONDUIT_DEBUG`
to any value.
```

This does no longer work in conduit/cohttp 3.0. What is the alternative?

dinosaure:
`COHTTP_DEBUG` still exists on `cohttp-lwt-unix-nossl` (which is used by any other LWT packages) and nothing change for `cohttp-async`. As far as I can tell, we should delete the last sentence where `conduit` does not have such mechanism (but it still uses `logs`).

mseri:
With `Logs` is it enough that we do what you were mentioning a few days ago?
```
let () = Fmt_tty.setup_std_outputs ~style_renderer:`Ansi_tty ~utf_8:true ()
let () = Logs.set_reporter (reporter Fmt.stderr)
let () = Logs.set_level ~all:true (Some Logs.Debug)
```

Would it be hard to modify cohttp-lwt to use the same mechanism?

mseri:
I was trying with both variables, including COHTTP_DEBUG, and did not have any logs earlier today

dinosaure:
> I was trying with both variables, including COHTTP_DEBUG, and did not get any logs earlier today (even if cohttp-server-lwt outputs Cohttp debugging output is active, there is nothing else printed while accessing the server)

I think the bug is about the implementation of the `default_reporter` (see `cohttp-lwt-unix-nossl/src/debug.ml`) which does a deadlock at one point... Let me figure out :+1:.

dinosaure:
And which output you expect? I mean, from the current code (and I did not deleted debug information I think), for the server, only a TLS server will gives you some output from `conduit` - where `conduit-lwt` (TCP/IP impl.) does not have any logs - and only when we got an error (404 does not print anything).

On other side, it sure that on the client side we have a deadlock.

mseri:
It is likely that my memory is wrong here. I thought to remember that the server at least was spitting a long list with the calls received and the response sent, with a nice detail of requests and responses metadata

PR #722, opened 2020-10-23
Fix logs

dinosaure:
This patch wants to fix logs, however, if we look on `conduit-lwt-server`, if we execute it with `COHTTP_DEBUG`, it seems that the verbose option takes the lead on which level should be printed. In other words, if you want logs on `cohttp_lwt_server` you must (even if you prepend with `COHTTP_DEBUG`) use `-vvv`.

This is a fix for #719


mseri:
Ugh, that was nasty!
Thanks for the fix.

PR #725, opened 2020-10-27
Media params

dinosaure:
Revive #542


mseri:
Do you want to add improve compliance to the specs as per the comments in the other PR?

dinosaure:
> Do you want to add improve compliance to the specs as per the comments in the other PR?

I don't have enough time to go further on this specific task, sorry 😕 .

mseri:
It's perfectly fine, I just wanted to hear your opinion on this and your plan, you have more experience on this. If you think the change is worth having, it looks good to merge to me

dinosaure:
This PR is more convenience for the end-user who does not really care about the difference between the quoted-string or the token on the MIME type value (which are semantically the same). However, I can see some limitations about the way to parse the Content-Type (and `typebeat` is more in conformance to the RFC).

I think, the better is to make an issue about the parsing of the Content-Type which does not strictly follow the RFC. Then, as is, this PR can be merged.

mseri:
Done: https://github.com/mirage/ocaml-cohttp/issues/726
Thanks

PR #734, opened 2020-11-07
Remove wrapped false

rgrinberg:
It should be avoided in non-legacy libraries


mseri (approved):

mseri:
I suggest to include this into 3.0.0-1 which entirely replaces 3.0.0

mseri:
It is only a minor change, only two of the libraries are exposed

PR #738, opened 2021-01-19
lwt_jsoo: Forward exceptions to caller when response is null.

### LINKED ISSUE #737: Jsoo implementation fails when the `xhr.response` field is null.
Even when the status of the `XmlHttpRequest` is `DONE`, the `response` field can be null in some scenarii:
* The request failed CORS. The browser will not let you access the response content for security reasons.
* The request was aborted because of a disconnection (on Chrome / v8 at least).

cohttp-lwt-jsoo is not prepared for this and reads `responseText` if `response` is not an arraybuffer, raising an exception from the onreadystagechange callback which is thus not catchable by the caml caller and escapes directly to the javascript toplevel (see attached console logs).

I can easily submit a fix for this, but would like to inquire about the preferred approach API-wise:
* Change the `call` API to return a `_ Result.t Lwt.t` since it actually can fail. My preferred solution, but I don't suppose changing the API is an option.
* Forward the javascript exception to the caller unwrapped whatsoever, so she can at least catch it.
* Wrap the exception in a neat Ocaml exception.

# Console output examples

Connection dropped:
image

CORS:
image

mefyl:
Fix #737. This is the very minimal approach, were the javascript exception is forwarded as-is to the caller.

### DIFF
--- a/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
+++ b/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
@@ -189,43 +189,48 @@ module Make_client_async(P : Params) = Make_api(struct
         Js.wrap_callback
           (fun _ ->
              match xml##.readyState with
-             | XmlHttpRequest.DONE -> begin
-                 (* construct body *)
-                 let body =
-                   let b =
-                     let respText () = 
-                       Js.Opt.case xml##.responseText (fun () -> `String (Js.string ""))
-                         (fun s -> `String s) in
-                     if xhr_response_supported then
-                       Js.Opt.case
-                          (File.CoerceTo.arrayBuffer xml##.response)
-                          (fun () -> Firebug.console##log
-                             (Js.string "XHR Response is not an arrayBuffer; using responseText");
-                             (respText ()))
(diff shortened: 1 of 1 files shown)


mseri:
Thanks! Since we cannot really change the signature of call to return error without larger scale refactoring, I think this is a reasonable compromise

mseri:
The failure is due to
```
cohttp-lwt-unix-nossl.opam: changes needed:
  "ppx_sexp_conv" {>= v0.14.1}             [from cohttp-lwt-unix-nossl/src]
cohttp-lwt-unix-ssl.opam: changes needed:
  "ppx_sexp_conv" {>= v0.14.1}             [from cohttp-lwt-unix-ssl/src]
cohttp-lwt-unix.opam: changes needed:
  "ppx_sexp_conv" {>= v0.14.1}             [from cohttp-lwt-unix/src]
cohttp-lwt.opam: changes needed:
  "uri" {>= 4.1.0}                         [from cohttp-lwt/src]
cohttp-mirage.opam: changes needed:
  "ppx_sexp_conv" {>= v0.14.1}             [from cohttp-mirage/src]
  ```

PR #739, opened 2021-01-27
Add GitHub Actions workflow

smorimoto:
We need benchmarks on some large projects so if you all don't mind, merge this.

Note for general user: the actions-ml project is still not stable enough for the general user to try. ​Don't think about using it at the very moment.


mseri:
Absolutely! It's never bad to have extra checks. Thanks!

PR #742, opened 2021-02-02
Adapt to conduit-mirage 2.3.0

samoht:
Adapt to https://github.com/mirage/ocaml-conduit/pull/376

### DIFF
--- a/cohttp-mirage/src/client.ml
+++ b/cohttp-mirage/src/client.ml
@@ -18,40 +18,58 @@
  *)
 
 open Lwt.Infix
-module Channel = Mirage_channel.Make (Conduit_mirage.Flow)
-module HTTP_IO = Io.Make (Channel)
 
-module Net_IO = struct
-  module IO = HTTP_IO
+module Make
+    (P : Mirage_clock.PCLOCK)
+    (R : Resolver_mirage.S)
+    (S : Conduit_mirage.S) =
+struct
+  module Channel = Mirage_channel.Make (S.Flow)
+  module HTTP_IO = Io.Make (Channel)
+  module Endpoint = Conduit_mirage.Endpoint (P)
 
-  type ctx = { resolver : Resolver_lwt.t; conduit : Conduit_mirage.t }
+  module Net_IO = struct
+    module IO = HTTP_IO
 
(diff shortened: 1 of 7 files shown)


dinosaure:
I would like to prefer to keep `Cohttp_mirage.Make` to be able to start a CoHTTP server without the Conduit plumbing - and, by this way, keep `Cohttp_mirage.Server_with_conduit`.

dinosaure:
This update becomes incompatible with `git-cohttp-mirage` where the `ctx` should not be a result of a _functor_. By design and initially, the `ctx` is defined outside a _functor_ to let the end-user to use/fill/create it even if the `Client` (_functor_ or not) is not available.

It's hard to follow the implication of that when I suspect some incompatible types somewhere with MirageOS/Functoria but I can not make a minimal example of it.

Note that `git-cohttp-mirage` can be easily updated - and _functorized_ over `PCLOCK`/`RESOLVER`/`CONDUIT` on the other hand.

samoht:
The current state of mirage-conduit (where first-class modules are mixed with functors) is super confusing. The current design simplify a lot of things (and allows to add http_client combinators to the mirage DSL). I will have a look at `git-cohttp-mirage` to see how hard it is to adapt.

mseri:
If we revert the last commit and run ocamlformat, this should also be good to go

samoht:
commit reverted + ocamlformat applied

cohttp-mirage/src/client.ml
```diff
@@ -18,40 +18,58 @@
...
+    let default_ctx =
+      { resolver = R.localhost; conduit = None; authenticator = None }
 
-  let close_in _ = ()
-  let close_out _ = ()
+    let connect_uri ~ctx:{ resolver; conduit; authenticator } uri =
+      R.resolve_uri ~uri resolver >>= fun endp ->
+      Endpoint.client ?tls_authenticator:authenticator endp >>= fun client ->
+      match conduit with
+      | None -> failwith "conduit not initialised"
```
mseri, line 49: Is it ok to use failwith instead of Lwt.raise or Lwt.failwith?

dinosaure: `Lwt.fail` does not record the backtraceI believe

mseri (approved):

PR #743, opened 2021-02-05
Disable opam sandboxing in GitHub Actions for dune cache

smorimoto:
This will speed up the opam install process, so the workflow will be more fast by up to 3 minutes in this project.


avsm:
We actually do want to test with sandboxing, to ensure that the package isn't accidentally trying to write outside of it.

smorimoto:
Yes, I thought the same, so I opened this PR instead: https://github.com/mirage/ocaml-cohttp/pull/744

PR #745, opened 2021-02-05
Remove dependency to base

samoht:
This is causing issues to link mirage unikernels:

```
ld: _build/main.native.o: in function `camlBase__Import0__entry':
/home/opam/.opam/4.10/.opam-switch/build/base.v0.14.1/_build/default/src/import0.ml:358: undefined reference to `Base_am_testing'
ld: _build/main.native.o: in function `camlBase__Import0__gc_roots':
(.data+0x82ddb8): undefined reference to `Base_am_testing'
```

Also we don't need this to generate field accesses.


mseri:
This (partially?) reverts https://github.com/mirage/ocaml-cohttp/pull/686, can we keep compare or that requires base for some reason?

samoht:
Why can't compare just work? Happy to specialize it by type but what does it bring?

mseri:
Can you directly compare requests and responses using Stdlib.compare? I thought it was not directly possible without, which I think was the reason of that PR

samoht:
Ha yes it needs special comparators to make sure the header maps are compared using the right equality. Will re-expose these functions too.

samoht:
I've now re-exposed `Response.compare` and `Request.compare`.

mseri (approved):

cohttp/test/dune
```diff
@@ -1,6 +1,7 @@
 (executable
  (name test_accept)
  (modules test_accept)
+ (forbidden_libraries base)
```
mseri, line 4: Do we need to update dune lower bound to use this?

samoht: good point, I've just bumped it in the opam file (the dune-project file was already using the right version)

mseri (approved):
Do you want to add something? If not I think this can be merged (don't forget to add a line to the changelog)

Issue #726, opened 2020-10-28
Parsing of Content-Type does not strictly follow the RFC


mseri:
For reference see https://github.com/mirage/ocaml-cohttp/pull/725 and the discussion in https://github.com/mirage/ocaml-cohttp/pull/542

dinosaure:
If someone wants to work on this part of CoHTTP, he should take a look on [typebeat](https://github.com/mirage/typebeat) (which will be archived!). The work of `typebeat` comes from what we did about [mrmime](https://github.com/mirage/mrmime) (email and HTTP share the same RFC about `Content-Type` - see [RFC 2045 § 5](https://tools.ietf.org/html/rfc2045#section-5)).

Note that, even if the format seems quite close, `Accept{-*}` field has another format described in [RFC 7231 § 5.3](https://tools.ietf.org/html/rfc7231#section-5.3) which is **not** handle by `typebeat`.

A final note is about `FWS` and `CFWS` token which can be the part of values (according [RFC 7230 § 3.2.4](https://tools.ietf.org/html/rfc7230#section-3.2.4)), Even if it's considered as an **obsolete** form of values, you should handle such layout. You should look into [unstrctrd](https://github.com/dinosaure/unstrctrd) which _folds_ these kind of whitespace - `mrmime` uses it to pre-process the value and delete all `FWS`/`CFWS`.

PR #703, opened 2020-07-22
Improve Header.update

mseri:
`Map.update` was introduced only in OCaml 4.06: the code has been updated to support older versions of the compiler, and the docstring has been improved.

See https://github.com/ocaml/opam-repository/pull/16849

PR #747, opened 2021-02-08
Associative list implementation for headers

lyrm:
## Purpose
Right now, headers are represented by a map, which is good to automatically insure certain RFCs properties but is also quite counterintuitive when debugging as it changes the headers order and enforces RFCs without enabling the user to control it.

The purpose of this PR is:
1. to propose a new header implementation
2. to remove every changes on the headers that is not explicitly asked for by the user. That includes, for example, the automatic addition of some headers, that occurs in various modules and can be quite hard to track. To keep some retro-compatibilities those header modifications are either grouped under a new function, that the user can call if needed, or  placed under a condition that the user can explicitly notified. 

### Main changes
Implementing the header with an associative list basically enables to keep the order in which the header are transmitted and so to maintain the invariant  ```to_list (of_list h) = h ```. 

That means: 
- ```add``` function does neither remove nor replace any headers;
- ```get``` function **only retrieves the last transmitted value**. To get all values associated to a key, the functions ```get_multi``` or ```get_multi_concat``` must be used;
-  ```map``` and ```iter``` types have changed to match the new header type; 

### Update function
There are two functions ```update``` and ```update_last``` now. Basically ```update h k f``` calls ```f``` on the concatenated values associated with ```k``` and affects all occurrences of ```k``` (meaning, for example, all are removed if the value returns by ```f``` is ```None```).

In opposition, ```update_last``` works with the last value associated to ```k``` and affects only this value. 

### New functions
- ```get_multi_concat``` concatenates all values associated to a key. The ```~list_value_only``` optional argument enables to get the last value only if the searched header is not a list-value header and all the values otherwise. Though, ```get ~list_value_only:true``` enables to get the same result that the previous ```get``` function.
- ```clean_dup``` enables to remove unauthorized duplicates and concatenated values of list-value headers. In other word, it does the same kind of clean up that a map automatically does but need an active call by the user and preserved order of transmission.

### Related changes
- ```Cohttp/test/header_test.ml``` and ```Cohttp_lwt_unix/test/test_parser.ml``` : some tests have been added and previous ones have been correcting accordingly to the new implementation.
-  ```Cohttp_lwt_jsoo.ml``` and ```Cohttp_curl_async.ml```: the few changes are due to ```iter``` and ```map``` new types.

## Removing and factoring header modifications
WIP

### DIFF
--- a/cohttp/fuzz/fuzz_header.ml
+++ b/cohttp/fuzz/fuzz_header.ml
@@ -73,7 +73,8 @@ let list_value_header_gen =
   let printer fmt str = pp fmt "%s" str in
   with_printer printer gen
 
-(** Generate a tchar following {{https://tools.ietf.org/html/rfc7230#appendix-B}RFC 7230}.
+(** Generate a tchar following
+    {{:https://tools.ietf.org/html/rfc7230#appendix-B} RFC 7230}.
 
     tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" /
     "_" / "`" / "|" / "~" / DIGIT / ALPHA *)
--- a/cohttp/fuzz/fuzz_header.ml
+++ b/cohttp/fuzz/fuzz_header.ml
@@ -1,0 +1,562 @@
+(*{{{ Copyright (c) 2021 Carine Morel <carine@tarides.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *}}}*)
+
+module H = Cohttp.Header
+
+(** Here, we test the Header module with fuzzing. Some of these tests may be
+    redundant with Alcotest tests.
+
+    The tests are launched with [dune runtest] but can also be run with [afl]
+    with the command line : [dune build @cohttp/fuzz/fuzz --no-buffer].
+
+    The tests below reflects the semantics we want for each function, however in
+    some cases, it may actually be specific to the current implementation and
+    does not necessary need to be enforced in future implementations. To make it
+    clear, tests are annoted by their categories:
+
+    - FS (Functions semantics): tests the semantics described in the
+      documentation.
+
+    - SI (Specific to current Implementation): these tests are here to check the
+      implementation is doing what we thing it is doing but may change
+      accordingly to implementation changes. *)
+
+(* Generators *)
+let list_value_headers =
+  [|
+    "accept";
+    "accept-charset";
+    "accept-encoding";
+    "accept-language";
+    "accept-ranges";
+    "allow";
+    "cache-control";
+    "connection";
+    "content-encoding";
+    "content-language";
+    "expect";
+    "if-match";
+    "if-none-match";
+    "link";
+    "pragma";
+    "proxy-authenticate";
+    "te";
+    "trailer";
+    "transfer-encoding";
+    "upgrade";
+    "vary";
+    "via";
+    "warning";
+    "www-authenticate";
+  |]
+
+(** Pick a random list-value header name from a predefined array of values. *)
+let list_value_header_gen =
+  let open Crowbar in
+  let gen =
+    map
+      [ range (Array.length list_value_headers) ]
+      (fun i -> list_value_headers.(i))
+  in
+  let printer fmt str = pp fmt "%s" str in
+  with_printer printer gen
+
+(** Generate a tchar following {{https://tools.ietf.org/html/rfc7230#appendix-B}RFC 7230}.
+
+    tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" /
+    "_" / "`" / "|" / "~" / DIGIT / ALPHA *)
+let tchar_gen =
+  let tchar_code_gen =
+    let uppercased_letter = Crowbar.range ~min:65 26 in
+    let lowercased_letter = Crowbar.range ~min:97 26 in
+    let others =
+      List.map
+        (fun i -> Crowbar.const i)
+        [
+          33 (* ! *);
+          35 (* # *);
+          36 (* $ *);
+          37 (* % *);
+          38 (* & *);
+          42 (* * *);
+          43 (* + *);
+          45 (* - *);
+          46 (* . *);
+          94 (* ^ *);
+          95 (* _ *);
+          96 (* ` *);
+          124 (* | *);
+          126 (* ~ *);
+        ]
+      |> Crowbar.choose
+    in
+    let digit_and_others = Crowbar.(choose [ others; range ~min:48 10 ]) in
+    Crowbar.(choose [ lowercased_letter; uppercased_letter; digit_and_others ])
+  in
+  Crowbar.(map [ tchar_code_gen ] (fun i -> Char.escaped (Char.chr i)))
+
+(** Generate a non-emoty word of arbitrary length (composed of letters only). *)
+let word_gen =
+  let open Crowbar in
+  let gen =
+    fix (fun word_gen ->
+        choose
+          [
+            (* one letter word *)
+            tchar_gen;
+            (* two letters word *)
+            map [ tchar_gen; tchar_gen ] (fun l1 l2 -> l1 ^ l2);
+            (* add one letter *)
+            map [ tchar_gen; word_gen ] (fun l w -> l ^ w);
+          ])
+  in
+  let printer = pp_string in
+  with_printer printer gen
+
+(** Generate an header name: either a predefined list-value header or a random
+    word *)
+let header_name_gen =
+  let open Crowbar in
+  let gen = choose [ list_value_header_gen; word_gen ] in
+  let printer = pp_string in
+  with_printer printer gen
+
+let header_printer fmt (k, v) = Crowbar.pp fmt "%s, %s" k v
+
+(** Generate a header key/value pair *)
+let header_gen : (string * string) Crowbar.gen =
+  let open Crowbar in
+  let gen_setcookie = pair (const "Set-cookie") word_gen in
+  let gen_otherheader = pair header_name_gen word_gen in
+  let gen =
+    (* one in ten generate header is a "set-cookie" header *)
+    choose (gen_setcookie :: List.init 9 (fun _ -> gen_otherheader))
+  in
+  with_printer header_printer gen
+
+(** Generate a list of headers *)
+let header_list_gen : (string * string) list Crowbar.gen =
+  let open Crowbar in
+  let gen = list header_gen in
+  let printer = pp_list header_printer in
+  with_printer printer gen
+
+(** Generate a [Cohttp.Header.t] headers. *)
+let headers_gen : H.t Crowbar.gen =
+  let open Crowbar in
+  let gen =
+    fix (fun headers_gen ->
+        choose
+          [
+            (* empty header *)
+            const (H.init ());
+            (* add one pair (k, v) *)
+            map [ header_gen; headers_gen ] (fun (k, v) h -> H.add h k v);
+            (* add a list of headers *)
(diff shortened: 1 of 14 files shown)


avsm:
Many thanks for this PR, and in particular for all the cleanups to header.mli along with your changes. A few quick questions:

- how much does this change the semantics for existing users of Cohttp.Header, es? From a look at the interface, only the `iter` and `map` functions seem to have different signatures any more.  A note in CHANGES about what existing users of Cohttp can expect to adapt would be useful, particularly if there are semantic changes that are not in the exposed types.  To be clear though: I'm in support of this change, but just want the "compatibility" functions you mention in your introduction text to be clearly marked for users to migrate more easily.

- what motivates this change -- you mention some RFC compatibility. It would be good to list some of those, so that we can perhaps add specific test cases to cover that behaviour.

dinosaure:
I would like to re-contextualize this part of CoHTTP when several _recent_ PRs are already merged:
- A correction about [`Header.replace`](https://github.com/mirage/ocaml-cohttp/pull/694)
- A [revert](https://github.com/mirage/ocaml-cohttp/pull/702) about the correction
- An improvement about [`Header.update`](https://github.com/mirage/ocaml-cohttp/pull/703)

So I follow @avsm about the migration process when we currently don't have a good test suit to check if we break something or not (as highlighted by #701).

avsm:
Very good point. One option would be to move the existing implementation to Cohttp.Header.V1, add Cohttp.Header.V2, alias Cohttp.Header to Cohttp.Header.V1 and then let users use the new V2 implementation and gradually migrate.  In a few releases, we can add a deprecation warning to Cohttp.Header.V1 and remove that code.

cohttp/src/header.mli
```diff
@@ -14,96 +14,142 @@
...
+
+val to_list : t -> (string * string) list
+(** [to_list h] convert HTTP headers h to a list. Order is preserved. *)
 
 val init_with : string -> string -> t
-(** Construct a fresh map of HTTP headers with a single key and value entry. *)
+(** [init_with k v] construct a fresh map of HTTP headers with a single pair of
+    key and value [(k, v)]. *)
 
+(** [add h k v] adds a key and value to an existing header list. *)
```
mseri, line 40: Double docstring

mseri:
The option of having V1 and V2 sounds like a good idea. Gives time to people to migrate and we can do it gradually without worrying about subtle semantics changes

samoht:
Instead of complexing the API (with V1 and V2) can't we just make sure we get the "correct" API? :-) The changes seems like a good improvement to me and they come with proper coverage testing. I guess the only missing thing is a CHANGE entry describing the semantic changes (and if these changes are too big, maybe we can try to rename some functions to avoid breakages).

lyrm:
Thanks for all your comments. I will add note in the CHANGE entry and comments about related RFCs. 

I'm working on adding fuzzing tests as it seems quite appropriate for this module and would enable to get a proper test suite to describe and maintain the important invariants (especially the ones linked to RFCs). 

I am also writing a short guide to explain the differences with the previous implementation and how to use the new functions (for example, ```get_multi_concat``` can be used to get the same result as the previous ```get``` function).

I don't really have a strong opinion about the V1/V2 idea. If a major release is expected soon, maybe we can just wait for it ?

lyrm:
(I just made a force-pushed because I did not used rebase the first time. Sorry !)

samoht:
Is `get` the only function whose semantics is changing? Maybe we can keep `get` the way it was before and add `get_one` (or something similar) for the new functions? This way we don't break existing users silently.

avsm:
@lyrm a big +1 for fuzz testing! I agree with @samoht that for the moment, if you can _add_ functions that do not change the semantics, we can migrate much more easily to a new major version.  The sequence could be:

- add your fuzz testing immediately and the new implementation, with the same external interface
- add your new functions (e.g. get_one) to expose the new semantics
- create a V1 and V2 module, with V1 being the existing code, and V2 being the new interfaces
- release a minor version of cohttp with these new modules exposed
- migrate existing users of cohttp to the V1 module, which is hopefully quite a mechanical change

With this scheme, we shouldn't need a major revision bump in the short term, since new users can switch to V2 in their own time.

samoht:
If we are extra careful to not change the semantics of existing functions I don't think we need to add the V1/V2 complexity. If that's not possible, I agree that we should make sure users are aware of that semantic changes and breaking compilation of their program is the best move :-)

lyrm:
The initial idea was to make the header module more predictable (no unrequested changes in header order for example) which is mainly done by changing the header type from a map to an associative list. 

However, it seemed also very appropriate to change the ```get``` function that was actually doing more that just getting the values bound to a key : 
- if the searched header was a list-value header, it was returning all the values concatenated;
- otherwise, if was returning only the last added value. 

So the issue I have with the previous ```get``` function is rather that it does an invisible check ("is the researched header a list-value one?").  Do we want the ```get``` function to still do that ? 

There is an other issue related to the original implementation of ```get``` that needs to be mentioned as corrected it may break a few things: in the case of a list-value header, the order in which the values are concatenated is actually in reverse transmission order. For example : 
```ocaml 
let h = add  (init ()) "accept" "text/*" in
let h = add h "accept" "application/xml in
get h "accept"
```
was previously returning "application/xml, text/*". You get the same result if you create headers from parsing this:
```
Accept: text/*
Accept: application/xml
```
Also, RFC [7230#section-3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2) says that order matters. 

It was corrected for the particular case of the ```transfer-encoding``` header, for which order is particularly important,  in PR #721 (see RFC [7230#section-3.3.1](https://tools.ietf.org/html/rfc7230#section-3.3.1)).

As it is the ```add``` function that create this order issue, other functions, like ```to_lines``` or ```to_frames``` or ```to_string``` do the same. I think we should definitely correct that part.

There are no other important semantic changes except the ```update``` function that depends on ```get```. However there are invariants that have changed or have been added (mostly related to the headers order). For example: ```to_list (of_list l) = l``` is true with the new implementation. 

(and there are other particular cases, like ```map``` and ```iter```, that have changed type and other functions ). 

And obviously, some tests have been changed to match the fact that the previous implementation did not preserve headers order. So I am not sure on how to do the V1/V2 idea properly.

mseri:
I have two suggestions, I'd like to hear @lyrm, @avsm and @samoht opinion on this (and of any other interested person, maybe @hannesm?). It seems to me that the semantics change is there to fix an issue of the api and that the change of some types will also likely break old code. I am no longer sure it is worth going light and keeping two copies of the code and tests for V1 and V2, and try to adjust the code to support both.

Since we will be releasing 3.1, the first new (available) major version. Why don't we just clearly emphasize the breaking changes in the release, leaving ample space to the issue of types and semantics in the headers module (we could even have a post-install message in this release for double safety), and ask users of the libraries to update code?

If we want to wait to have the fuzzing before actually releasing this, why don't we release `3.1.0~beta1` including this change and warning about the api changes, and try to get the fuzzy tests in as soon as possible with a subsequent 3.1.0 release?

hannesm:
Since I got mentioned here, my opinion is that versioning is to be done on the opam package level, not on the module / API level. I do not think that two (multiple) API V1/V2 solve anything or are convenient. In contrast, they imply more code and are rather confusing (when will V1 be removed (will it ever be)) - leading to more maintenance overhead and more complex code someone has to read and review.

If there's no "versioned API", this means for existing clients of this library: they can move forward and adapt version 3, including the new API. For new clients: they are not confused about which version to use (documentation is easier to read and understand).

samoht:
I agree with Hannes on this one :-) V1/V2 could makes sense for long-lived APIs where we want clients to use both. Here I feel this is not needed if the changes are copiously documented.

avsm:
Thanks for all the extra discussion.  The reason I pushed the V1/V2 suggestion is that it helps with the situation of having a version upgrade of cohttp _silently_ (i.e. no build failure) changing the semantics of a program.  Consider if (for example) something adding a security-critical CORS header changed in some way because of this API change that caused it to not emit it.   With V1/V2, there is a decision from the application author to move their application to the new behaviour that is independent of the cohttp version.  It is indeed more heavyweight, but for good reason -- HTTP header handling is a pretty important of any non-trivial webserver.

I'd be happy to approve this PR without V1/V2 if we could get a clear CHANGES entry in this PR that distills down @lyrm's [comment above](https://github.com/mirage/ocaml-cohttp/pull/747#issuecomment-783431396) into a programmer-facing guide for what they need to check for in their code when migrating to this new interface.  Then, we can run this CHANGES entry past (for example) @talex5 and the https://github.com/ocurrent/overview stack and just check that there are no surprises with upgrading.

If it's the case that upgrading can be done cleanly by following the CHANGES entry (and I do now believe this to be the case thanks to @lyrm's clarifications above), I am fine to move ahead with the improvements in this PR and get them into a release soon.

hannesm:
To avoid "silently changing semantics underneath", another option is to rename all the functions or the module (plus the mentioned CHANGES entry), so nobody will accidentally use the new semantics.

talex5 (commented):
OCurrent does very little with headers, so probably isn't much of a test.

Looking at the diff, it's a bit hard to see exactly what changed here. It seems concerning that the unit-tests had to be changed so much!

I wonder if we should just expose the list type directly and let users manage the headers using OCaml's `List` module. e.g. it's not clear from the name whether `add` adds to the end of the headers (which sounds like it would be slow), or prepends (which is fast, but maybe unexpected). But it would be clear with something like:

```ocaml
let headers = Header.[
  v "content-type" "text/plain";
  v "accept" "foo";
]

**Comment by lyrm** [comment]
I finally pushed the fuzzing tests as well as the change log and some improvements in the header module doc.

**Comment by samoht** [comment]
Thanks @lyrm !

By looking at the API again, I am wondering if that could work if we turn headers into an heterogeneous list with typed keys to know if an entry is a list of string of just a string. Are the header entries typed in the RFCs? Would that make sense to ensure it in the header module? The goal would be to only have one `get` function. Something along the lines of:

```
val content_range : int64 key
val media_type : string
val connection_close : bool
val get: t -> 'a key -> 'a

mseri:
Fantastic! @lyrm if you run `dune build @fmt --auto-promote` and commit the changes, the build should succeed

mseri:
> By looking at the API again, I am wondering if that could work if we turn headers into an heterogeneous list with typed keys to know if an entry is a list of string of just a string. Are the header entries typed in the RFCs?

This sounds like a very nice idea and it likely also make it nicer, later on, to address https://github.com/mirage/ocaml-cohttp/issues/726

dinosaure:
> This sounds like a very nice idea and it likely also make it nicer, later on, to address #726

For a concrete example of such thing, `mrmime` implements this way with the [`Field` module](https://github.com/mirage/mrmime/blob/master/lib/field.mli). As far as I can tell and from my usage in several projects (such as `ocaml-dkim` or `ptt`), it supports many benefits (mainly extract values directly to their OCaml form). However, a question remains: which kind of type we want to handle. RFC 5322 defines an exhaustive list of what an email parser should be aware (`From`, `Date`, etc.) and `mrmime` implements only them. I think HTTP should list them too.

If you want to go to this way, it requires more (orthogonal) works:
1) implement such type like `Content_type.t`/`Accept_*.t`/etc.
2) tweak the underlying list of `Header` to be an heterogeneous list of GADT constructors
3) handle non-specified values (such as `Unstructured` for `mrmime`)

It requires more work about the parser of fields and may be it's the good time to:
- use Angstrom (as HTTP/AF) and `unstrctrd` (to be sure to handle any cases)
- or directly use `Unstrctrd.of_string`

And post-process values with associated parser (the `Content-Type` parser, the `Accept-*` parser, etc.).

cohttp/fuzz/fuzz_header.ml
```diff
@@ -0,0 +1,563 @@
...
+    (*FS*)
+    add_test ~name:"[update_all h k (fun _ -> [])] removes all occurences of k."
+      [ headers_gen; header_name_gen ] (fun h k ->
+        let h1 = H.update_all h k (fun _ -> []) in
+        check_eq H.(get_multi h1 k) []);
+    (*FS*)
+    add_test
+      ~name:
+        "[update_all h k (function _ -> [w])] removes all occurences of k and \
+         add one." [ headers_gen; header_name_gen; word_gen ] (fun h k w ->
```
mseri, line 448:
```suggestion
         adds [w]." [ headers_gen; header_name_gen; word_gen ] (fun h k w ->
```

cohttp/fuzz/fuzz_header.ml
```diff
@@ -0,0 +1,563 @@
...
+  add_multi ();
+  get_test ();
+  get_multi_test ();
+  get_multi_concat_test ();
+  remove_test ();
+  replace_test ();
+  update_test ();
+  update_all_test ();
+  clean_dup_test ();
+  ()
```
mseri, line 563: This module is fantastic. I am going to point here people that want to get an idea on how to use crowbar!

cohttp/test/unitary_test_header.ml
```diff
@@ -0,0 +1,372 @@
...
+  aessl "map (fun _ v -> v ^ a ) (init ())"
+    [
+      ("accept", "application/xml, a");
+      ("transfer-encoding", "chunked, a");
+      ("accept", "text/html, a");
+      ("content-length", "100, a");
+    ]
+    H.(to_list (map (fun _k v -> v ^ a) prebuilt))
+
+let fold_tests () = ()
```
mseri, line 269: Should we create an issue tracking the implementation of these two tests? Do you have in mind something specific already?

cohttp/fuzz/fuzz_header.ml
lyrm: Thanks!

cohttp/test/unitary_test_header.ml
lyrm: I just added these missing tests.

mseri:
Manually merged on `feature/headersV2` branch to resolve a merge conflict. Thank you very much, we have a very solid and tested base to start further experimentation

PR #749, opened 2021-02-10
Add dune-lint job to GitHub Actions workflow

smorimoto:
As the title says. And now pin and depext are now performed automatically.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,27 +28,6 @@ jobs:
         uses: actions-ml/setup-ocaml@master
         with:
           ocaml-version: ${{ matrix.ocaml-version }}
-          opam-depext: false
-          opam-pin: false
-
-      - run: |
-          opam pin add cohttp-async.dev . --no-action
-          opam pin add cohttp-lwt-jsoo.dev . --no-action
-          opam pin add cohttp-lwt-unix.dev . --no-action
-          opam pin add cohttp-lwt.dev . --no-action
-          opam pin add cohttp-mirage.dev . --no-action
-          opam pin add cohttp-top.dev . --no-action
-          opam pin add cohttp.dev . --no-action
-
-      - run: |
-          opam depext --yes \
-          cohttp \
-          cohttp-async \
-          cohttp-lwt \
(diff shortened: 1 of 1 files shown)


mseri (approved):

smorimoto:
It's a little less obvious, but it's done in the post step.
image

mseri:
I think it's nice that it does not clutter the environment if it succeeds, and if it fails we will see that from the action panel

smorimoto:
I'm doing this because there is no smarter way than this at the moment, but I'm glad you felt like that.

samoht:
It's already done in https://ci.ocamllabs.io/github/mirage/ocaml-cohttp/commit/ea53c63339cedaf2bf77878f9eaa34034fc9032b/variant/(lint-opam), why do we need to duplicate it in Github Actions?

smorimoto:
There is no serious reason to duplicate this. For now, this is just some kind of test for the next major release of setup-ocaml, so we can choose not to add this. (I'm testing it on my fork.)

samoht:
Ha nice :-) btw, the Github Actions plugin is very nice, great job!

smorimoto:
Thank you! OCaml CI is really cool and fast enough, but I believe GitHub Actions is a good option for the regular user. Our goal is to allow as many users to use it and test it on many platforms without huge effort. The next major version also includes official container support, so we can test on more platforms such as Alpine and CentOS that are not provided natively by GitHub!

```yml
jobs:
  build:
    strategy:
      fail-fast: false
      matrix:
        container:
          - debian:latest
          - ubuntu:latest
        ocaml-version:
          - 4.11.x

    runs-on: ubuntu-latest

    container:
      image: ${{ matrix.container }}

    steps:
      - run: apt update && apt install --yes bubblewrap curl darcs gcc git m4 make mercurial patch rsync sudo unzip

      - name: Use OCaml ${{ matrix.ocaml-version }}
        uses: actions-ml/setup-ocaml@master
        with:
          ocaml-version: ${{ matrix.ocaml-version }}
          opam-disable-sandboxing: true

      - run: ...
```

smorimoto:
I'm rethinking this interface a bit, so I close it once.

PR #753, opened 2021-03-12
response: add extra space after http response status

### LINKED ISSUE #752: Invalid HTTP Status line when a custom status code is used
In https://github.com/janestreet/cohttp_async_websocket/pull/1 a user reported that using  `` `Code 101 `` as a HTTP response status makes cohttp generate a response header that can't be parsed by Safari.

The response status line generated is `HTTP/1.1 101`, which, I would argue, is not compliant with the [spec](https://tools.ietf.org/html/rfc7230#section-3.1.2) because it doesn't have the second space character.

The culprit seems to be this line:
https://github.com/mirage/ocaml-cohttp/blob/c3a59cd11fae2ccf084fbfc3eb02b75773511d25/cohttp/scripts/generate.ml#L241

Maybe it should be changed to generate `HTTP/1.1 101 `, instead (note the extra space)?
Or, less confusingly, `HTTP/1.1 101 Status 101`?

mseri:
This is required, according to the spec: https://tools.ietf.org/html/rfc7230#section-3.1.2
The rest of the code seems to (correctly) ignore this extra space so the change should not affect anything except maybe some too restrictive tests

Should fix #752

### DIFF
--- a/cohttp-lwt-unix/test/test_parser.ml
+++ b/cohttp-lwt-unix/test/test_parser.ml
@@ -19,7 +19,7 @@ open OUnit
 let basic_req = "GET /index.html HTTP/1.1\r\nHost: www.example.com\r\n\r\n"
 
 let basic_res =
-  "HTTP/1.1 200 OK\r\n\
+  "HTTP/1.1 200 OK \r\n\
    Date: Mon, 23 May 2005 22:38:34 GMT\r\n\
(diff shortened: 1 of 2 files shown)


mseri:
The description is optional so I left it out to minimize the changes. I could add it if we decide is worth. Any strong opinion?

mseri:
This is the wrong fix

PR #754, opened 2021-03-12
generate: add space to string_of_status

### LINKED ISSUE #752: Invalid HTTP Status line when a custom status code is used

mseri:
This is required by the spec.
The change could have gone into the response module,
adding a specific logic checking the presence of the
space before the string termination, but it seemed
cleaner to add it here.

Should fix #752. 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## current
 
+- cohttp.response: fix malformed status header for custom status codes (@mseri @aalekseyev #752)
 - Remove dependency to base (@samoht #745)
 - fix opam files and dependencies
 - add GitHub Actions workflow (@smorimoto #739)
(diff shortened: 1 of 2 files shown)


mseri:
We can produce a better response with https://github.com/mseri/ocaml-cohttp/blob/4ab0ba92747848c54e35c63135a63b4257caa70d/cohttp/src/response.ml#L107-L115 but it probably adds extra complexity for no reason.

samoht (approved):

avsm (approved):
The simpler one looks good to me -- no need to put extra bytes over the wire when the longer description can be looked up easily enough :-)

Just needs a CHANGES entry and good to merge.

PR #756, opened 2021-03-15
Merge resolve_file / resolve_local_file implementations (fixes #755)

### LINKED ISSUE #755: resolve_file does not decode percent encoding
`Cohttp_lwt_unix.Server.resolve_file` does not decode the percent encoding in the given URI's path, meaning that it returns an incorrect filename for anything with an encoded character. In the code below, `Uri.path` (from mirage/ocaml-uri) returns the encoded form of the path.

```
# cohttp-lwt-unix/src/server.ml:

let resolve_file ~docroot ~uri =
  (* This normalises the Uri and strips out .. characters *)
  let frag = Uri.path (Uri.resolve "" blank_uri uri) in
  Filename.concat docroot frag
```

The fix is easy, but investigating this I found that there are two further implementations of essentially the same function:

```
# ./cohttp-lwt/src/server.ml:

 let resolve_local_file ~docroot ~uri =
    let path = Uri.(pct_decode (path (resolve "http" (of_string "/") uri))) in
    let rel_path = String.sub path 1 (String.length path - 1) in
    Filename.concat docroot rel_path

# ./cohttp-async/src/server.ml:

let resolve_local_file ~docroot ~uri =
  (* This normalises the Uri and strips out .. characters *)
  Uri.(pct_decode (path (resolve "" (of_string "/") uri)))
  |> Caml.Filename.concat docroot
```

(Note that these two implementations at least get the percent-decode correct!)

Why do we have three different implementations of the same function? Is there somewhere sensible that we can put a canonical version?

ewanmellor:
Merge Cohttp_async.resolve_local_file, Cohttp_lwt.resolve_local_file,
and Cohttp_lwt_unix.resolve_file into one new function,
Cohttp.Server_utils.resolve_local_file.  (This adds Server_utils as a
new module, since there didn't seem to be any other good place for this
code).

The implementation now correctly handles ../ in the URI
(Cohttp_lwt_unix.resolve_file claimed to do so, but actually didn't
for file URIs). It percent-decodes the URI path
(Cohttp_lwt_unix.resolve_file did not, but the other two already did).
It correctly handles the path separators so that we don't get two slashes
together (both Cohttp_lwt_unix.resolve_file and
Cohttp_async.resolve_local_file got that one wrong).  It also doesn't
crash if the URI is blank (Cohttp_lwt.resolve_local_file got that wrong).

Signed-off-by: Ewan Mellor <ewan@tarides.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -76,10 +76,10 @@
 
 - `Cohttp_async.resolve_local_file`, `Cohttp_lwt.resolve_local_file` and `Cohttp_lwt_unix.resolve_file`
   are now the same code under the hood (`Cohttp.Path.resolve_local_file`). The old names
-  have been preserved for compatibility, but marked as deprecated. This changes the behavior of
-  `Cohttp_lwt_unix.resolve_file`: it now percent-decodes the paths and blocks escaping from the
-  docroot correctly. This also fixes and tests the corner cases in these methods when the docroot
-  is empty. (#755)
+  have been preserved for compatibility, but will be marked as deprecated in the next release. This
+  changes the behavior of `Cohttp_lwt_unix.resolve_file`: it now percent-decodes the paths and blocks
+  escaping from the docroot correctly. This also fixes and tests the corner cases in these methods
+  when the docroot is empty. (@ewanmellor #755)
 
 ## v2.5.4 (2020-07-21)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -72,6 +72,15 @@
 - improve media type parsing (@seliopou #542, @dinosaure #725)
 - add comparison functions for Request.t and Response.t via ppx_compare (@msaffer-js @dinosaure #686)
 
+## v2.5.5 (2021-03-15)
+
+- `Cohttp_async.resolve_local_file`, `Cohttp_lwt.resolve_local_file` and `Cohttp_lwt_unix.resolve_file`
+  are now the same code under the hood (`Cohttp.Path.resolve_local_file`). The old names
+  have been preserved for compatibility, but marked as deprecated. This changes the behavior of
+  `Cohttp_lwt_unix.resolve_file`: it now percent-decodes the paths and blocks escaping from the
+  docroot correctly. This also fixes and tests the corner cases in these methods when the docroot
+  is empty. (#755)
+
 ## v2.5.4 (2020-07-21)
 
 - cohttp: a change in #694 modified the semantics of Header.replace.
--- a/cohttp-lwt-unix/src/server.mli
+++ b/cohttp-lwt-unix/src/server.mli
@@ -4,6 +4,7 @@
 include Cohttp_lwt.S.Server with module IO = Io
 
 val resolve_file : docroot:string -> uri:Uri.t -> string
+(** Deprecated. Please use Cohttp.Path.resolve_local_file. *)
 
 val respond_file :
   ?headers:Cohttp.Header.t ->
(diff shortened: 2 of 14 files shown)


cohttp/src/server_utils.ml
```diff
@@ -0,0 +1,4 @@
+let resolve_local_file ~docroot ~uri =
```
avsm, line 1: Could you rename Cohttp.Server_utils to Cohttp.Path, perhaps? It's more descriptive.

cohttp-lwt-unix/src/server.mli
```diff
@@ -4,6 +4,7 @@
 include Cohttp_lwt.S.Server with module IO = Io
 
 val resolve_file : docroot:string -> uri:Uri.t -> string
+[@@@ocaml.alert deprecated "Please use Cohttp.Server_utils.resolve_local_file."]
```
avsm, line 7: Let's hold off on the deprecation warnings due to the implications of this fix.  We want people to upgrade quickly -- the deprecations can be added in a future release.

avsm (changes requested):
Looks good to me -- just a small change about the name of the Cohttp module. Thanks for all the test cases!

CHANGES.md
```diff
@@ -72,6 +72,15 @@
 - improve media type parsing (@seliopou #542, @dinosaure #725)
 - add comparison functions for Request.t and Response.t via ppx_compare (@msaffer-js @dinosaure #686)
 
+## v2.5.5 (2021-03-15)
+
+- `Cohttp_async.resolve_local_file`, `Cohttp_lwt.resolve_local_file` and `Cohttp_lwt_unix.resolve_file`
+  are now the same code under the hood (`Cohttp.Server_utils.resolve_local_file`). The old names
```
avsm, line 78:
```suggestion
  are now the same code under the hood (`Cohttp.Path.resolve_local_file`). The old names
```

ewanmellor: Done.

cohttp-lwt-unix/src/server.mli
ewanmellor: Done.

cohttp/src/server_utils.ml
ewanmellor: Done.

samoht (approved):

avsm (approved):

cohttp-lwt/src/s.ml
```diff
@@ -166,7 +166,9 @@ module type Server = sig
     t
 
   val resolve_local_file : docroot:string -> uri:Uri.t -> string
-  (** Resolve a URI and a docroot into a concrete local filename. *)
```
mseri, line 169: Should we add the deprecation annotation to these?

mseri (approved):

CHANGES.md
```diff
@@ -72,6 +72,15 @@
 - improve media type parsing (@seliopou #542, @dinosaure #725)
 - add comparison functions for Request.t and Response.t via ppx_compare (@msaffer-js @dinosaure #686)
 
+## v2.5.5 (2021-03-15)
+
+- `Cohttp_async.resolve_local_file`, `Cohttp_lwt.resolve_local_file` and `Cohttp_lwt_unix.resolve_file`
+  are now the same code under the hood (`Cohttp.Path.resolve_local_file`). The old names
+  have been preserved for compatibility, but marked as deprecated. This changes the behavior of
```
mseri, line 79: Instead of marked as deprecated I will modify this as “will be marked as deprecated in the next released” if that is fine

PR #760, opened 2021-03-27
Improve the opam files

kit-ty-kate:
dune subst should be used only in the dev case, reformat and synchronise the dune constraint accross packages


mseri:
Oh I should have been more careful with those, thanks a lot!

PR #761, opened 2021-03-31
Fix exception on connection errors in chrome.

mefyl:
On chrome/nodejs (ie. v8), the response field of XHR requests is `null` when the requests fails, as opposed to the empty string on firefox. In that case the coercion would fail, `respText ()` is called assuming it's a string and disaster ensues.


samoht:
Is there a way to add a test for this?

mefyl:
I don't think there are tests for cohttp-http-jsoo yet ? It's doable, but we'd need to depend on nodejs to run the tests, and have a mock server to answer requests - I've done this for internal projects and it works like a charm. I can look into it but can't guarantee the timeline, I have a lot on my plate internally right now =)

mefyl:
Good news #1 : I took the time to write some cohttp-lwt-jsoo tests.
Good news #2 : I could not reproduce and I think this was actually already fixed in e18e6a8059cc324f62da7961e0af4d84013c8258 . I must have been stuck pre-2.5 before the release of 4.0.0 and didn't recheck after rebasing. Sorry for the noise, good catch @samoht .

I will open up a separate PR to introduce the cohttp-lwt-jsoo test.

mefyl:
Actually, this was only partially fixed by e18e6a8059cc324f62da7961e0af4d84013c8258, as described in #768 .

mefyl:
The failure is due to an action setup error, and I can't seem to find the "try again" button 🤔

cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
```diff
@@ -202,15 +202,18 @@ module Make_client_async (P : Params) = Make_api (struct
...
-                      Js.Opt.case
-                        (File.CoerceTo.arrayBuffer xml##.response)
-                        (fun () ->
-                          Firebug.console##log
-                            (Js.string
-                               "XHR Response is not an arrayBuffer; using \
-                                responseText");
-                          respText ())
-                        (fun ab -> `ArrayBuffer ab)
+                      if Js.Opt.return xml##.response == Js.null then
```
mseri, line 205: - Does this need to be done also at: https://github.com/mirage/ocaml-cohttp/pull/761/files#diff-17a4a55959fc8c74371af346bb1da1f88718607b785b1f1ca651b3b226d3f3cfR311
- If so, can we extract this function instead of having so much duplication?
- Is there a way to test this?

mefyl: Sure, IIRC there's even a comment by @avsm there that it is duplicated.

mefyl: I gave it a try, the only difference is the async version wrapping the response in a `Lwt.t`, so we'd need to parameterize the whole thing over the "future" monad. But really the whole `Make_client_sync` and `Make_client_async` are 99% duplicated, a simple abstraction over that monad would factor it (someone has been veeeery lazy or in a hurry when writing this 😅).

I don't mind taking a look, but honestly that should be a different PR.

mefyl: Regarding the reproduction, I know of no easy way. AFAICT it's the internal XHR implementation of chrome, and it's not clear when to me why `responseType` is `"arraybuffer"`, but `response` is `null`. It looks like a mild bug [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType), but since it's meant to be used from javascript and everything fits everywhere™ in that magical world, it's probably not much of an issue - and by that I mean not a bigger issue that the rest of the language and libraries.

I can't reproduce it myself, I just get that exception reported a few times a day from our app running in the wild, all I can say is that it's possible to reach this state with a fairly recent chrome (electron 12). I thought I had a reproduction in the js test suite, which prompted me to drop this PR, but I was empirically wrong it seems.

mseri: Fair enough. Given how complicated is managing exceptions across ocaml and js, I think it makes sense to be a bit proactive here. We should probably add a LOG message that the call returned a `null` response though, so it does not come unnoticed if it happens by mistake

mseri: > I don't mind taking a look, but honestly that should be a different PR.

That's good for me. Thanks for putting in the effort!

mefyl: Done

cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
```diff
@@ -94,6 +98,32 @@ module Body_builder (P : Params) = struct
...
+
+  let construct_body xml =
+    (* construct body *)
+    let b =
+      let respText () =
+        Js.Opt.case xml##.responseText
+          (fun () -> `String (Js.string ""))
+          (fun s -> `String s)
+      in
+      if xhr_response_supported then
```
mseri, line 110:
```ocaml
match  xhr_response_supported with
| true when Js.Opt.return xml##.response = Js.null -> 
  Firebug.console##log (Js.string "XHR Response is null; using empty string");
  `String (Js.string "")
| true ->
  Js.Opt.case
     (File.CoerceTo.arrayBuffer xml##.response)
     (fun () -> Firebug.console##log
                (Js.string
                   "XHR Response is not an arrayBuffer; using responseText");
              respText ())
            (fun ab -> `ArrayBuffer ab)
| -> respText ()
```
This is a readability suggestion, do you think it would be better or worse than the nested ifs?

mseri:
I think this is good to go in, I can make extra changes -- if needed -- separately

PR #762, opened 2021-03-31
Fix Invalid_arg exception on stack overflow when waking up a request.

mefyl:
This one is a bit of a brain twister : if you trigger a javascript stack overflow in `Lwt.wakeup` - which I do on a daily basis at this precise point in the code - the promise will still be marked as completed. Trying to wake it again with the Stack_overflow exception will raise Invalid_arg as you may not wake a promise twice, ending up with a very strange exception in the javascript toplevel. Letting the exception through if the promise was already woken seems like a better option.

Side note: this is definitely not entirely satisfactory since the promise is in a completely broken state : it is considered woken, but not all callbacks have been entirely called. That would require fixing in Lwt though, which knows no exception can escape from callbacks and thus is not exception safe in this regard. Running OCaml code through Jsoo makes ANY call susceptible to throwing an exception, but we're probably not going to alter all existing OCaml code for this possibility. I would argue that getting a `Stack_overflow` exception implies undefined behavior from the program and everything needs to be started over.

### DIFF
--- a/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
+++ b/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
@@ -231,7 +231,20 @@ module Make_client_async (P : Params) = Make_api (struct
                  * Remove the type constraint on Lwt.task above and return any old
                  * guff here.  It'll compile and crash in the browser! *)
                 Lwt.wakeup wake (response, body)
-              with e -> Lwt.wakeup_exn wake e)
+              with
+              | e
+              (* If we exhaust the stack, it is possible that
+                 Lwt.wakeup just aboves marks the promise as
+                 completed, but raises Stack_overflow while
+                 running the promise callbacks. In this case
+                 waking calling wakeup_exn on the already
+                 completed promise would raise an Invalid_arg
+                 exception, so although the promise is in a
+                 really bad state we may as well let the actual
+                 Stack_overflow exception go through. *)
+              when Lwt.state res = Lwt.Sleep
+              ->
+                Lwt.wakeup_exn wake e)
           | _ -> ());
 
     (* perform call *)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## current
+
+- lwt_jsoo: Fix `Lwt.wakeup_exn` `Invalid_arg` exception when a js
+  stack overflow happens in the XHR completion handler (@mefyl #762).
+
 ## v4.0.0 (2021-03-24)
 
 - cohttp.response: fix malformed status header for custom status codes (@mseri @aalekseyev #752)


mseri:
Is this really an lwt problem or could be tackled at the level of js_of_ocaml_lwt?

cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
```diff
@@ -231,7 +231,20 @@ module Make_client_async (P : Params) = Make_api (struct
...
+              (* If we exhaust the stack, it is possible that
+                 Lwt.wakeup just aboves marks the promise as
+                 completed, but raises Stack_overflow while
+                 running the promise callbacks. In this case
+                 waking calling wakeup_exn on the already
+                 completed promise would raise an Invalid_arg
+                 exception, so although the promise is in a
+                 really bad state we may as well let the actual
+                 Stack_overflow exception go through. *)
+              when Lwt.state res = Lwt.Sleep
```
mseri, line 245: What anout other genuine exceptions that we would have raised otherwise?

mefyl: The exception doesn't really matter : if we end up in this state and `Lwt.state res` is not `Sleep`, calling `Lwt.wakeup_exn state _` will raise `Invalid_arg` since you can't wake a promise twice, so we're a bit stuck between a rock and a hard place. In the end calling `wakeup_exn` will not wake anything and makes an `Invalid_arg` exception escape to the javascript toplevel, hiding the actual exception. This patch makes the actual exception, `Stack_overflow`, reach the toplevel instead, which I argue is always better, although not perfect.

In practice: I often get a Stack_overflow error in javascript, but I also had this `Lwt.wakeup_exn: Invalid_arg` exception sometime in production logs. After digging I saw this was actually a hidden stack overflow too, with this patch it is successfully reported as so without altering anything else.

mefyl:
> Is this really an lwt problem or could be tackled at the level of js_of_ocaml_lwt?

I don't think it's possible (but happy to be proven wrong). The stack overflow is raised from `Lwt.wakeup` and the `Lwt.t` is marked as resolved, so I'm pretty sure all Lwt semantics is broken - the promise is considered done, but we were probably stopped mid-callbacks and dependent promise did not wake. This is (in my interpretation) because Lwt.wakeup is not exception safe, but can hardly be blamed since its content *cannot* throw in plain OCaml. It's the switch to Jsoo that makes any piece of code susceptible of throwing an exception.

The way I see it we can phrase this three ways:
* Jsoo breaks OCaml semantics and produces invalid code since exception arise where they previously couldn't with the OCaml spec: yikes.
* Any library that pretends to be Jsoo compatible must be 100% exception safe at any single point of its code. While this would be the cleanest solution, I don't think it's realistic to expect this.
* When you get a Javascript `Stack_overflow` exception, consider your program broken and all semantics lost. In other words Stack_overflow = undefined behavior. We get the exception cleanly to the toplevel as a courtesy so you understand you're exhausting the (toy-sized) javascript stack, but past that point everything goes.

It seems to me solution 3 is the more realistic and quite satisfying.

In any case this discussion goes almost past that MR the way I see things. This patch makes - if my not mistaken - zero behavioral change to the program, it just make the right exception reach the toplevel instead of a cryptic `Invalid_arg`.

cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
mseri: I understand now, thanks. Then, indeed, solution seems reasonable and also my other comment makes less sense :P

mseri:
I agree with you, thanks for the detailed explanation. Please add a changelog entry for this change, referring to the PR and crediting yourself

PR #764, opened 2021-04-01
Add cohttp-lwt-jsoo test suite.

mefyl:


cohttp-lwt-jsoo.opam
```diff
@@ -27,6 +27,7 @@ depends: [
   "cohttp" {= version}
   "cohttp-lwt" {= version}
   "lwt" {>= "3.0.0"}
+  "lwt_ppx" {with-test}
   "js_of_ocaml" {>= "3.3.0"}
```
mseri, line 31:
```suggestion
  "conf-npm" {with-test}
  "js_of_ocaml" {>= "3.3.0"}
```
Libraries using `node` for tests generally use `conf-npm` to ensure node is installed

mseri: Examples are https://github.com/mnxn/jsonoo and https://github.com/mnxn/promise_jsoo

mefyl: Well I learned something, thanks :)

mseri:
```
(cd _build/default/cohttp_lwt_jsoo_test/src && ./test.js)
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module 'xmlhttprequest'
Require stack:
- /src/_build/default/cohttp_lwt_jsoo_test/src/test.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/src/_build/default/cohttp_lwt_jsoo_test/src/test.js:5:25)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/src/_build/default/cohttp_lwt_jsoo_test/src/test.js' ]
}
```

To make it work I think you need to do the install at the level of the dune runtest, maybe something like  https://github.com/jacquev6/JsOfOCairo/blob/a00a31000b1a7cabe6ac74e48e86b52ec4699cff/tst/dune#L31

mseri:
Alternatively we create a different alias, say `runjstests` and then call it explicitly only from the GitHub Actions, this way the opam-ci does not need to know about it

mefyl:
Note: I tried using `promise_jsoo` instead of my manual binding, but it conflicts with the `ocaml-migrate-parsetree` needed for ocamformat 0.16.

mseri:
> Note: I tried using `promise_jsoo` instead of my manual binding, but it conflicts with the `ocaml-migrate-parsetree` needed for ocamformat 0.16.

I was just suggesting to look at how they run the tests since they also use node, not implying to use them, sorry if that seemed the case.

mefyl:
No you were perfectly clear, I just thought "Woo proper promise bindings let's use that", but that bit me back :)

cohttp_lwt_jsoo_test/src/test.js
```diff
@@ -0,0 +1,19 @@
...
+const tests = require('./cohttp_lwt_jsoo_test.bc.js')
+
+async function main() {
+  {
+    const [status, body] = await tests.request("https://mirage.io");
+    assert(status == 200);
+  }
+  {
+    const [status, body] = await tests.request("https://this.domain.does.not.exist");
+    assert(status == 0);
```
mefyl, line 15: This test checks what e18e6a8059cc324f62da7961e0af4d84013c8258 fixed.

cohttp_lwt_jsoo_test/src/test.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
+
+const assert = require('assert');
+
+global.XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
+const tests = require('./cohttp_lwt_jsoo_test.bc.js')
+
+async function main() {
+  {
+    const [status, body] = await tests.request("https://mirage.io");
```
mefyl, line 10: We may want to spin our own local http server to avoid depending on external connectivity.

mseri: We could recycle https://github.com/mirage/ocaml-cohttp/blob/master/cohttp_lwt_unix_test/src/cohttp_lwt_unix_test.ml#L27 or some other example from the code for it, like https://github.com/mirage/ocaml-cohttp/blob/master/examples/lwt_unix_doc/server_lwt.ml

mefyl:
This looks like a decent enough first step, I'll let you merge it or improve on it if that's okay.

mseri:
Is it possible to move package.json and yarn.lock in the cohttp_jsoo_test folder or generate them on the fly via dune?

mseri:
One last thing, can you add a changelog entry?

mseri:
Thanks!

PR #769, opened 2021-04-14
Minor cleanup

mseri:
CHANGES: add lates merged PR and removes all the @ to avoid unnecessarily spamming all the contributors
lwt_jsoo: `==` -> `=` (noticed after merging) and used match instead of nested if...then...else

### DIFF
--- a/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
+++ b/cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
@@ -107,12 +107,12 @@ module Body_builder (P : Params) = struct
           (fun () -> `String (Js.string ""))
           (fun s -> `String s)
       in
-      if xhr_response_supported then
-        if Js.Opt.return xml##.response == Js.null then (
+      match xhr_response_supported with
+      | true when Js.Opt.return xml##.response = Js.null ->
           Firebug.console##log
             (Js.string "XHR Response is null; using empty string");
-          `String (Js.string ""))
-        else
+          `String (Js.string "")
+      | true ->
           Js.Opt.case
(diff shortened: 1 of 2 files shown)


cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml
```diff
@@ -107,12 +107,12 @@ module Body_builder (P : Params) = struct
           (fun () -> `String (Js.string ""))
           (fun s -> `String s)
       in
-      if xhr_response_supported then
-        if Js.Opt.return xml##.response == Js.null then (
```
mseri, line 111: The only worth change is `==` to `=` here. The rest is just because I find a match easier to read than nested if then elses and I am happy to drop it if deemed unnecessary.

mseri: @mefyl was there a specific reason you used the physical equality `==` instead of `=`?

mefyl: Yes, I wanted `response === null` in javascript. It checks the value is actually the null singleton, the other alternative would also catch `undefined` (and would thus be marginally slower). It's a common stylistic precept in js to use `===` in most places, even strings which are flyweight instances. So the alternative will probably work the same,  but I think it's marginally better stylistically (juste like you should probabl `==` between two `Js.js_string Js.t`).

mseri: Ok thanks for the explanation (which I stole for the commit message 😄 ) Then I revert it!

PR #771, opened 2021-04-14
lwt.client: log drain body warning instead of printing it to stderr

mseri:
I think it is worth having the improved warning message and the extra documentation.

But I am not sure logging is better: currently, you are immediately bitten by the error messages appearing everywhere in the tests, by hiding it in the logging you need to consciously look for the problem.

### DIFF
--- a/cohttp-lwt/src/client.ml
+++ b/cohttp-lwt/src/client.ml
@@ -6,6 +6,10 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
   module Response = Make.Response (IO)
   module Request = Make.Request (IO)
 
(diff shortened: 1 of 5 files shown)


cohttp-lwt/src/client.ml
```diff
@@ -25,7 +29,11 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
         Gc.finalise_last
           (fun () ->
             if not !closed then
-              prerr_endline "Cohttp_lwt: body not consumed - leaking stream!")
```
mseri, line 28: @dinosaure what is the effect of `prerr_endline` on `mirage`? Is that a problematic call or it is fine?

dinosaure: Hmmhmm nothing as far as know. We should definitely remove it or print it with `Logs`.

Issue #770, opened 2021-04-14
Weird behaviour of cohttp-lwt-unix logging


mseri:
This can be seen from the server:
```
$ cohttp-server-lwt
[DEBUG][application]: Cohttp debugging output is active

$ cohttp-server-lwt --verbose
[DEBUG][application]: Cohttp debugging output is active
[INFO][cohttp.lwt.server]: Listening for HTTP request on: :: 8080

```

Without the verbose flag I would expect just the info to appear, instead the debug message appears always while the info message only when we use the verbose flag.

I also think that we should make the verbose output more verbose, more like python's `http.server`:
```
$ python -m http.server
Serving HTTP on :: port 8000 (http://[::]:8000/) ...
::ffff:127.0.0.1 - - [14/Apr/2021 21:24:04] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [14/Apr/2021 21:24:04] code 404, message File not found
::ffff:127.0.0.1 - - [14/Apr/2021 21:24:04] "GET /favicon.ico HTTP/1.1" 404 -
::ffff:127.0.0.1 - - [14/Apr/2021 21:24:06] "GET /bin/ HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [14/Apr/2021 21:24:07] "GET /bin/lib.ml HTTP/1.1" 200 -
```

mseri:
We should probably deal with it when we deal with https://github.com/mirage/ocaml-cohttp/issues/645

mseri:
Closing since most of it has been improved by #772.
We can open a new issue if other people are unhappy with the current quiet logging

PR #772, opened 2021-04-15
Revisit logs in cohttp-lwt

mseri:
This should partially address #645 and #770 by, at least, documenting the logs sources names and how to silence them. And by fixing a few issues in the debug module.

I think the use of COHTTP_DEBUG and the `enable_debug` function, as convenient as they may be, should be deprecated. Instead we should only provide a reference reporter (maybe the one currently shadowed in the Debug module)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## current
 
+- lwt, lwt_unix: Improve use of logs and the documentation, fix bug in the Debug.enable_debug function (mseri #772)
 - lwt_jsoo: Fix exception on connection errors in chrome (mefyl #761)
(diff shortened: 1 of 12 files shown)


mseri:
This goes hands in hands with https://github.com/mirage/ocaml-cohttp/pull/771

mefyl:
The `Logs` module works with JSOO. What would you think of using it so the `response` warnings can be silenced too ? I'm always bugged by a library that prints unconditionally.

mseri:
@mefyl I like the idea

dinosaure (approved):
It seems very good, I was lazy to upgrade such part of `cohttp` before. So I'm happy to see such upgrade :+1:.

mseri:
Thanks. I will merge and send a PR for jsoo separately

PR #774, opened 2021-04-15
Upgrade CoHTTP to use conduit.4.0.0

dinosaure:
This PR is related to several releases:
- `tls.0.13.0`
- `dns.5.0.0`
- `conduit.4.0.0`

It's completely transparent for CoHTTP on implementations but it requires a fix on dependencies.

### DIFF
--- a/cohttp-async.opam
+++ b/cohttp-async.opam
@@ -31,7 +31,7 @@ depends: [
   "base" {>= "v0.11.0"}
   "core" {with-test}
   "cohttp" {= version}
-  "conduit-async" {>= "1.2.0" & < "3.0.0"}
+  "conduit-async" {>= "4.0.0"}
   "magic-mime"
   "mirage-crypto" {with-test}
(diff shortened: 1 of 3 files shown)


mseri:
Looks like no change is needed, am I wrong? Once conduit 3.0.0 is marked unavailable we can simply remove the upper bound (or add conflicts with ">=3.0.0" & "<4.0.0").

dinosaure:
Looks like no change is needed, am I wrong? Once conduit 3.0.0 is marked unavailable we can simply remove the upper bound (or add conflicts with ">=3.0.0" & "<4.0.0").

Yes, so we don't need to release `cohttp` :+1:. I need to test some _unikernels_ with such path of dependencies and then, I will propose a release of `conduit` with a large lint on dependencies of packages.

mseri:
I am going to close this now that conduit is merged and remove upper bounds from the opam files

PR #776, opened 2021-04-17
lwt_jsoo: use log instead of always printing the warnings on the console

mseri:


mseri:
Ping @mefyl

PR #778, opened 2021-04-23
cohttp.headers: use faster comparison

mseri:
I have benchmarked the latency of cohttp using https://github.com/ocaml-multicore/retro-httpaf-bench/tree/master/cohttp-lwt-unix and, with the new headers module, the measures are ~33% slower on my machine.

With this change, we seem to get ~10% slowdown compared to the old metrics on my laptop. Still suboptimal but it is an improvement.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Use OCaml ${{ matrix.ocaml-version }}
-        uses: actions-ml/setup-ocaml@master
+        uses: ocaml/setup-ocaml@v2
         with:
           ocaml-version: ${{ matrix.ocaml-version }}
           opam-depext: false
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
   stack overflow happens in the XHR completion handler (mefyl #762).
 - lwt_jsoo: Add test suite (mefyl #764).
 
-- Cohttp.Header: new implementation (@lyrm #747)
+- Cohttp.Header: new implementation (lyrm #747)
 
   + New implementation of Header modules using an associative list instead of a map, with one major semantic change (function ```get```, see below), and some new functions (```clean_dup```, ```get_multi_concat```)
   + More Alcotest tests as well as fuzzing tests for this particular module.
@@ -38,6 +38,7 @@
 
   + ```clean_dup```  enables the user to clean headers that follows the {{:https://tools.ietf.org/html/rfc7230#section-3.2.2} RFC7230§3.2.2} (no duplicate, except ```set-cookie```)
   + ```get_multi_concat``` has been added to get a result similar to the previous ```get``` function.
+- Cohttp.Header: optimize internal of cohttp.headers (mseri #778)
 
 ## v4.0.0 (2021-03-24)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,9 @@
   + New implementation of Header modules using an associative list instead of a map, with one major semantic change (function ```get```, see below), and some new functions (```clean_dup```, ```get_multi_concat```)
   + More Alcotest tests as well as fuzzing tests for this particular module.
 
+- Cohttp.Header: performance improvement (mseri, anuragsoni #778)
+  **Breaking** the headers are no-longer lowercased when parsed, the headers key comparison is case insensitive instead.
+
   ### Purpose
 
   The new header implementation uses an associative list instead of a map to represent headers and is focused on predictability and intuitivity: except for some specific and documented functions, the headers are always kept in transmission order, which makes debugging easier and is also important for [RFC7230§3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2) that states that multiple values of a header must be kept in order.
--- a/cohttp-lwt-unix/test/test_parser.ml
+++ b/cohttp-lwt-unix/test/test_parser.ml
@@ -246,7 +246,7 @@ let make_simple_req () =
   let open Cohttp in
   let open Cohttp_lwt_unix in
   let expected =
-    "POST /foo/bar HTTP/1.1\r\nfoo: bar\r\nhost: localhost\r\nuser-agent: "
+    "POST /foo/bar HTTP/1.1\r\nFoo: bar\r\nhost: localhost\r\nuser-agent: "
     ^ user_agent
     ^ "\r\ntransfer-encoding: chunked\r\n\r\n6\r\nfoobar\r\n0\r\n\r\n"
   in
--- a/cohttp/fuzz/fuzz_header.ml
+++ b/cohttp/fuzz/fuzz_header.ml
@@ -191,8 +191,7 @@ let init_with_test () =
     (* FS *)
     (* forall k v. to_list (init_with k v) = [k, v] *)
     add_test ~name:"[init_list k v] is [k, v]" [ header_name_gen; word_gen ]
-      (fun k v ->
-        check_eq H.(to_list (init_with k v)) [ (k, v) ]))
+      (fun k v -> check_eq H.(to_list (init_with k v)) [ (k, v) ]))
 
 let mem_test () =
   Crowbar.(
@@ -201,12 +200,16 @@ let mem_test () =
     add_test ~name:"[mem h k] on an empty header is always false"
       [ header_name_gen ] (fun k -> check_eq false H.(mem (init ()) k));
     (* SI *)
-    (* forall h, k. H.mem h k = List.(mem_assoc k (H.to_list h)) *)
+    (* forall h, k. H.mem h k = List.(mem_assoc (String.lowercase_ascii x) (List.map (fun (k, v) -> String.lowercase_ascii k, v) (H.to_list h))) *)
     add_test ~name:"Header.mem has the same behavior than List.mem_assoc"
       [ headers_gen; header_name_gen ] (fun h k ->
         check_eq
           H.(mem h k)
-          List.(mem_assoc k (H.to_list h))))
+          List.(
+            mem_assoc (String.lowercase_ascii k)
+              (List.map
+                 (fun (k, v) -> (String.lowercase_ascii k, v))
+                 (H.to_list h)))))
 
 let add_test () =
   Crowbar.(
@@ -220,9 +223,7 @@ let add_test () =
     (* forall h, k, v. to_list (add h k v) = to_list h @ [lowercase k, v] *)
       ~name:"[add] adds a value at the header end"
       [ headers_gen; header_name_gen; word_gen ] (fun h k v ->
-        check_eq
-          (H.to_list h @ [ (k, v) ])
-          H.(to_list (add h k v))))
+        check_eq (H.to_list h @ [ (k, v) ]) H.(to_list (add h k v))))
 
 let to_list_of_list_test () =
   Crowbar.(
--- a/cohttp/fuzz/fuzz_header.ml
+++ b/cohttp/fuzz/fuzz_header.ml
@@ -189,10 +189,10 @@ let is_empty_test () =
 let init_with_test () =
   Crowbar.(
     (* FS *)
-    (* forall k v. to_list (init_with k v) = [String.lowercase k, v] *)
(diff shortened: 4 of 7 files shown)


dinosaure:
It's weird that the standard implementation of `String.equal` was too slow comparing to what you did 😕 .

mseri:
> It's weird that the standard implementation of `String.equal` was too slow comparing to what you did 😕 .

Actually I did not check that 😂 let me try

mseri:
These are very unscientific and unrealistic benchmarks, since they run on my laptop and test a very dumb server, but here they give practically equivalent results 🤷 I added a commit to use `String.equal` since it is cleaner and hopefully faster on more realistic examples

anuragsoni:
> @mseri These are very unscientific and unrealistic benchmarks, since they run on my laptop and test a very dumb server, but here they give practically equivalent results 

I have recently switched to using Cohttp types for my http 1.0/1.1 parser and I ran a small benchmark on one of my parser tests. The change in your PR does seem to offer improvements in my little microbenchmark when compared to the previous version on the headersv2 branch. That said there could potentially be further improvements if breaking changes are acceptable, ex: using case insensitive comparison for header keys instead of modifying user input via `String.lowercase_ascii`. It should offer similar behavior but it'd be a breaking change compared to the current Cohttp header module since if anyone relies on keys being lowercase (not that they should since the spec mentions that header keys' case shouldn't change any behavior), that would no longer be the case. I'll attach the microbenchmark runs from my test:

The parser is defined at: https://github.com/anuragsoni/h1/blob/d659ad37f1222523725e8f37f6250de549f64112/h1_parser/src/h1_parser.ml and the benchmark code is https://github.com/anuragsoni/h1/blob/d659ad37f1222523725e8f37f6250de549f64112/bench/main.ml

The payload is:

```ocaml
let req = "GET /wp-content/uploads/2010/03/hello-kitty-darth-vader-pink.jpg HTTP/1.1\r\n\
   Host: www.kittyhell.com\r\n\
   User-Agent: Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10.6; ja-JP-mac; \
   rv:1.9.2.3) Gecko/20100401 Firefox/3.6.3 Pathtraq/0.9\r\n\
   Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8\r\n\
   Accept-Language: ja,en-us;q=0.7,en;q=0.3\r\n\
   Accept-Encoding: gzip,deflate\r\n\
   Accept-Charset: Shift_JIS,utf-8;q=0.7,*;q=0.7\r\n\
   Keep-Alive: 115\r\n\
   Connection: keep-alive\r\n\
   Cookie: wp_ozh_wsa_visits=2; wp_ozh_wsa_visit_lasttime=xxxxxxxxxx; \
   __utma=xxxxxxxxx.xxxxxxxxxx.xxxxxxxxxx.xxxxxxxxxx.xxxxxxxxxx.x; \
   __utmz=xxxxxxxxx.xxxxxxxxxx.x.x.utmccn=(referral)|utmcsr=reader.livedoor.com|utmcct=/reader/|utmcmd=referral\r\n\
   \r\n"
```
Results were as follows:

1. The version in the headers V2 branch

```
Estimated testing time 6s (3 benchmarks x 2s). Change using '-quota'.
┌───────────────────────┬────────────┬─────────┬──────────┬──────────┬────────────┐
│ Name                  │   Time/Run │ mWd/Run │ mjWd/Run │ Prom/Run │ Percentage │
├───────────────────────┼────────────┼─────────┼──────────┼──────────┼────────────┤
│ H1 (httparse example) │ 1_630.81ns │ 605.00w │    0.30w │    0.30w │    100.00% │
│ Parse chunk size      │    27.56ns │  17.00w │          │          │      1.69% │
│ Parse hex number      │    36.43ns │   6.00w │          │          │      2.23% │
└───────────────────────┴────────────┴─────────┴──────────┴──────────┴────────────┘
```

2. Version using String.equal

```
Estimated testing time 6s (3 benchmarks x 2s). Change using '-quota'.
┌───────────────────────┬────────────┬─────────┬──────────┬──────────┬────────────┐
│ Name                  │   Time/Run │ mWd/Run │ mjWd/Run │ Prom/Run │ Percentage │
├───────────────────────┼────────────┼─────────┼──────────┼──────────┼────────────┤
│ H1 (httparse example) │ 1_140.57ns │ 475.00w │    0.19w │    0.19w │    100.00% │
│ Parse chunk size      │    27.60ns │  17.00w │          │          │      2.42% │
│ Parse hex number      │    36.55ns │   6.00w │          │          │      3.20% │
└───────────────────────┴────────────┴─────────┴──────────┴──────────┴────────────┘
```
3. A version using a case insensitive comparison for header keys instead of making any modification to user input

```
Estimated testing time 6s (3 benchmarks x 2s). Change using '-quota'.
┌───────────────────────┬──────────┬─────────┬──────────┬──────────┬────────────┐
│ Name                  │ Time/Run │ mWd/Run │ mjWd/Run │ Prom/Run │ Percentage │
├───────────────────────┼──────────┼─────────┼──────────┼──────────┼────────────┤
│ H1 (httparse example) │ 968.67ns │ 414.00w │    0.15w │    0.15w │    100.00% │
│ Parse chunk size      │  27.61ns │  17.00w │          │          │      2.85% │
│ Parse hex number      │  36.52ns │   6.00w │          │          │      3.77% │
└───────────────────────┴──────────┴─────────┴──────────┴──────────┴────────────┘
```

The caseless comparison change can be seen at https://github.com/mirage/ocaml-cohttp/compare/feature/headersV2...anuragsoni:caseless-compare-header-key?expand=1

mseri:
Thanks a lot for these tests.

Since we are breaking the header module anyway, given the small difference, I think this is one of the instances in which we could go with it if we can justify it.

I don't personally tend to rely on the case of the keys, since I use the Header module functions for comparisons in general. I hope I am not too biased thinking that this is what most people do. And seeing how little needed to change in the tests, I think this breakage is subtle but should be minimal in practice (and we could still be very vocal and explicit about it).

I would like to hear some other bell on this. Is a 200 nanoseconds improvement over this PR worth the extra semantic change?

Since it seems that a lot of the latency cost is [hidden somewhere in the guts of the IO module and Lwt](https://github.com/mirage/ocaml-conduit/pull/387), are we overdoing for the header module?

mseri:
PS If I had to cast a vote, I would vote for using @anuragsoni case-insensitive patch.

anuragsoni:
> Since it seems that a lot of the latency cost is [hidden somewhere in the guts of the IO module and Lwt](https://github.com/mirage/ocaml-conduit/pull/387), are we overdoing for the header module?

There are definitely other improvements that can be made to cohttp that'll make more of an impact :) I think your `String.equal` change would be good to merge. As for the case-insensitive comparisons, i agree that even if its technically a breaking change i'm not sure if relying on the case of http keys was a safe idea at any point! That said, there are other more high impact changes that should help with the latency issues so i'd probably lean towards keeping whichever implementation feels better from a maintainability perspective :)

mseri:
I think your solution is also low maintenance, it is very close to my [first commit](https://github.com/mirage/ocaml-cohttp/pull/778/commits/8a9d2dab25305ddef7bf22ae1fd5fe3debe0beb5) fwiw :)

lyrm:
It seems like a good (and cheap !) improvement 👍.

mseri:
I have added your change here, and now I also remember why I had ditched the case-insensitive comparison when working on my initial PR. There are a number of property tests that need to be fixed :P

anuragsoni:
> There are a number of property tests that need to be fixed :P

Thanks for pulling in the commit here. I can help updating the necessary tests later in the week if that works :)

mseri:
Oh, thanks a lot. I fixed the immediate one. We only need to update the `Header.mem has the same behavior than List.mem_assoc` test, it would be great if you have time

anuragsoni:
> We only need to update the `Header.mem has the same behavior than List.mem_assoc` test, it would be great if you have time

I proposed a PR to this branch on your fork https://github.com/mseri/ocaml-cohttp/pull/2 
I believe these were the last 2 test failures.

mseri:
Thanks all for the help

gasche:
I came here from [the Discuss release announcement](https://discuss.ocaml.org/t/ann-cohttp-5-0-0-and-2-5-6/9109), always curious about some performance hacking. It looks like the fun is in the following `caseless_equal` function:

```ocaml
let caseless_equal a b =
  if a == b then true
  else
    let len = String.length a in
    len = String.length b
    &&
    let stop = ref false in
    let idx = ref 0 in
    while (not !stop) && !idx < len do
      let c1 = String.unsafe_get a !idx in
      let c2 = String.unsafe_get b !idx in
      if Char.lowercase_ascii c1 <> Char.lowercase_ascii c2 then stop := true;
      incr idx
    done;
    not !stop
```

I have a suggestion and an idle idea:
- Suggestion: setup a microbenchmark for this function so that you can easily evaluate the impact of performance changes -- it sounds like it is performance-critical for cohttp.
- Idle idea: I think you could speedup the function by using word-sized reads of both string: if the words at (word-aligned) index `i` are equal, you can skip the lowercase comparison of the 4/8 characters.

Word-sized reads are available in the standard library as `String.get_int{32,64}_ne`. (They use `int32`, `int64`, so one needs to check that the uses are correctly unboxed, but I expect they would.)

gasche:
(Technically it's possible to go even further as those functions perform bound checks and also assume that the memory-access is not word-aligned, so one could generate simpler unsafe code, but it's probably not worth it, and certainly not before trying with the exposed, safe `get_int<n>` functions.)

gasche:
Hacking on a quick benchmark is less demanding or less boring than most of my other day tasks, so I couldn't resist.

```ocaml
module Orig = struct
let caseless_equal a b =
  if a == b then true
  else
    let len = String.length a in
    len = String.length b
    &&
    let stop = ref false in
    let idx = ref 0 in
    while (not !stop) && !idx < len do
      let c1 = String.unsafe_get a !idx in
      let c2 = String.unsafe_get b !idx in
      if Char.lowercase_ascii c1 <> Char.lowercase_ascii c2 then stop := true;
      incr idx
    done;
    not !stop
end

module New = struct
let caseless_equal a b =
  if a == b then true
  else
    let len = String.length a in
    len = String.length b
    &&
    let rec loop a b len i = 
      i = len
      ||
      begin
        let i' = i + 8 in
        i' <= len
        &&
        String.get_int64_ne a i = String.get_int64_ne b i
        &&
        loop a b len i'
      end
      ||
      begin
        let c1 = String.unsafe_get a i in
        let c2 = String.unsafe_get b i in
        Char.lowercase_ascii c1 = Char.lowercase_ascii c2
        && loop a b len (i + 1)
      end
    in loop a b len 0
end

let impls = [
  "orig", Orig.caseless_equal;
  "new", New.caseless_equal;
]

let () =
  match List.assoc Sys.argv.(1) impls,
        int_of_string Sys.argv.(2),
        Sys.argv.(3),
        Sys.argv.(4)
  with
  | exception _ ->
    prerr_endline
      "Usage:\n./test [orig|new] <iter_count> '<input1>' '<input2>'";
    exit 2
  | caseless_equal, iter_count, input1, input2 ->

    (* some quick correctnss testing *)
    assert (caseless_equal "foo" "foo");
    assert (caseless_equal "transfer-encoding" "traNsfer-eNcoding");
    assert (caseless_equal "transfer-encoding" "transfer-Encoding");
    assert
      (* dash missing on the right-hand side*)
      (not (caseless_equal "transfer-encoding" "traNsfer eNcoding"));

    (* actual benchmark *)
    for i = 1 to iter_count do
      for i = 1 to 1_000 do
        ignore (caseless_equal input1 input2);
      done;
    done;

    Printf.printf "%b\n%!" (caseless_equal input1 input2)
```

```
$ time ./test orig 100_000 'transfer-encoding' 'transfer-encoding'
true

real	0m5.191s
$ time ./test new 100_000 'transfer-encoding' 'transfer-encoding'
true

real	0m2.043s
```

rgrinberg:
Would you like to throw up a PR gasche?

gasche:
@rgrinberg sure, done. But I don't know how to benchmark cohttp, so it would help if someone else could do that.

rgrinberg:
Okay, we'll take care of the benchmarking for you.

Rudi.

PR #784, opened 2021-10-19
[Preview] upgrade to ocamlformat.0.20.0

gpetiot:
This is a preview of the not-yet-released `ocamlformat.0.20.0`, please wait until the package is published in opam to merge this PR. The output is still likely to slightly change before the package is released.
The changes are due to:
- `module-item-spacing` is now set to `compact` for the default profile
- `module-item-spacing` is now correctly applied to mutually recursive type definitions
- the `.ocamlformat` file has been simplified using the conventional profile
- formatting of list elements has been improved

### DIFF
--- a/cohttp/src/link.ml
+++ b/cohttp/src/link.ml
@@ -382,7 +382,7 @@ let rec params_of_string s i ps =
           let charset, language, v, i = star_of_string s i in
           params_of_string s i
             (Star { Ext.charset; language; value = Link_extension (main, v) }
-             :: ps)
+            :: ps)
         else
           let v, i = quoted_string_of_string s i in
           params_of_string s i (Link_extension (other, v) :: ps)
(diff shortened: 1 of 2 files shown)


cohttp/src/link.ml
```diff
@@ -382,7 +382,7 @@ let rec params_of_string s i ps =
           let charset, language, v, i = star_of_string s i in
           params_of_string s i
             (Star { Ext.charset; language; value = Link_extension (main, v) }
-             :: ps)
+            :: ps)
```
mseri, line 385: Can we also leave this as it was?

gpetiot: This was the original formatting (pre-`ocamlformat.0.17.0`), that has been broken in 0.17.0 then fixed back in `ocamlformat.0.19.0`

mseri: Oh… Fine then, even though like the broken one better :P

mseri:
Then this is good to merge once the new ocamlformat is released :)

mseri:
ocamlformat is released, this should be merged

PR #786, opened 2021-10-21
Add more OCaml variants to GitHub Actions workflow

smorimoto:


mseri:
Can you rebase this on master? I'd like to check that it succeeds before merging it

mseri:
async does not run on win32 afaik. In this case we should avoid building in on the make win32 target, but maybe I am wrong

mseri:
Can we avoid testing async on windows (not compatible) and js_of_ocaml on multicore (not available) ?

smorimoto:
I found this not very useful at this point, so I close this once, but do this after switching to better packaging.

PR #791, opened 2021-12-09
Use package core_unix instead of core, to prepare for core release 0.15.

aalekseyev:
In the upcoming Jane Street release, many of the libraries move from package `core` to `core_unix` (package preview: https://github.com/janestreet/core_unix).

`core_unix.v0.14.0` is a compatibility package that is already on opam, so it should be possible to release this change before Jane Street packages v0.15 land.

In fact some of our packages depend on `cohttp-async`, so we'd appreciate it if a minor release that includes this change was made before we can release v0.15.

Signed-off-by: Arseniy Alekseyev <aalekseyev@janestreet.com>

### DIFF
--- a/cohttp-async.opam
+++ b/cohttp-async.opam
@@ -30,6 +30,7 @@ depends: [
   "async" {>= "v0.14.0"}
   "base" {>= "v0.11.0"}
   "core" {with-test}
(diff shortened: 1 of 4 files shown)


mseri:
Thanks! I am also going to backport this to the 2.x branch

PR #794, opened 2021-12-29
refactor: remove jsonm as build dependency

rgrinberg:
we now generate the codes and promote them.
to regenerate, run $ dune build @codes


mseri:
The script still requires jsonm and makes the build fail

rgrinberg:
I'll add jsonm as a test dependency for CI

rgrinberg:
The only thing that fails is the fmt check. I'll try to think of a workaround.

mseri:
We could make the script become an actual test, which fails if the generated file differs from the one that is already there.
Then both lints would succeed, and we have an extra check that we are shipping the right codes

rgrinberg:
Sure, I can do that but it's not the problem. The problem is that ocaml-ci doesn't install the test dependencies to run `@fmt`. @kit-ty-kate any way to configure this?

kit-ty-kate:
I’m not an ocaml-ci maintainer. cc @talex5 @tmcgilchrist

mseri:
I see, indeed it succeeds everywhere else

rgrinberg:
On second thought, how about we just get rid of this script? http isn't changing very often, we don't need the flexibility of generating a few dozen constructors.

mseri:
Fine but then let's put it on a gist, and add a README pointing at it so that we don't have to reinvent the wheel in the rare case we need to update it

rgrinberg:
You mean the script? It will always be present in the git history.

mseri:
I thought you merged this, why did you close it instead?

rgrinberg:
It was folded into the http package pr

PR #796, opened 2022-01-02
refactor: deprecate using attributes

rgrinberg:
instead of comments


mseri:
Yes we wanted to wait a bit before deprecating, could have been done already in 5.0, I had forgotten. Thanks

PR #797, opened 2022-01-02
refactor: stop relying on name mangling

rgrinberg:
expose a [Private] sub module for testing instead


mseri:
Does this need a lower bound on dune?

rgrinberg:
It does not

PR #798, opened 2022-01-02
refactor: deprecate connection value

rgrinberg:
In it's current form it's completely useless. It should be implemented by the individual
backends where it will gain meaning and functionality.


cohttp/src/cohttp.ml
```diff
@@ -2,7 +2,10 @@ module Accept = Accept
 module Auth = Auth
 module Body = Body
 module Conf = Conf
+
 module Connection = Connection
+[@@deprecated "Values of this type are useless and need to be used"]
```
mseri, line 7: I don’t understand the deprecation comment here

rgrinberg: Whoops. It is fixed

mseri:
Once we remove Connection.t, will the solution just be to rewirte the module separately in cohttp-lwt and cohttp-mirage?

rgrinberg:
I don't the details for mirage, but for Unix, I imagine it would look like:

```
module Context : sig
  type t
  val sockaddr : t -> Unix.sockaddr
end
```

I don't quite like the name `Connection` because it might not be all that meaningful behind a proxy or fcgi.

mseri:
Ok, so the idea is to deprecate it and replace it with something else. Maybe we should also introduce the replacement here, to also see what it looks like, and redefine `module Connection = Context` (this should allow old code to compile, right?)

rgrinberg:
Sure, I can work on introducing the replacement here. Although i'll note that no functionality is lost by getting rid of this module. In its current form, it cannot be used to do anything other than statelessness of the HTTP protocol.

mseri:
Ok, fair enough. Let's merge in the meantime then

psafont:
I found out this deprecation warning while updating to cohttp 6:
> Connection.t values are useless.

This happens because the callback parameter for creating a server is a function that allows users to inspect a Connection.t, this is used to keep track in sessions in the server (message switch)


> Although i'll note that no functionality is lost by getting rid of this module. In its current form, it cannot be used to do anything other than break the statelessness of the HTTP protocol.

But clients are allowed to get these values? what's altrenative are users being given here?

ansiwen:
We use the Connection.t for the logging of server callbacks (`callback` and `conn_closed`) with `Connection.to_string`, so that we can trace when which connection got closed. So, I don't understand, what can be used instead?

PR #799, opened 2022-01-02
refactor: explicit signature for async

rgrinberg:


cohttp-async/src/cohttp_async.ml
```diff
@@ -0,0 +1,7 @@
+module Body = Body
```
mseri, line 1: Afaik Body and Body_raw are essentially the same module. Is there a specific reason to keep them both? Should Body_raw at least be private?

rgrinberg: Just because I don't want to make any breaking changes in this PR. We can do deprecations or breakages separately.

PR #802, opened 2022-01-03
fix: fast comparison of request/response

rgrinberg:
in particular, we remove the costly allocation when the headers are
equal


mseri:
Thanks

PR #803, opened 2022-01-03
fix: deprecate more aliases

rgrinberg:
Deprecate Cohttp_{lwt,async}.{Request,Response}. Using Cohttp directly
is just as good, and the chain of aliases confuse readers, merlin,
documentation generators, etc.


mseri:
Nice!

PR #809, opened 2022-01-04
async(server): allow reading number of active connections

anuragsoni:


rgrinberg:
Does this need a lower bound?

anuragsoni:
> Does this need a lower bound?

`cohttp-async` depends on async >= 0.14.0 already so no further change to the package metadata is needed.

PR #811, opened 2022-01-04
refactor: move opam metadata to dune-project

rgrinberg:


rgrinberg:
Okay, finally this passes. Merging as the only material change is that I bumped the minimum dune version to 2.9

PR #812, opened 2022-01-05
feature: introduce http package

rgrinberg:
The new package contains no dependencies and is therefore useful for interoperability. In particular, the upcoming curl client will use this package instead of the cohttp package. I also remove the jsonm based code generation layer as the generated code hasn't been updated for ages.


samoht (approved):

samoht:
Is the plan to use this in httapf as well?

rgrinberg:
> Is the plan to use this in httapf as well?

I suppose it will be possible once we shed the transfer encoding fields from the request and response types. I'm not sure there's much desire left on the httpaf side though.

cohttp/src/code.ml
```diff
@@ -0,0 +1,142 @@
+(* Auto-Generated by 'ocaml generate.ml' *)
```
avsm, line 1:
```suggestion
```

avsm (approved):

mseri (approved):

PR #813, opened 2022-01-05
feature: Curl Based client

rgrinberg:
I'd like to get agreement on the interface before doing the Async version.

- [X] Lwt
- [X] Async
- [X] Tests
- [x] Docs


cohttp-curl-lwt/src/cohttp_curl_lwt.ml
```diff
@@ -0,0 +1,247 @@
...
+module Context = struct
+  type t = {
+    mt : Curl.Multi.mt;
+    wakeners : (Curl.t, Curl.curlCode Lwt.u) Hashtbl.t;
+    all_events : (Unix.file_descr, Lwt_engine.event list) Hashtbl.t;
+    mutable timer_event : Lwt_engine.event;
+  }
+
+  let create () =
+    (* Most of this is taken from https://github.com/ygrek/ocurl/blob/master/curl_lwt.ml *)
```
avsm, line 38: this should preserve copyright from https://github.com/ygrek/ocurl/blob/master/COPYING

cohttp-curl-lwt/src/cohttp_curl_lwt.ml
```diff
@@ -0,0 +1,247 @@
...
+        all_events = Hashtbl.create 32;
+        timer_event = Lwt_engine.fake_event;
+      }
+    in
+    let rec finished s =
+      match Curl.Multi.remove_finished t.mt with
+      | None -> ()
+      | Some (h, code) ->
+          (match Hashtbl.find_opt t.wakeners h with
+          | None -> Printf.eprintf "curl_lwt(%s): orphan handle" s
```
avsm, line 52: is this a debug thing, or should it raise an assertion failure?

rgrinberg: I'm trying to reproduce when this happens. If I don't manage to, I will turn this into an assertion.

rgrinberg:
@anuragsoni I cobbled together something for Async. Could you review it?

cohttp-curl-lwt/src/cohttp_curl_lwt.mli
```diff
@@ -0,0 +1,49 @@
...
+  module Expert : sig
+    val curl : _ t -> Curl.t
+  end
+end
+
+module Request : sig
+  type 'a t
+
+  val create :
+    ?timeout_ms:int ->
```
mseri, line 36: It's a win already for this :)

cohttp-curl-async/src/cohttp_curl_async.ml
```diff
@@ -0,0 +1,304 @@
...
+
+  type t = {
+    mt : Curl.Multi.mt;
+    wakeners : (Curl.t, Curl.curlCode Ivar.t) Hashtbl.t;
+    all_events : (Unix.file_descr, fd_events) Hashtbl.t;
+    mutable timeout : (unit, unit) Clock.Event.t option;
+  }
+
+  let create () =
+    (* Most of this is taken from https://github.com/ygrek/ocurl/blob/master/curl_lwt.ml *)
```
mseri, line 51: Is this comment a leftover from the port from lwt? should this also preserve the copyright from https://github.com/ygrek/ocurl/blob/master/COPYING?

rgrinberg: removed it. the async bindings are written from scratch at this point.

rgrinberg:
What's the deal with macos? Is the version of curl out of date or what?

avsm:
the macos running is usually `latest-release - 1`.   Try `macos-12` instead of `macos-latest` in the github action and see if that helps bump it up ...

mseri:
I can confirm that on MacOS Monterey with macports it works fine

mseri:
Why did we get suddenly all those EADDRINUSE errors?

rgrinberg:
> Why did we get suddenly all those EADDRINUSE errors?

I think it's because I'm using the same port for the lwt/async tests. Will fix.

rgrinberg:
For some reason the macos CI is not even running now :/

cohttp-curl-async/src/cohttp_curl_async.ml
```diff
@@ -0,0 +1,303 @@
...
+        all_events = Hashtbl.create 32;
+        timeout = None;
+      }
+    in
+    let rec finished s =
+      match Curl.Multi.remove_finished t.mt with
+      | None -> ()
+      | Some (h, code) ->
+          (match Hashtbl.find_opt t.wakeners h with
+          | None -> Printf.eprintf "curl_async(%s): orphan handle" s
```
anuragsoni, line 64: maybe use the async log module for this so it gets logged via the standard logging setup for async applications?

anuragsoni:
> @rgrinberg For some reason the macos CI is not even running now :/

According to the GitHub actions docs https://github.com/actions/virtual-environments`macos-12` isn't available yet. `macOS-latest` was pointing to 10.x series, and `macOS-11` is the newest version available.

cohttp-curl-async/src/cohttp_curl_async.ml
rgrinberg: I got rid of the logging.

.github/workflows/workflow.yml
```diff
@@ -33,12 +33,13 @@ jobs:
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
           dune-cache: ${{ matrix.os == 'ubuntu-latest' }}
+          opam-depext: ${{ matrix.os == 'macos-11' }}
```
mseri, line 36:
```suggestion
          opam-depext: true
```

mseri:
I think I know the issue now. [conf-libcurl](https://github.com/ocaml/opam-repository/blob/a63468ba24dede95f904c48645d560602296f2dd/packages/conf-libcurl/conf-libcurl.1/opam) does not include depexts for mac, so unless curl has been installed by hand for some reason, it always uses the old one shipped on mac. We cannot fix that here, I will send a patch to opam repository for that

.github/workflows/workflow.yml
smorimoto: If you enable it in all environments, you can remove the field. It's enabled by default.

mseri: Yeah. I thought it was disabled by default, we will just remove the line it was only a test

mseri:
This PR https://github.com/ocaml/opam-repository/pull/20459 should fix it

cohttp-curl/src/cohttp_curl.ml
```diff
@@ -0,0 +1,138 @@
...
+    let h = Curl.init () in
+    Curl.setopt h (CURLOPT_URL uri);
+    Curl.setopt h (CURLOPT_CUSTOMREQUEST (Http.Method.to_string method_));
+    let () =
+      match headers with
+      | None -> ()
+      | Some headers ->
+          let buf = Buffer.create 64 in
+          let headers =
+            Http.Header.fold
```
mseri, line 64: Why not using https://github.com/mirage/ocaml-cohttp/blob/master/http/src/http.ml#L174?
If it is faster to use buggers, we could rewrite that, to_frames and to_string to use the buffer

mseri:
About the design. I like that this is low level, but should we also provide an higher level interface akin to what we have now, where you can limit yuorself to do `Cohttp_curl_lwt.Client.get` to make your request and get your body (maybe keeping passing the sink manually or maybe not)?

cohttp-curl/src/cohttp_curl.ml
anuragsoni: I think in a separate pull request it could be interesting to explore using buffers to serialize the response and headers in one go (it’ll help avoid creating some temporary strings) and maybe even explore keeping a long lived buffer per request lifecycle so we don’t need to create a new one when sending a response every time.

cohttp-curl-lwt/src/cohttp_curl_lwt.ml
rgrinberg: It's gone.

rgrinberg:
> , but should we also provide an higher level interface akin to what we have now, where, e.g., you can limit yourself to do Cohttp_curl_lwt.Client.get to make your request and get your body (maybe keeping passing the sink manually or maybe not)?

`get` is useful because it allows you to elide the `input` parameter. The sink for the body result will have to remain.

If you find them useful then feel free to add them.

cohttp-curl/src/cohttp_curl.ml
rgrinberg: > Why not using https://github.com/mirage/ocaml-cohttp/blob/master/http/src/http.ml#L174?
If it is faster to use buffers, we could rewrite that, to_frames and to_string to use the buffer

It is because those functions add the `\r\n` which we don't need. I could share the code, but it would require a `Private` module in `Header`.

> I think in a separate pull request it could be interesting to explore using buffers to serialize the response and headers in one go (it’ll help avoid creating some temporary strings) and maybe even explore keeping a long lived buffer per request lifecycle so we don’t need to create a new one when sending a response every time.

Do you mean serialize the request? We don't really serialize responses in the client. Do you mean reuse the buffer between serializing the body and headers?

rgrinberg: > It is because those functions add the \r\n which we don't need. I could share the code, but it would require a Private module in Header.

I just tried, and there's no good way to do this unfortunately.

anuragsoni:
The changes look good to me. Thanks for adding this! The low level interface is a good start as that can let people experiment with the curl backend, and a higher level interface can be added in the future as separate PRs maybe?

rgrinberg:
I don't get the error from opam-lint:

```
cohttp-async.opam: OK
cohttp-bench.opam: OK
cohttp-curl-async.opam: changes needed:
  "core" {with-test & >= "v0.14.1"}        [from cohttp-curl-async/test]
cohttp-curl-lwt.opam: changes needed:
  "uri" {>= "4.2.0"}                       [from cohttp-curl-lwt/src]
```

Why does it think I need those changes?

rgrinberg:
Never mind, I fixed the constraints.

rgrinberg:
Okay, the lint seems wrong:

```
cohttp-curl-async.opam: OK
cohttp-curl-lwt.opam: changes needed:
  "uri"                                    [from cohttp-curl-lwt/src] (remove {with-test})
```

`uri` is definitely a test dependency. There's no reference to `Uri` in this library.

cohttp-curl/src/cohttp_curl.ml
mseri: > It is because those functions add the `\r\n` which we don't need.

The to_frames function does not add them, it should be equivalent to what you have here. The only difference is that it uses sprintf i stead of a buffer

cohttp-curl-lwt/bin/curl.ml
```diff
@@ -0,0 +1,96 @@
...
+module Curl = Cohttp_curl_lwt
+module Sexp = Sexplib0.Sexp
+
+let src =
+  Logs.Src.create "cohttp.lwt.curl" ~doc:"Cohttp Lwt curl implementation"
+
+module Log = (val Logs.src_log src : Logs.LOG)
+
+let client uri ofile meth' =
+  Log.debug (fun d -> d "Client with URI %s" (Uri.to_string uri));
```
mseri, line 12: Maybe it complains about the use of uri because it is needed here

cohttp-curl-lwt/bin/dune
```diff
@@ -0,0 +1,12 @@
...
+ (name curl)
+ (libraries
+  sexplib0
+  cohttp
+  cohttp_curl_lwt
+  lwt.unix
+  lwt
+  logs
+  logs.cli
+  fmt.tty
```
mseri, line 11: This also needs `uri`

cohttp-curl-async/src/dune
```diff
@@ -0,0 +1,11 @@
+(library
+ (name cohttp_curl_async)
+ (public_name cohttp-curl-async)
```
mseri, line 3:
```suggestion
 (public_name cohttp-curl-async)
 (package cohttp-curl-async)
```
Why is package not needed here? Is it because the public name and the package name coincide?

cohttp-curl-lwt/bin/dune
```diff
@@ -0,0 +1,12 @@
...
+  cmdliner))
```
mseri, line 12:
```suggestion
  cmdliner)
  (public_name cohttp-curl)
  (package cohttp-curl-lwt))
```
Should we install the binary like cohttp-lwt-unix does? The name omits lwt since it may conflict with cohttp-lwt-unix i stalled binaries. We should probably reconsider installing them at all or create a separate `bin` package. 

It could be part of a separate pr, not asking to do it here, but I want to understand what to do with this binary: does a cohttp-async install with test try to build this package as well for example or dune is smart enough to ignore it?

cohttp-curl-lwt/bin/curl.ml
rgrinberg: This is a binary that isn't public though. It's just there for easy testing.

cohttp-curl-lwt/bin/dune
rgrinberg: > Should we install the binary like cohttp-lwt-unix does? The name omits lwt since it may conflict with cohttp-lwt-unix i stalled binaries. We should probably reconsider installing them at all or create a separate bin package.

Probably not. I don't see  a use for it other than testing.

> It could be part of a separate pr, not asking to do it here, but I want to understand what to do with this binary: does a cohttp-async install with test try to build this package as well for example or dune is smart enough to ignore it?

Dune is smart enough to ignore it. Since the binary doesn't belong to any package, it will not be mentioned in the .install file in any way and dune will not build it.

cohttp-curl-async/src/cohttp_curl_async.mli
```diff
@@ -0,0 +1,69 @@
...
+  (** A source defines where the request body is read from *)
+
+  type t
+
+  val empty : t
+  val string : string -> t
+end
+
+module Context : sig
+  (** A context shares the curl event handling logic for all all curl requests
```
mseri, line 22:
```suggestion
  (** A context shares the curl event handling logic for all curl requests
```

cohttp-curl-async/src/cohttp_curl_async.mli
```diff
@@ -0,0 +1,69 @@
...
+  module Expert : sig
+    val curl : _ t -> Curl.t
+  end
+end
+
+module Request : sig
+  (** Http requests *)
+
+  type 'a t
+  (** ['a t] represents an http request ['a] determines who the response body is
```
mseri, line 50:
```suggestion
  (** ['a t] represents an http request ['a] determines how the response body is
```
Or maybe `who handles the response body`

cohttp-curl-lwt/src/cohttp_curl_lwt.mli
```diff
@@ -0,0 +1,68 @@
...
+  (** A source defines where the request body is read from *)
+
+  type t
+
+  val empty : t
+  val string : string -> t
+end
+
+module Context : sig
+  (** A context shares the curl event handling logic for all all curl requests
```
mseri, line 22:
```suggestion
  (** A context shares the curl event handling logic for all curl requests
```

cohttp-curl-lwt/src/cohttp_curl_lwt.mli
```diff
@@ -0,0 +1,68 @@
...
+  module Expert : sig
+    val curl : _ t -> Curl.t
+  end
+end
+
+module Request : sig
+  (** Http requests *)
+
+  type 'a t
+  (** ['a t] represents an http request ['a] determines who the response body is
```
mseri, line 49:
```suggestion
  (** ['a t] represents an http request ['a] determines how the response body is
```

rgrinberg:
> Not sure how to fix the macos issue, it should have been fixed by the opam-repository update. Otherwise this seems fine to merge imho

Sure, but I at least want to fix the nix CI

mseri:
Does it work if you call it with `export PKG_CONFIG_PATH=$HOME/.nix-profile/lib/pkgconfig"` ?

mseri:
It is annoying that this is failing both on the mac builder and nix. For the latter I have no experience unfortunately, for mac it seems to fail only for 4.11 or earlier version. I wonder what changed. 

With 4.13 and macports (on my mac) it works seamlessly (I cannot test older compilers until monday). I don’t know what else to do to fix this in the Ci (besides stopping testing those older compilers)

rgrinberg:
I found the issue with nix. The following line in conf-libcurl:

```
depends: ["conf-pkg-config" {build & os != "macos"}]
```

Means that solving for the package set on a mac yields a wrong derivation that excludes pkg-config from the dependencies.

PR #814, opened 2022-01-06
fix: no do not stringify for comparison

rgrinberg:
it is needlessly slow

### DIFF
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -1,5 +1,7 @@
 module Transfer = struct
   type encoding = Chunked | Fixed of int64 | Unknown
+
+  let compare_encoding (x : encoding) (y : encoding) = Stdlib.compare x y
 end
 
 module Header = struct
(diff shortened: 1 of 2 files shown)


http/src/http.ml
```diff
@@ -1,5 +1,7 @@
 module Transfer = struct
   type encoding = Chunked | Fixed of int64 | Unknown
+
+  let compare_encoding (x : encoding) (y : encoding) = Stdlib.compare x y
```
mseri, line 4: Is this just to expose encoding’s comparison in a type enforcing way or some kind of specialization here also helps speed the code up?

mseri (approved):

http/src/http.ml
rgrinberg: No, there's no specialized comparison generated. It's just to prevent things like `Transfer.compare_encoding 42 24`

PR #816, opened 2022-01-06
fix: use raise_notrace in header

rgrinberg:


mseri:
Neat, I was not aware of it!

PR #817, opened 2022-01-06
refactor(cohttp-async): insantiate IO only once

rgrinberg:
cc @anuragsoni


anuragsoni:
The changes here look good to me!

PR #818, opened 2022-01-06
cohttp.headers: faster faster comparison

gasche:
see
  https://github.com/mirage/ocaml-cohttp/pull/778#issuecomment-1006459031

### DIFF
--- a/http/bench/dune
+++ b/http/bench/dune
@@ -2,10 +2,10 @@
  (name bench)
  (libraries http core core_bench))
 
-(rule
- (alias runtest)
- (package http)
- (enabled_if
-  (= %{ocaml-config:word_size} 64))
- (action
-  (run ./bench.exe time cycles)))
+; (rule
+;  (alias runtest)
+;  (package http)
+;  (enabled_if
+;   (= %{ocaml-config:word_size} 64))
+;  (action
+;   (run ./bench.exe time cycles)))
--- a/http/bench/dune
+++ b/http/bench/dune
@@ -2,10 +2,10 @@
  (name bench)
  (libraries http core core_bench))
 
-; (rule
-;  (alias runtest)
-;  (package http)
-;  (enabled_if
-;   (= %{ocaml-config:word_size} 64))
-;  (action
-;   (run ./bench.exe time cycles)))
+(rule
+ (alias bench)
+ (package http)
+ (enabled_if
+  (= %{ocaml-config:word_size} 64))
+ (action
+  (run ./bench.exe time cycles)))
--- a/http/bench/dune
+++ b/http/bench/dune
@@ -8,4 +8,4 @@
  (enabled_if
   (= %{ocaml-config:word_size} 64))
  (action
-  (run ./bench.exe -quota 5s)))
+  (run ./bench.exe time cycles)))
--- a/http/bench/bench.ml
+++ b/http/bench/bench.ml
@@ -2,26 +2,32 @@ module Command = Core.Command
 module Staged = Core.Staged
 open Core_bench
 
-let header_mem =
-  let variations =
-    [
-      [ ("foo", "bar") ];
-      [ ("transfer-encoding", "bar") ];
-      [ ("traNsfer-eNcoding", "bar") ];
-      [ ("traNsfer eNcoding", "bar") ];
-      [ ("traNsfer-eNcodkng", "bar") ];
-      [ ("transfer-encodinG", "bar") ];
-    ]
-  in
-  let key = "transfer-encoding" in
-  List.mapi
-    (fun i header ->
-      let header = Http.Header.of_list header in
-      Bench.Test.create
-        ~name:("Header.mem - " ^ string_of_int i)
-        (fun () ->
-          let (_ : bool) = Http.Header.mem header key in
-          ()))
-    variations
+let header_names = [
+  "Accept";
+  "Accept-Encoding";
+  "Accept-Language";
+  "Cache-Control";
+  "Connection";
+  "Host";
+  "If-Modified-Since";
+  "If-None-Match";
+  "Origin";
+  "Referer";
+  "Sec-Fetch-Dest";
+  "Sec-Fetch-Mode";
+  "Sec-Fetch-Site";
+]
 
-let () = Command.run @@ Bench.make_command header_mem
+let header =
+  header_names
+  |> List.map (fun s -> (s, "value"))
+  |> Http.Header.of_list
+
+let bench_header_mem =
+  Bench.Test.create ~name:"Header.mem"
+    (fun () ->
+       List.iter (fun key -> assert (Http.Header.mem header key))
+         header_names
+    )
+
+let () = Command.run @@ Bench.make_command [ bench_header_mem ]
--- a/http/bench/bench.ml
+++ b/http/bench/bench.ml
@@ -2,32 +2,28 @@ module Command = Core.Command
 module Staged = Core.Staged
 open Core_bench
 
-let header_names = [
-  "Accept";
-  "Accept-Encoding";
-  "Accept-Language";
-  "Cache-Control";
-  "Connection";
-  "Host";
-  "If-Modified-Since";
-  "If-None-Match";
-  "Origin";
-  "Referer";
-  "Sec-Fetch-Dest";
-  "Sec-Fetch-Mode";
-  "Sec-Fetch-Site";
-]
+let header_names =
+  [
+    "Accept";
+    "Accept-Encoding";
+    "Accept-Language";
+    "Cache-Control";
+    "Connection";
+    "Host";
+    "If-Modified-Since";
+    "If-None-Match";
+    "Origin";
+    "Referer";
+    "Sec-Fetch-Dest";
+    "Sec-Fetch-Mode";
+    "Sec-Fetch-Site";
+  ]
 
 let header =
-  header_names
-  |> List.map (fun s -> (s, "value"))
-  |> Http.Header.of_list
+  header_names |> List.map (fun s -> (s, "value")) |> Http.Header.of_list
 
 let bench_header_mem =
-  Bench.Test.create ~name:"Header.mem"
-    (fun () ->
-       List.iter (fun key -> assert (Http.Header.mem header key))
-         header_names
-    )
+  Bench.Test.create ~name:"Header.mem" (fun () ->
+      List.iter (fun key -> assert (Http.Header.mem header key)) header_names)
 
 let () = Command.run @@ Bench.make_command [ bench_header_mem ]
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -25,8 +25,11 @@ module Header = struct
           (* If [i' > len], what remains to be compared is strictly
              less than a word long, use byte-per-byte comparison. *)
           if i' > len then byte_loop a b i len
-          else if String.get_int64_ne a i = String.get_int64_ne b i then
-            word_loop a b i' len
+          else if
+            Bytes.(
+              get_int64_ne (unsafe_of_string a) i
+              = get_int64_ne (unsafe_of_string b) i)
+          then word_loop a b i' len
           else
             (* If the words at [i] differ, it may due to a case
                difference; we check the individual bytes of this
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -5,6 +5,7 @@ module Transfer = struct
 end
 
 module Header = struct
+  external string_unsafe_get64 : string -> int -> int64 = "%caml_string_get64u"
   (* [caseless_equal a b] must be equivalent to
      [String.equal (String.lowercase_ascii a) (String.lowercase_ascii b)]. *)
   let caseless_equal a b =
@@ -25,10 +26,7 @@ module Header = struct
           (* If [i' > len], what remains to be compared is strictly
              less than a word long, use byte-per-byte comparison. *)
           if i' > len then byte_loop a b i len
-          else if
-            Bytes.(
-              get_int64_ne (unsafe_of_string a) i
-              = get_int64_ne (unsafe_of_string b) i)
+          else if string_unsafe_get64 a i = string_unsafe_get64 b i
           then word_loop a b i' len
           else
             (* If the words at [i] differ, it may due to a case
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -5,21 +5,49 @@ module Transfer = struct
 end
 
 module Header = struct
+  (* [caseless_equal a b] must be equivalent to
+     [String.equal (String.lowercase_ascii a) (String.lowercase_ascii b)]. *)
   let caseless_equal a b =
     if a == b then true
     else
       let len = String.length a in
       len = String.length b
+      (* Note: at this point we konw that [a] and [b] have the same length. *)
       &&
-      let stop = ref false in
-      let idx = ref 0 in
-      while (not !stop) && !idx < len do
-        let c1 = String.unsafe_get a !idx in
-        let c2 = String.unsafe_get b !idx in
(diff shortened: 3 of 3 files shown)


gasche:
I'm unfamiliar with cohttp development, so here are some basic questions:

- How does one benchmark this change? @mseri, would you by chance be willing to do the benchmarking? (At least giving me the command to run to benchmark, and what measure to look at, would be helpful.)
- Where do I write tests for this function? There is a risk for off-by-one error / boundary conditions here, that could be solved by just checking the behavior exhaustively for all small lengths. But I did not find where unit tests are located.

rgrinberg:
> Where do I write tests for this function? There is a risk for off-by-one error / boundary conditions here, that could be solved by just checking the behavior exhaustively for all small lengths. But I did not find where unit tests are located.

The `http` package is very new and the existing tests are in the `cohttp` package. You can add the first test to `http/test` though. To expose the function for unit testing, add a `Private` module either in `Http` or in `Http.Header`.

gasche:
The CI says that formatting is wrong, and that the String primitives are not exposed on older OCaml versions (4.08-4.12). Both can be easily fixed (Bytes has the primitives exposed), but I would welcome performance feedback first.

rgrinberg:
I added a very simple benchmark using core_bench that isn't at all realistic but can be easily expanded to include more cases.

gasche:
Thanks!

Would you by chance be willing to push a similar skeleton for tests, in addition to benchmarks? I don't have your favorite dune-based testing infrastructure in muscle memory, so it would be a bit harder for me. (I hoped to avoid this work by letting a more familiar contributor make a PR :-)

Note (unrelated): the new version only significantly improves over the previous version when the first difference in the string is far in the two strings; in particular, in the common case where the two strings are in fact equal. (When they are  distinct, in the common case their length will differ, so they will be discarded immediately with both the new and old versions.)

mseri:
Very cool, thanks a lot.

Fwiw the headers module also have fuzzy tests (which we can probably move to the http package in a later PR), so that should help finding out issues in this function.

In the past I did something on the line of what @rgrinberg added, but local to my machine, and I used a more global micro benchmark like the one I added to [multicore httpaf benchmarks](https://github.com/ocaml-multicore/retro-httpaf-bench/blob/master/cohttp-lwt-unix/cohttp_lwt_unix.ml) to check if it affected the overall speed in any way. Nothing scientific in practice.

mseri:
> Would you by chance be willing to push a similar skeleton for tests, in addition to benchmarks? I don't have your favorite dune-based testing infrastructure in muscle memory, so it would be a bit harder for me. (I hoped to avoid this work by letting a more familiar contributor make a PR :-)

In 1 hour or so I should be able to spend some time to move the headers tests here

mseri:
I don't have wrk2 available to try the other tests, but your microbenchmark seems quite promising

mseri:
The Ocaml-CI is actually successfully testing everything, but fails since it tries to build the benchmark anyway.

The github CI, which was succeeding until yesterday, now fails to run the js tests for some reason 🤷

gasche:
Shouldn't `core_bench` be added as a `test` dependency in `http.opam`, so that the CI has a chance to think of installing it?

gasche:
Ah, I guess: the benchmark is not actually declared as a test (so it is not run by `runtest` and it is not obviously a test dependency); the build failure comes from the `@all` or `@check` targets that still try to build it anyway.

avsm:
If it's a short running benchmark, then I'd prefer it to be declared as a test dependency and always run. (this sometimes finds crashes for us; the actual results aren't relevant in the test run obviously)

mseri:
```
bench alias http/bench/runtest (exit 1)
(cd _build/default/http/bench && ./bench.exe)
Estimated testing time 40s (4 benchmarks x 10s). Change using '-quota'.
Uncaught exception:
  
  (Failure
   "conversion from int64 to int failed: -4253731984 is out of range")
```
Looks like core_bench does not like 32 bit systems

mseri:
I have tried to run the benchmark on my old i5 a few times in fact.
Using `core_bench` I get the following:

```
Estimated testing time 40s (4 benchmarks x 10s). Change using '-quota'.

New comparison function:
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │  10.19ns │   4.00w │     17.04% │
│ Header.mem - 1 │   6.64ns │   4.00w │     11.11% │
│ Header.mem - 2 │  59.80ns │   4.00w │    100.00% │
│ Header.mem - 3 │  40.94ns │   4.00w │     68.46% │
└────────────────┴──────────┴─────────┴────────────┘

Old comparison function:
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │   9.95ns │   4.00w │     19.57% │
│ Header.mem - 1 │   6.99ns │   4.00w │     13.74% │
│ Header.mem - 2 │  50.86ns │   4.00w │    100.00% │
│ Header.mem - 3 │  33.17ns │   4.00w │     65.21% │
└────────────────┴──────────┴─────────┴────────────┘
```

From which it seems that for such short strings and my old computer the old implementation is faster.
I am curious to try on the laptop, but I will likely not be able before next week.

mseri:
Tested on the macbook (core i7 from 2019, in battery mode).
```
Old comparison
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │  11.57ns │   4.00w │     21.70% │
│ Header.mem - 1 │   7.75ns │   4.00w │     14.53% │
│ Header.mem - 2 │  53.32ns │   4.00w │    100.00% │
│ Header.mem - 3 │  35.59ns │   4.00w │     66.75% │
└────────────────┴──────────┴─────────┴────────────┘


New comparison
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │  11.18ns │   4.00w │     21.98% │
│ Header.mem - 1 │   7.77ns │   4.00w │     15.28% │
│ Header.mem - 2 │  50.84ns │   4.00w │    100.00% │
│ Header.mem - 3 │  37.04ns │   4.00w │     72.85% │
└────────────────┴──────────┴─────────┴────────────┘
```

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: build clean test clean
 
 build:
-	dune build
+	dune build @install
```
rgrinberg, line 4: Why this change? `@install` builds strictly less and makes it easy to forget to avoid examples/tests.

mseri: It was to make it build without requiring core_bench. I am going to send a PR to re-activate it.
We still build all on the ocaml-CI in any case

mseri:
I have added one more check to the benchmark, of a difference that happens towards the end of the string.
Now one can see a large improvement
```
Old comparison
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │  11.73ns │   4.00w │     23.36% │
│ Header.mem - 1 │   7.61ns │   4.00w │     15.16% │
│ Header.mem - 2 │  50.22ns │   4.00w │    100.00% │
│ Header.mem - 3 │  35.62ns │   4.00w │     70.93% │
│ Header.mem - 4 │  48.70ns │   4.00w │     96.98% │
└────────────────┴──────────┴─────────┴────────────┘

New comparison
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │  11.58ns │   4.00w │     17.09% │
│ Header.mem - 1 │   7.78ns │   4.00w │     11.48% │
│ Header.mem - 2 │  67.79ns │   4.00w │    100.00% │
│ Header.mem - 3 │  46.45ns │   4.00w │     68.52% │
│ Header.mem - 4 │  21.38ns │   4.00w │     31.54% │
└────────────────┴──────────┴─────────┴────────────┘
```

mseri:
There are (going to be) two more cases in the benchmark on the master branch, which contain differences only towards the end of the string. There you can see the improvement from the new PR

```
Old implementation
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │   7.47ns │   4.00w │     23.44% │
│ Header.mem - 1 │   4.87ns │   4.00w │     15.28% │
│ Header.mem - 2 │  31.56ns │   4.00w │     98.96% │
│ Header.mem - 3 │  22.23ns │   4.00w │     69.71% │
│ Header.mem - 4 │  30.85ns │   4.00w │     96.74% │
│ Header.mem - 5 │  31.89ns │   4.00w │    100.00% │
└────────────────┴──────────┴─────────┴────────────┘

New implementation
┌────────────────┬──────────┬─────────┬────────────┐
│ Name           │ Time/Run │ mWd/Run │ Percentage │
├────────────────┼──────────┼─────────┼────────────┤
│ Header.mem - 0 │   7.34ns │   4.00w │     17.17% │
│ Header.mem - 1 │   4.97ns │   4.00w │     11.62% │
│ Header.mem - 2 │  42.74ns │   4.00w │    100.00% │
│ Header.mem - 3 │  29.68ns │   4.00w │     69.44% │
│ Header.mem - 4 │  39.90ns │   4.00w │     93.36% │
│ Header.mem - 5 │  13.44ns │   4.00w │     31.45% │
└────────────────┴──────────┴─────────┴────────────┘
```

gasche:
The best case for such a change is when we compare two strings (large enough to span several words) that are actually equal. This should be the case for benchmark 1, if I read the code correctly -- it compares `transfer-encoding` with itself. This is what I did in my micro-test (I think the cases of different strings are going to be instantenous anyway in practice, so they matter less; it would be nice I think if the benchmark used header-key names that occur in real life, not just weird names). I don't understand why the proposed benchmark does not seem to pick any difference in this case of comparing a string with itself, when my silly microbenchmark did.

Then: the code could be made faster by using an unsafe access (the safe access adds two bound-checks that require two memory reads each, which diminishes the performance benefits when we are trying to reduce memory reads in the first place).

gasche:
I pushed a benchmark which I found more interesting (it is using real header keys from a request), and a version using the unsafe-access primitives. Because the benchmark does more comparisons, in most cases with different lengths (discarded right away), we can expect the differences between implementations to result in smaller changes than with my previous silly micro-benchmark.

Below are some numbers on my machine:

Performance of `master` (`dune exec -- http/bench/bench.exe time cycles`):

```
│ Name       │ Time/Run │ Cycls/Run │ mWd/Run │ Percentage │
├────────────┼──────────┼───────────┼─────────┼────────────┤
│ Header.mem │ 388.51ns │   895.10c │  52.00w │    100.00% │
```

Performance with bound-checked array accesses:

```
│ Name       │ Time/Run │ Cycls/Run │ mWd/Run │ Percentage │
├────────────┼──────────┼───────────┼─────────┼────────────┤
│ Header.mem │ 379.83ns │   875.10c │  52.00w │    100.00% │
```

Performance with unsafe accesses:

```
│ Name       │ Time/Run │ Cycls/Run │ mWd/Run │ Percentage │
├────────────┼──────────┼───────────┼─────────┼────────────┤
│ Header.mem │ 345.69ns │   796.44c │  52.00w │    100.00% │
```

mseri:
Super, thanks!

mseri:
Indeed, on my laptop I get
```
Old compare
┌────────────┬──────────┬───────────┬─────────┬────────────┐
│ Name       │ Time/Run │ Cycls/Run │ mWd/Run │ Percentage │
├────────────┼──────────┼───────────┼─────────┼────────────┤
│ Header.mem │ 434.70ns │    1.13kc │  52.00w │    100.00% │
└────────────┴──────────┴───────────┴─────────┴────────────┘

New compare
┌────────────┬──────────┬───────────┬─────────┬────────────┐
│ Name       │ Time/Run │ Cycls/Run │ mWd/Run │ Percentage │
├────────────┼──────────┼───────────┼─────────┼────────────┤
│ Header.mem │ 372.00ns │   964.22c │  52.00w │    100.00% │
└────────────┴──────────┴───────────┴─────────┴────────────┘
```

mseri (approved):

mseri:
If also @rgrinberg is happy with this, I can rebase it on master and merge

http/bench/dune
```diff
@@ -1,3 +1,9 @@
 (executable
  (name bench)
  (libraries http core core_bench))
+
+(rule
+ (alias runtest)
+ (package http)
+ (action
+  (run ./bench.exe)))
```
rgrinberg: I don't think this should be done. The benchmarks are quite slow.

mseri: It is just 10 extra seconds, we could reduce it further to 1 or 2
seconds or remove it. If we leave it there the ocaml-CI fails. 

An alternative would be to delete the dune file and put there a README for building and running the benchmarks

rgrinberg: Why does it fail if we leave it?

mseri: `dune build` in the GH CI tries to build everything, as well as `dune build @all` in ocaml CI. But I was thinking to also remove core_bench from the test dependencies. You would leave it there?

gasche: Maybe you could create a separate `cotthp-benchmarks` package that correspond to the benchmarks (not installed by default), and includes the benchmarking dependencies?

mseri: Sounds nice, it would also be a place to put test servers to use for benchmarking purposes

rgrinberg: >  But I was thinking to also remove core_bench from the test dependencies. You would leave it there?

Yes, I would leave it there. You can also make sure the tests build the benchmark suite but not run it.

Gasche's suggestion also sounds fine.

mseri: I am going to leave it here for now. And open a separate issue to discuss the opportunity to have a package with executables for testing

mseri:
I have rebased it on master, linted the code and disabled running the benchmarks. I think this is good to merge

http/bench/dune
```diff
@@ -2,10 +2,10 @@
...
-  (= %{ocaml-config:word_size} 64))
- (action
-  (run ./bench.exe -quota 5s)))
+; (rule
+;  (alias runtest)
+;  (package http)
+;  (enabled_if
+;   (= %{ocaml-config:word_size} 64))
+;  (action
+;   (run ./bench.exe time cycles)))
```
mseri, line 11: I kept this commented to hint at how to run it

rgrinberg: I'll just attach it to a separate alias.

rgrinberg: Alternatively, you can also make the argument `-quota 1x`. This way it will not slow down the test suite.  We should also discard the stdout as it's garbage in this case.

mseri: There is no point in compiling it and running it, I have added a `runtest` alias.

I am going to follow up with the optional c`ohttp-bench` package in the next few days and move everything there, including the latency benchmarks for `cohttp-{lwt,async}-server`. It is convenient to keep it at hand and will get rid of a lot of unnecessary test dependencies in the `http` package

mseri:
Thanks @gasche !

I am going to update the changelog inlcuding all the other recent PRs tomorrow

PR #819, opened 2022-01-07
Use new request parser

anuragsoni:
Hello,

I'll preface this by saying that this is far from being ready, and I've had to comment/break a lot of things to get a minimal server example working. 

#### Background
I've been working on a HTTP request (and chunked body) parser and prototype for a [HTTP 1.1 server](https://github.com/anuragsoni/shuttle/tree/d850f414ef5e3c750fe74bfe3434d97674e3841e/http/src) and I decided to use the Cohttp types as the base layer for my explorations. The server experiments in `shuttle` have shown promise, but instead of adding yet another server library I'd like to explore if it's possible to integrate some of that work directly into Cohttp. I'm opening this really early in the process to see if an approach similar to what I'm proposing here will fit well with the Cohttp maintainer's vision (and to hear if there are suggestions on improving the IO layer even further).

I am starting with `cohttp-lwt-unix` first as that's the cohttp backend I use at work, and I wanted to try my work on a non async backend to validate that the work in `shuttle` can be ported to non async libraries as well. All of the new code in this diff is a direct port of implementations from the `shuttle` library.

I am using the test payload referenced in https://github.com/mirage/ocaml-cohttp/issues/328 to measure progress.

**Disclaimer**: All of these test runs were made on my laptop, and if someone has a better hardware setup for such benchmarking then I'd appreciate getting some more measurements from there.

**Library Versions:** Cohttp (the version in the current trunk as of January 6 2022), LWT (5.5.0), OCaml (4.13.1)

I ran an initial set of tests using `wrk2` using 1000 active connections and two different request rates (80_000, and 100_000). The results from the runs on my laptop can be seen at https://gist.github.com/anuragsoni/eb095a7f4c3aec1f4ba65b9866d9f9d3 . The summary from that gist is:
* Cohttp_lwt_unix from the current trunk had trouble getting more than 60,000 requests per second and had average latencies around 4-7 seconds with tail latencies > 10 seconds at times
* Cohttp_lwt_unix with the new request parser and still using Lwt_io via conduit was able to comfortable serve 80_000 RPS with average latency around 2ms and tail latencies around 7ms. When attempting to server 100_000 RPS the average latencies degraded to 1.2 seconds with tail latencies of 2 seconds
* As an experiment I tried to run cohttp-lwt-unix server that doesn't use conduit + Lwt_io for the read path, and instead just uses `Lwt_unix.file_descr` paired with the buffer that's part of this PR. With that the server was able to handle 100_000. RPS with average latencies of 2ms with tail latencies of 7.6ms - link [here](https://github.com/mirage/ocaml-cohttp/pull/819/commits/8876adb574e973ce661b41296d167e2c556fcbfe)
* Httpaf_lwt_unix has been used as the benchmark here and it can easily server 100_000 rps with an average latency of 2.5ms with a tail of 9ms

Apologies for wall of text, but I wanted to start a discussion even if I only have a partially working solution so far, as I'd like to get some insights into whether the proposed implementation here makes sense as a starting point, and I'm also hoping to learn how best to test/benchmark such work so we can have an accurate picture of any potential improvements.

Thanks,
Anurag

### DIFF
--- a/cohttp-async/src/io.ml
+++ b/cohttp-async/src/io.ml
@@ -69,28 +69,23 @@ module IO = struct
   let ( >>= ) = Deferred.( >>= )
   let return = Deferred.return
 
-  type ic = Reader.t
+  type ic = Input_channel.t
   type oc = Writer.t
   type conn = unit
 
   let read_line =
     check_debug
+      (fun ic -> Input_channel.read_line_opt ic)
       (fun ic ->
-        Reader.read_line ic >>| function `Ok s -> Some s | `Eof -> None)
-      (fun ic ->
-        Reader.read_line ic >>| function
-        | `Ok s ->
+        Input_channel.read_line_opt ic >>| function
+        | Some s ->
             Log.debug (fun fmt -> fmt "<<< %s" s);
             Some s
-        | `Eof ->
+        | None ->
             Log.debug (fun fmt -> fmt "<<<EOF");
             None)
 
-  let read ic len =
-    let buf = Bytes.create len in
-    Reader.read ic ~len buf >>| function
-    | `Ok len' -> Bytes.To_string.sub buf ~pos:0 ~len:len'
-    | `Eof -> ""
+  let read ic len = Input_channel.read ic len
 
   let write =
     check_debug
@@ -102,6 +97,8 @@ module IO = struct
         Writer.write oc buf;
         return ())
 
+  let refill ic = Input_channel.refill ic
+  let with_input_buffer ic = Input_channel.with_input_buffer ic
   let flush = Writer.flushed
 end
 
--- a/cohttp-lwt-unix/src/input_channel.ml
+++ b/cohttp-lwt-unix/src/input_channel.ml
@@ -1,0 +1,25 @@
+open Lwt.Infix
+module Bytebuffer = Cohttp_lwt.Private.Bytebuffer
+
+type t = { buf : Bytebuffer.t; chan : Lwt_io.input_channel }
+
+let refill ic buf ~pos ~len =
+  Lwt.catch
+    (fun () ->
+      Lwt_io.read_into ic buf pos len >|= fun c -> if c > 0 then `Ok c else `Eof)
+    (function Lwt_io.Channel_closed _ -> Lwt.return `Eof | exn -> raise exn)
+
+let create ?(buf_len = 0x4000) chan = { buf = Bytebuffer.create buf_len; chan }
+let read_line_opt t = Bytebuffer.read_line t.buf (refill t.chan)
+let read t count = Bytebuffer.read t.buf (refill t.chan) count
+let refill t = Bytebuffer.refill t.buf (refill t.chan)
+
+let with_input_buffer t ~f =
+  let buf = Bytebuffer.unsafe_buf t.buf in
+  let pos = Bytebuffer.pos t.buf in
+  let len = Bytebuffer.length t.buf in
+  let res, consumed = f (Bytes.unsafe_to_string buf) ~pos ~len in
+  Bytebuffer.drop t.buf consumed;
+  res
+
+let close t = Lwt_io.close t.chan
--- a/cohttp/src/s.ml
+++ b/cohttp/src/s.ml
@@ -39,6 +39,11 @@ module type IO = sig
   type conn
   (** [conn] represents the underlying network flow *)
 
+  val refill : ic -> [ `Ok | `Eof ] t
+
+  val with_input_buffer :
+    ic -> f:(string -> pos:int -> len:int -> 'a * int) -> 'a
+
   val read_line : ic -> string option t
   (** [read_line ic] will read a single line terminated by CR or CRLF from the
       input channel [ic]. It returns {!None} if EOF or other error condition is
--- a/cohttp/test/test_request.ml
+++ b/cohttp/test/test_request.ml
@@ -218,7 +218,7 @@ let parse_request_uri_host_query_no_slash _ =
   parse_request_uri_ r bad_request "parse_request_uri_host_query_no_slash"
 
 let parse_request_connect _ =
-  let r = "CONNECT vpn.example.net:443 HTTP/1.1\r\n" in
+  let r = "CONNECT vpn.example.net:443 HTTP/1.1\r\n\r\n" in
   let uri = `Ok (Uri.of_string "//vpn.example.net:443") in
   parse_request_uri_ r uri "parse_request_connect"
 
--- a/examples/bench/async_server.ml
+++ b/examples/bench/async_server.ml
@@ -1,0 +1,27 @@
+open Core
+open Async
+open Cohttp_async
+
+let text = String.make 2053 'a'
+let handler ~body:_ _sock _req = Server.respond_string text
+
+let start_server port () =
+  Cohttp_async.Server.create ~on_handler_error:`Raise
+    (Tcp.Where_to_listen.of_port port)
+    handler
+  >>= fun server ->
+  Deferred.forever () (fun () ->
+      after Time.Span.(of_sec 0.5) >>| fun () ->
+      Log.Global.printf "Active connections: %d" (Server.num_connections server));
+  Deferred.never ()
+
+let () =
+  let module Command = Async_command in
+  Command.async_spec ~summary:"Start a hello world Async server"
+    Command.Spec.(
+      empty
+      +> flag "-p"
+           (optional_with_default 8080 int)
+           ~doc:"int Source port to listen on")
+    start_server
+  |> Command.run
--- a/examples/bench/lwt_unix_server.ml
+++ b/examples/bench/lwt_unix_server.ml
@@ -1,0 +1,15 @@
+open Cohttp_lwt_unix
+
+let text = String.make 2053 'a'
+
+let server_callback _conn _req _body =
+  Server.respond_string ~status:`OK ~body:text ()
+
+let main () =
+  Server.create ~backlog:11_000 (Server.make ~callback:server_callback ())
+
+let () =
+  Printexc.record_backtrace true;
+  Logs.set_level (Some Info);
+  Logs.set_reporter (Logs_fmt.reporter ());
+  ignore (Lwt_main.run (main ()))
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -732,3 +732,317 @@ module Response = struct
   let status t = t.status
   let flush t = t.flush
 end
+
+module Parser = struct
+  let[@inline always] is_tchar = function
+    | '0' .. '9'
+    | 'a' .. 'z'
+    | 'A' .. 'Z'
+    | '!' | '#' | '$' | '%' | '&' | '\'' | '*' | '+' | '-' | '.' | '^' | '_'
+    | '`' | '|' | '~' ->
+        true
+    | _ -> false
+
+  module Source = struct
+    type t = { buffer : string; mutable pos : int; upper_bound : int }
+
+    let of_bytes ~pos ?len buffer =
+      let buf_len = String.length buffer in
+      if pos < 0 || pos > buf_len then
+        invalid_arg
+          (Printf.sprintf
+             "Http_parser.Source.of_bigstring: Invalid offset %d. Buffer \
+              length: %d"
+             pos buf_len);
+      let len = Option.value len ~default:(buf_len - pos) in
+      if len < 0 || pos + len > buf_len then
+        invalid_arg
+          (Printf.sprintf
+             "Http_parser.Source.of_bigstring: Invalid len %d. offset: %d, \
+              buffer_length: %d, requested_length: %d"
+             len pos buf_len (pos + len));
+      { buffer; pos; upper_bound = pos + len }
+
+    let[@inline always] get_unsafe t idx =
+      String.unsafe_get t.buffer (t.pos + idx)
+
+    let[@inline always] get t idx =
+      if idx < 0 || t.pos + idx >= t.upper_bound then
+        invalid_arg "Http_parser.Source.get: Index out of bounds";
+      String.unsafe_get t.buffer (t.pos + idx)
+
+    let[@inline always] advance_unsafe t count = t.pos <- t.pos + count
+
+    let[@inline always] advance t count =
+      if count < 0 || t.pos + count > t.upper_bound then
+        invalid_arg
+          (Printf.sprintf
+             "Http_parser.Source.advance: Index out of bounds. Requested \
+              count: %d"
+             count);
+      t.pos <- t.pos + count
+
+    let[@inline always] length t = t.upper_bound - t.pos
+    let[@inline always] is_empty t = t.pos = t.upper_bound
+
+    let[@inline always] to_string t ~pos ~len =
+      if
+        pos < 0
+        || t.pos + pos >= t.upper_bound
+        || len < 0
+        || t.pos + pos + len > t.upper_bound
+      then
+        invalid_arg
+          (Format.asprintf
+             "Http_parser.Source.substring: Index out of bounds., Requested \
+              off: %d, len: %d"
+             pos len);
+      String.sub t.buffer (t.pos + pos) len
+
+    let[@inline always] is_space = function
+      | ' ' | '\012' | '\n' | '\r' | '\t' -> true
+      | _ -> false
+
+    let[@inline always] to_string_trim t ~pos ~len =
+      if
+        pos < 0
+        || t.pos + pos >= t.upper_bound
+        || len < 0
+        || t.pos + pos + len > t.upper_bound
+      then
+        invalid_arg
+          (Format.asprintf
+             "Http_parser.Source.substring: Index out of bounds., Requested \
+              off: %d, len: %d"
+             pos len);
+      let last = ref (t.pos + len - 1) in
+      let pos = ref (t.pos + pos) in
+      while is_space (String.unsafe_get t.buffer !pos) do
+        incr pos
+      done;
+      while is_space (String.unsafe_get t.buffer !last) do
+        decr last
+      done;
+      let len = !last - !pos + 1 in
+      String.sub t.buffer !pos len
+
+    let rec index_rec t ch idx len =
+      if idx = len then -1
+      else if String.unsafe_get t.buffer (t.pos + idx) = ch then idx
+      else index_rec t ch (idx + 1) len
+
+    let index t ch = index_rec t ch 0 (length t)
+
+    let for_all_is_tchar t ~pos ~len =
+      if
+        pos < 0
+        || t.pos + pos >= t.upper_bound
+        || len < 0
+        || t.pos + pos + len > t.upper_bound
+      then
+        invalid_arg
+          (Format.asprintf
+             "Http_parser.Source.substring: Index out of bounds. Requested \
+              off: %d, len: %d"
+             pos len);
+      let pos = ref (t.pos + pos) in
+      let len = t.pos + len in
+      while !pos < len && is_tchar (String.unsafe_get t.buffer !pos) do
+        incr pos
+      done;
+      !pos = len
+
+    let unsafe_memcmp t pos str =
+      let rec loop t pos str len =
+        if pos = len then true
+        else
+          Char.equal (get_unsafe t pos) (String.unsafe_get str pos)
+          && loop t (pos + 1) str len
+      in
+      loop t pos str (String.length str)
+  end
+
+  exception Msg of string
+  exception Partial
+
+  let string str source =
+    let len = String.length str in
+    if Source.length source < len then raise_notrace Partial
+    else if Source.unsafe_memcmp source 0 str then Source.advance source len
+    else raise_notrace (Msg (Printf.sprintf "Could not match: %S" str))
+
+  let any_char source =
+    if Source.is_empty source then raise_notrace Partial
+    else
+      let c = Source.get_unsafe source 0 in
+      Source.advance_unsafe source 1;
+      c
+
+  let eol = string "\r\n"
+
+  (* token = 1*tchar tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^"
+     / "_" / "`" / "|" / "~" / DIGIT / ALPHA ; any VCHAR, except delimiters *)
+
+  let token source =
+    let pos = Source.index source ' ' in
+    if pos = -1 then raise_notrace Partial
+    else
+      let res = Source.to_string source ~pos:0 ~len:pos in
+      Source.advance source (pos + 1);
+      res
+
+  let meth source =
+    let token = token source in
+    Method.of_string token
+
+  let version_source source =
+    string "HTTP/1." source;
+    any_char source
+
+  let version source =
+    let ch = version_source source in
+    match ch with
+    | '1' -> `HTTP_1_1
+    | '0' -> `HTTP_1_0
+    | _ -> raise_notrace (Msg "Invalid http version")
+
+  let header source =
+    let pos = Source.index source ':' in
+    if pos = -1 then raise_notrace Partial
+    else if pos = 0 then raise_notrace (Msg "Invalid header: Empty header key")
+    else if Source.for_all_is_tchar source ~pos:0 ~len:pos then (
+      let key = Source.to_string source ~pos:0 ~len:pos in
+      Source.advance_unsafe source (pos + 1);
+      while
+        (not (Source.is_empty source)) && Source.get_unsafe source 0 = ' '
+      do
+        Source.advance_unsafe source 1
+      done;
+      let pos = Source.index source '\r' in
+      if pos = -1 then raise_notrace Partial
+      else
+        let v = Source.to_string_trim source ~pos:0 ~len:pos in
+        Source.advance_unsafe source pos;
+        (key, v))
+    else raise_notrace (Msg "Invalid Header Key")
+
+  let headers =
+    let rec loop source acc =
+      if (not (Source.is_empty source)) && Source.get_unsafe source 0 = '\r'
+      then (
+        eol source;
+        Header.of_list (List.rev acc))
+      else
+        let v = header source in
+        eol source;
+        loop source (v :: acc)
+    in
+    fun source -> loop source []
+
+  let chunk_length source =
+    let ( lsl ) = Int64.shift_left in
+    let ( lor ) = Int64.logor in
+    let length = ref 0L in
+    let stop = ref false in
+    let state = ref `Ok in
+    let count = ref 0 in
+    let processing_chunk = ref true in
+    let in_chunk_extension = ref false in
+    while not !stop do
+      if Source.is_empty source then (
+        stop := true;
+        state := `Partial)
+      else if !count = 16 && not !in_chunk_extension then (
+        stop := true;
+        state := `Chunk_too_big)
+      else
+        let ch = Source.get source 0 in
+        Source.advance source 1;
+        incr count;
+        match ch with
+        | '0' .. '9' as ch when !processing_chunk ->
+            let curr = Int64.of_int (Char.code ch - Char.code '0') in
+            length := (!length lsl 4) lor curr
+        | 'a' .. 'f' as ch when !processing_chunk ->
+            let curr = Int64.of_int (Char.code ch - Char.code 'a' + 10) in
+            length := (!length lsl 4) lor curr
+        | 'A' .. 'F' as ch when !processing_chunk ->
+            let curr = Int64.of_int (Char.code ch - Char.code 'A' + 10) in
+            length := (!length lsl 4) lor curr
+        | ';' when not !in_chunk_extension ->
+            in_chunk_extension := true;
+            processing_chunk := false
+        | ('\t' | ' ') when !processing_chunk -> processing_chunk := false
+        | ('\t' | ' ') when (not !in_chunk_extension) && not !processing_chunk
+          ->
+            ()
+        | '\r' ->
+            if Source.is_empty source then (
+              stop := true;
+              state := `Partial)
+            else if Source.get source 0 = '\n' then (
+              Source.advance source 1;
+              stop := true)
+            else (
+              stop := true;
+              state := `Expected_newline)
+        | _ when !in_chunk_extension ->
+            (* Chunk extensions aren't very common, see:
+               https://tools.ietf.org/html/rfc7230#section-4.1.1 Chunk extensions aren't
+               pre-defined, and they are specific to invidividual connections. In the future
+               we might surface these to the user somehow, but for now we will ignore any
+               extensions. TODO: Should there be any limit on the size of chunk extensions we
+               parse? We might want to error if a request contains really large chunk
+               extensions. *)
+            ()
+        | ch ->
+            stop := true;
+            state := `Invalid_char ch
+    done;
+    match !state with
+    | `Ok -> !length
+    | `Partial -> raise_notrace Partial
+    | `Expected_newline -> raise_notrace (Msg "Expected_newline")
+    | `Chunk_too_big -> raise_notrace (Msg "Chunk size is too large")
+    | `Invalid_char ch ->
+        raise_notrace
+          (Msg (Printf.sprintf "Invalid chunk_length character %C" ch))
+
+  let version source =
+    let version = version source in
+    eol source;
+    version
+
+  let[@warning "-3"] request source =
+    let meth = meth source in
+    let path = token source in
+    let version = version source in
+    let headers = headers source in
+    {
+      Request.headers;
+      meth;
+      scheme = None;
+      resource = path;
+      version;
+      encoding = Header.get_transfer_encoding headers;
+    }
+
+  type error = Partial | Msg of string
+
+  let run_parser ?pos ?len buf p =
+    let pos = Option.value pos ~default:0 in
+    let source = Source.of_bytes ~pos ?len buf in
+    match p source with
+    | exception Partial -> Error Partial
+    | exception Msg m -> Error (Msg m)
+    | v ->
+        let consumed = source.pos - pos in
+        Ok (v, consumed)
+
+  let parse_request ?pos ?len buf = run_parser ?pos ?len buf request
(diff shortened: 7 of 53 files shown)


rgrinberg:
> As an experiment I tried to run cohttp-lwt-unix server that doesn't use conduit + Lwt_io for the read path, and instead just uses Lwt_unix.file_descr paired with the buffer that's part of this PR. With that the server was able to handle 100_000. RPS with average latencies of 2ms with tail latencies of 7.6ms - link here

If I understand correctly, the slow down is because Lwt_io allocates an additional bigarray?

cohttp/src/s.ml
```diff
@@ -39,6 +39,11 @@ module type IO = sig
   type conn
   (** [conn] represents the underlying network flow *)
 
+  val refill : ic -> [ `Ok | `Eof ] t
+
+  val with_input_buffer :
+    ic -> f:(string -> pos:int -> len:int -> consume:(int -> unit) -> 'a) -> 'a
+
```
anuragsoni, line 46: I wanted to have an api that lets me see an immutable view within the reader's buffer, and use it to feed data to the parser. The request parser itself always runs in one shot mode and doesn't hold any state within it so this should be pretty safe. We'll either have a successfully parsed result, in which case the parser will inform us how many bytes were consumed (which will be used to inform the reader about the same), or it will respond with a message saying we need more data, or if we had an error while parsing. In both these cases none of the bytes will be consumed.

For dealing with partial reads, the intention is that we will just use `refill` again to fetch more bytes and if that succeeds we'll attempt to parse the request again.

rgrinberg:
> Httpaf_lwt_unix has been used as the benchmark here and it can easily server 100_000 rps with an average latency of 2.5ms with a tail of 9ms

So what was the limit for Httpaf here for reference? In other words, how much further can it push past 100k with decent latency

http-parser/src/http_parser.mli
```diff
@@ -0,0 +1,24 @@
...
+type chunk_parser_result =
+  | Chunk_complete of string
+  | Done
+  | Partial_chunk of string * int
+
+val parse_request :
+  ?pos:int -> ?len:int -> string -> (Http.Request.t * int, error) result
+
+val parse_chunk_length :
+  ?pos:int -> ?len:int -> string -> (int * int, error) result
```
anuragsoni, line 17: I have used  `int` for Chunk lengths as the initial work that led to this parser made an assumption that we'll always run on 64bit systems. With that assumption the `int` would be enough to represent the largest chunk size that's allowed by the parser.

http-parser/src/http_parser.mli
```diff
@@ -0,0 +1,24 @@
...
+
+val parse_chunk :
+  ?pos:int ->
+  ?len:int ->
+  string ->
+  chunk_kind ->
+  (chunk_parser_result * int, error) result
```
anuragsoni, line 24: My intention was to not have to use `parse_chunk_length` manually, and instead go via the `parse_chunk` which helps orchestrate threading through the chunk body state

anuragsoni: I'll also note that I don't make use of any of the chunked parsing code in this PR and I'm limiting this diff to just the request (upto the headers) parsing for now

cohttp-lwt-unix/src/input_channel.ml
```diff
@@ -0,0 +1,24 @@
+open Lwt.Infix
+
+type t = { buf : Bytebuffer.t; chan : Lwt_io.input_channel }
```
anuragsoni, line 3: Adding a buffer layer alongside `Lwt_io` as it helps in driving the core IO layer that needs access to the internal view within the buffer to feed data to the new parser.

rgrinberg:
Is the intention to make http-parser a public API? If that's the case, a package is appropriate. But if it's not, we need to find a new place for it to live (either in http or cohttp).

cohttp/src/s.ml
rgrinberg: Why is consume a function instead of a something that we return along with `'a`?

anuragsoni:
> So what was the limit for Httpaf here for reference? In other words, how much further can it push past 100k with decent latency

On my laptop its able to go up to 125k with decent latency. With a run of [150k rps](https://gist.github.com/anuragsoni/eb095a7f4c3aec1f4ba65b9866d9f9d3#file-7_httpaf_lwt_unix_150_000) it degraded into average latency 2.4 seconds

cohttp-lwt-unix/src/input_channel.ml
anuragsoni: Before this is considered ready, we might want a way to grow the buffer too I suppose (with an upper bound beyond which its considered an error maybe?)

cohttp/src/s.ml
anuragsoni: Hmm, no particular reason. I think `'a * int` as a return type should also work here. I'll switch to that.

cohttp-lwt-unix/src/input_channel.ml
rgrinberg: Could you expand on this a bit more. When does the buffer need to grow?

anuragsoni: I don't think it needs to grow as such as long we pick a value that we feel is reasonable to parse the request + headers. But I noticed that the body decoding implementation in cohttp references that it'd like to read up to a max of 32k bytes at a time when reading chunked bodies https://github.com/mirage/ocaml-cohttp/blob/efe05994fcab28c321caae5fb3ece599738979ce/cohttp/src/transfer_io.ml#L52 So I figured if we want to start with a smaller buffer than that, we'll need to grow it if we still want to eventually read 32k bytes when we get to bodies. Or we could just pick a large enough buffer to start with and never touch it again.

anuragsoni:
> Is the intention to make http-parser a public API? If that's the case, a package is appropriate. But if it's not, we need to find a new place for it to live (either in http or cohttp).

I am not fully sure about this. I think I'd be fine with both a public package, or something that lives within either `http` or `cohttp`.

cohttp/src/s.ml
anuragsoni: This has been switched to using a `'a * int` as the return value instead of using `consume` as a function

examples/bench/lwt_unix_server.ml
```diff
@@ -0,0 +1,43 @@
+open Cohttp_lwt_unix
+
+let text =
```
anuragsoni, line 3: This is the lwt server I'm using to benchmark cohttp

cohttp-lwt-unix/src/input_channel.ml
rgrinberg: 32K sounds like a reasonable maximum. I suppose we should determine the default buffer size experimentally and let the user customize it if need be.

rgrinberg:
> I am not fully sure about this. I think I'd be fine with both a public package, or something that lives within either http or cohttp.

Do you think the current API is stable? If not, then we should probably include it in Http under a Private module. Once it stabilizes, we can promote it into a separate package, library, etc.

cohttp-async/src/io.ml
```diff
@@ -102,6 +102,8 @@ module IO = struct
         Writer.write oc buf;
         return ())
 
+  let refill _ = failwith "Not implemented yet"
+  let with_input_buffer _ = failwith "not implemented yet"
```
rgrinberg, line 106: I'm thinking about how to implement this, and I think it would actually make things easier if `Byte_buffer` was Bigstring based or at least functorized on the buffer type. Reason being is that I don't see how to implement these functions in an efficient manner for Async's Reader for example. I'd imagine we want to use `read_one_chunk_at_a_time` to implement `with_input_buffer`. I understand how bigstrings don't are a wash performance wise, but it helps keep compatibility with existing implementations. Perhaps we could functorize Byte_buffer to make it possible to continue experimenting in Shuttle/Fiber without normal Bytes.

cohttp-lwt-unix/src/input_channel.ml
mseri: For reference, I went back through the history to find the PR that first introduced the limit: https://github.com/mirage/ocaml-cohttp/pull/330 Looks like at that stage it was arbitrary. I think it would be nice to make it configurable, something similar to `Lwt_io.set_default_buffer_size io_buf`

cohttp-async/src/io.ml
mseri: I agree but if we want to keep it simple we could check how big of a performace hit would be to just use bigstring

anuragsoni: Moving to bigstrings will also mean moving the parser to bigstrings. Otherwise we pay the price of allocating more strings?

I was thinking of implementing the async backend in a similar manner as the lwt where there is a byte buffer paired with the reader.

That said, now that I think about it there is going to be some breaking changes in the sense that people using the expert response in servers will need to adapt to the switch to the new input type in the IO backends.

mseri: > Moving to bigstrings will also mean moving the parser to bigstrings. Otherwise we pay the price of allocating more strings?

Yes but that is very localized to the `Source` module, no? In any case it was just a curiosity, and as you mentioned you also already have an idea to make this a non-issue

> That said, now that I think about it there is going to be some breaking changes in the sense that people using the expert response in servers will need to adapt to the switch to the new input type in the IO backends.

That is the last of the problems. We are allowed to break things in cohttp now, as long as we can justify it. Imho the speedup and latency numbers shown here would justify the breakage

avsm: `string` is the way to go in OCaml 5.0, since it's both non-moving and allocated via malloc (all large ocaml objects are mallocated in multicore)

anuragsoni: @avsm That's good to know! With that info I think keeping the API within cohttp to use strings/bytes seems like the way to go. 

> @mseri  That is the last of the problems. We are allowed to break things in cohttp now, as long as we can justify it. Imho the speedup and latency numbers shown here would justify the breakage

👍🏼 That sounds good. And I believe if the libraries using the expert response reference the IO layers by using the IO modules within `cohttp-async`, `cohttp-lwt` etc then we might not even break anything as long as the IO signature stays compatible!

cohttp-async/src/bytebuffer.ml
```diff
@@ -0,0 +1,87 @@
+open Core
```
anuragsoni, line 1: I need to think about how to remove some of this duplication as this is very similar to the module proposed for the lwt backend

anuragsoni:
I pushed an async update as well. But unlike the lwt backend I see no difference in perf between the old implementation and the new implementation 😞 I'll spend more time on this to see what's happening.

anuragsoni:
I reworked the async server loop and now I can see improvement when compared to the existing cohttp implementation. With the older cohttp-async server when I attempt to target 80_000 RPS with 1000 active clients I have an average latency of 8 seconds with tails of 12 seconds. With the new implementation the average latency drops to 2.3 seconds with tails of 3.5 seconds. This is still quite a way behind the lwt implementation.

That said, I think on the async backend the current limiting factor might be the conduit setup. I wrote another async backend for cohttp using shuttle https://github.com/anuragsoni/shuttle/blob/b296ac77f590735ea56d94ef19ae503251dcd317/shuttle_cohttp/shuttle_cohttp.ml and with that backend the average latency when serving 80_000 RPS is 6.9ms with tails of 16ms . These numbers are closer to what the new lwt backend gets here.

All of the async tests were run on macOS where it uses the select backend (the lwt tests were with libel so it probably picks up kqueue on macOS?), and the async backend might perform a little better when run on linux where it has access to epoll.

mseri:
I think you can use the libuv backend now on macos. It is experimental but in my tests it has been working pretty reliably

anuragsoni:
> I think you can use the libuv backend now on macos. It is experimental but in my tests it has been working pretty reliably

I will try that! That said the libev backend for lwt seems to be working well so I'm not expecting much difference between that and libuv since the current state of `libuv` support in lwt is mostly for polling file descriptors for IO readiness only (same as libev). I'd be more interested in seeing test runs of async on a linux system so we can try the epoll backend for it :)

cohttp/test/test_request.ml
```diff
@@ -218,7 +218,7 @@ let parse_request_uri_host_query_no_slash _ =
   parse_request_uri_ r bad_request "parse_request_uri_host_query_no_slash"
 
 let parse_request_connect _ =
-  let r = "CONNECT vpn.example.net:443 HTTP/1.1\r\n" in
+  let r = "CONNECT vpn.example.net:443 HTTP/1.1\r\n\r\n" in
```
mseri, line 221: Shouldn't CONNECT only end with a carriage return?

cohttp-async/src/io.ml
rgrinberg: > That is the last of the problems. We are allowed to break things in cohttp now, as long as we can justify it. Imho the speedup and latency numbers shown here would justify the breakage

My point is not to avoid breakage, but the observation that we don't have a convenient replacement to offer to the user. Users have a ton of code written using `Writer.t` and `Lwt_io.t` and removing support for these fundamentals is a loss feature wise.

But I think we can still keep normal bytes around. We should just make it possible for users to use Lwt_io/Writer on top even if it means copying. Hopefully, the cost will only be there if these streaming options are used.

mseri: Fair point!

examples/bench/async_server.ml
```diff
@@ -0,0 +1,56 @@
...
+   it pop down a large rabbit-hole under the hedge. In another moment down \
+   went Alice after it, never once considering how in the world she was to get \
+   out again. The rabbit-hole went straight on like a tunnel for some way, and \
+   then dipped suddenly down, so suddenly that Alice had not a moment to think \
+   about stopping herself before she found herself falling down a very deep \
+   well. Either the well was very deep, or she fell very slowly, for she had \
+   plenty of time as she went down to look about her and to wonder what was \
+   going to happen next. First, she tried to look down and make out what she \
+   was coming to, but it was too dark to see anything; then she looked at the \
+   sides of the well, and noticed that they were filled with cupboards......"
```
rgrinberg, line 33: Can we replace this with `String.make` with an equivalent length instead?

cohttp-async/src/bytebuffer.ml
rgrinberg: What are the differences preventing you from making this generic? It seems like you just need a function over a monad + reader.

cohttp/test/test_request.ml
anuragsoni: I will check again, but I believe the full request does need to finish on an `\r\n` after the first CRLF that indicates the end of the first line of the request.

cohttp-async/src/io.ml
anuragsoni: We can try an implementation on top of reader and Lwt_io without an additional buffer. With Async.Reader we will have a way to peek inside the reader's buffer and forward a slice to the parser. I think Lwt_io has something similar via `direct_access`. But I'm not sure this will work for mirage as the mirage-channel api doesn't offer something similar. So we'll need buffering at-least for that backend.

cohttp-async/src/bytebuffer.ml
anuragsoni: no difference. This can be moved to something that's instantiated using the IO functor (like we do for other modules like request/response)

cohttp/test/test_request.ml
mseri: No, [you are right](https://datatracker.ietf.org/doc/html/rfc7230#section-3), it is the request line (with ends with a CRLF) + CRLF

anuragsoni: I'm not sure how the test passed before. Maybe it was only testing the parsing of the first line. The new parser always attempts to consume the first line + all headers + CRLF

cohttp-async/src/io.ml
anuragsoni: We can probably deal with this in another manner by pushing the buffer to the core `cohttp` library, and then the refill becomes a way for the IO layers to fill the buffer managed by cohttp. That will let us keep the IO types for the IO layers the same as they are today, while still benefiting from the buffering we need to drive the parser.

anuragsoni: I'm not a huge fan of moving a buffer to the core library as that removes one benefit for libraries like shuttle that could otherwise drive the whole cohttp server machinery with just the 1 buffer that's used by its channel implementation.

rgrinberg: > That will let us keep the IO types for the IO layers the same as they are today, while still benefiting from the buffering we need to drive the parser.

Note that we don't necessarily need to retain the same IO types. For example, if we could just create Reader.t/Writer.t on the fly when the user provides Expert that would be good enough. Though I'm not sure if it's at all possible.

rgrinberg: > I'm not a huge fan of moving a buffer to the core library as that removes one benefit for libraries like shuttle that could otherwise drive the whole cohttp server machinery with just the 1 buffer that's used by its channel implementation.

Yeah, it's not worth it. If we must break the current api and make Reader/Writer impossible to use, then so be it.

http/src/http.ml
```diff
@@ -732,3 +732,355 @@ module Response = struct
...
+              ()
+          | '\r' ->
+              if Source.is_empty source then (
+                stop := true;
+                state := `Partial)
+              else if Source.get source 0 = '\n' then (
+                Source.advance source 1;
+                stop := true)
+              else (
+                stop := true;
```
rgrinberg, line 987: I think we can make `stop := true` outside of the `if`.

cohttp-async/src/io.ml
anuragsoni: `f we could just create Reader.t/Writer.t on the fly`
This should be doable. There will be some awkwardness for sure, but we can definitely get something working. We only need to recreate the reader since I haven't switched the writer in the existing implementation. pseudo code for how I'd do this in async

```ocaml
let reader_of_in_channel t =
   let unconsumed_payload = Bytebuffer.unconsumed_data t.buf in
   let reader_pipe = Reader.pipe t.reader in
   Unix.pipe ~info:(Info.create ...)
  >>| fun (`Reader rd, `Writer wr) ->
        let new_reader = Reader.create rd in
        let writer = Writer.create wr in
       Writer.write writer unconsumed_payload;
        don't_wait_for (
           Writer.transfer writer reader_pipe ~stop:(Reader.close_finished new_reader) (fun s -> Writer.write writer s)
           >>= fun () -> Writer.close writer );
      new_reader
```

anuragsoni:
@rgrinberg @mseri I should've asked this before I rebased. How would you like me to sync with new changes from the main branch? I'm happy to rebase but if merge commits will make it easier to review new changes I can switch to that for new updates.

rgrinberg:
Rebase as much as you want.

Rudi.

mseri:
I agree!

mseri:
Is this expected?

``` test_sanity alias cohttp-lwt-unix/test/runtest (exit 1)
[ERROR][cohttp.lwt.server]: Unhandled exception: Lwt_io.Channel_closed("input")
.....E.
==============================================================================
Error: Cohttp Server Test:5:expert response

(Failure "Test produced 1 log messages at level >= warn")
------------------------------------------------------------------------------
Ran: 7 tests in: 0.04 seconds.
FAILED: Cases: 7 Tried: 7 Errors: 1 Failures: 0 Skip:  0 Todo: 0 Timeouts: 0.
```

I think it happens when a consumer reads from a channel that has already been exhausted or that has been closed for other reasons (timeouts?)

mseri:
When [expert mode was introduced](https://github.com/mirage/ocaml-cohttp/pull/647) this was handled by https://github.com/mirage/ocaml-cohttp/commit/fee808aa5c652eab7e4948a6c130883481afaa35 (See the commentary on https://github.com/mirage/ocaml-cohttp/pull/647#discussion_r234760672)

cohttp-lwt-unix/src/bytebuffer.ml
```diff
@@ -0,0 +1,105 @@
...
+let length t = t.pos_fill - t.pos_read
+
+let drop t len =
+  if len < 0 || len > length t then
+    invalid_arg "Bytebuffer.drop: Index out of bounds";
+  t.pos_read <- t.pos_read + len
+
+let refill t reader =
+  compact t;
+  Lwt_io.read_into reader t.buf t.pos_fill (Bytes.length t.buf - t.pos_fill)
```
mseri, line 50:
```suggestion
  Lwt.catch
    (fun () ->
      Lwt_io.read_into reader t.buf t.pos_fill (Bytes.length t.buf - t.pos_fill)
      >|= fun count ->
      if count > 0 then (
        t.pos_fill <- t.pos_fill + count;
        `Ok)
      else `Eof)
    (function Lwt_io.Channel_closed _ -> Lwt.return `Eof | exn -> raise exn)
```
or
```suggestion
  Lwt.catch
    (fun () ->
      Lwt_io.read_into reader t.buf t.pos_fill (Bytes.length t.buf - t.pos_fill))
    (function Lwt_io.Channel_closed _ -> Lwt.return 0 | exn -> raise exn)
  >|= fun count ->
  if count > 0 then (
    t.pos_fill <- t.pos_fill + count;
    `Ok)
      else `Eof
```
both fix it in my tests

anuragsoni: Thanks! I missed that the channels can be closed in the expert response flow!

anuragsoni:
The CI failure on `ocaml-ci` is because of a test for http parser. I was using `int` to represent chunk length, but one of the test uses a value that is too large for `int` when running on 32 bit systems.

rgrinberg:
Do you think it's acceptable to change to int64?

anuragsoni:
> Do you think it's acceptable to change to int64?

Yes. Using int64 for chunk length should be fine.

anuragsoni:
I factored out the buffer implementation from cohttp-lwt-unix, and I now have the mirage backend compiling with the new parser implementation. 

The mirage backend is the piece I'm least confident about as I don't have much experience with it, and I don't see any existing tests for it within the repository. 

For the Async and lwt-unix backend I think I have an idea about how we can still keep the same public api by ensuring that the callback needed for the `Expert` response still uses Reader.t or Lwt_io.t (see comment https://github.com/mirage/ocaml-cohttp/pull/819#discussion_r780528692), but I'm not sure how to achieve the same for the mirage backend.

rgrinberg:
Thanks for rewriting half of cohttp! I'll merge it now and we can solve the remaining issues in separate PR's.

@samoht @dinosaure @hannesm could you have a look/optimize the mirage backend? The mirage should also be able to reap the performance benefits.

anuragsoni:
I did a quick test with cohttp mirage today. I used the macOS backend as I don't have an easy access to a linux machine at the moment.

`wrk2 -t2 -c1000 -R 80000 --latency "http://localhost:8080" -d 30s`
<details>
<summary>wrk bench against cohttp-mirage 5.0.0</summary>

```
Initialised 2 threads in 0 ms.
Running 30s test @ http://localhost:8080
  2 threads and 1000 connections
  Thread calibration: mean lat.: 1707.276ms, rate sampling interval: 6586ms
  Thread calibration: mean lat.: 1707.088ms, rate sampling interval: 6586ms
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     6.78s     1.79s   11.26s    58.17%
    Req/Sec    25.87k     6.93    25.88k    25.00%
  Latency Distribution (HdrHistogram - Recorded Latency)
 50.000%    6.78s 
 75.000%    8.33s 
 90.000%    9.26s 
 99.000%    9.95s 
 99.900%   10.74s 
 99.990%   11.04s 
 99.999%   11.21s 
100.000%   11.27s 

  Detailed Percentile spectrum:
       Value   Percentile   TotalCount 1/(1-Percentile)

    3583.999     0.000000            1         1.00
    4313.087     0.100000        91013         1.11
    4927.487     0.200000       181366         1.25
    5550.079     0.300000       271885         1.43
    6164.479     0.400000       362114         1.67
    6782.975     0.500000       452952         2.00
    7090.175     0.550000       498130         2.22
    7397.375     0.600000       543526         2.50
    7708.671     0.650000       588484         2.86
    8015.871     0.700000       633877         3.33
    8327.167     0.750000       679194         4.00
    8478.719     0.775000       701774         4.44
    8634.367     0.800000       724806         5.00
    8790.015     0.825000       747399         5.71
    8945.663     0.850000       769649         6.67
    9101.311     0.875000       793039         8.00
    9175.039     0.887500       803666         8.89
    9256.959     0.900000       815683        10.00
    9330.687     0.912500       826238        11.43
    9404.415     0.925000       837298        13.33
    9486.335     0.937500       849214        16.00
    9527.295     0.943750       855257        17.78
    9560.063     0.950000       860116        20.00
    9601.023     0.956250       866107        22.86
    9641.983     0.962500       871998        26.67
    9682.943     0.968750       877757        32.00
    9699.327     0.971875       880092        35.56
    9723.903     0.975000       883610        40.00
    9740.287     0.978125       885882        45.71
    9764.863     0.981250       889004        53.33
    9789.439     0.984375       891641        64.00
    9805.823     0.985938       892816        71.11
    9838.591     0.987500       893844        80.00
    9912.319     0.989062       895295        91.43
    9986.047     0.990625       896624       106.67
   10067.967     0.992188       898031       128.00
   10117.119     0.992969       898780       142.22
   10166.271     0.993750       899456       160.00
   10223.615     0.994531       900185       182.86
   10280.959     0.995313       900847       213.33
   10346.495     0.996094       901569       256.00
   10379.263     0.996484       901925       284.44
   10420.223     0.996875       902305       320.00
   10461.183     0.997266       902621       365.71
   10518.527     0.997656       903012       426.67
   10567.679     0.998047       903342       512.00
   10600.447     0.998242       903542       568.89
   10625.023     0.998437       903676       640.00
   10665.983     0.998633       903878       731.43
   10698.751     0.998828       904043       853.33
   10739.711     0.999023       904213      1024.00
   10764.287     0.999121       904318      1137.78
   10788.863     0.999219       904397      1280.00
   10813.439     0.999316       904487      1462.86
   10838.015     0.999414       904574      1706.67
   10862.591     0.999512       904657      2048.00
   10878.975     0.999561       904711      2275.56
   10895.359     0.999609       904756      2560.00
   10903.551     0.999658       904783      2925.71
   10919.935     0.999707       904823      3413.33
   10944.511     0.999756       904879      4096.00
   10952.703     0.999780       904895      4551.11
   10969.087     0.999805       904929      5120.00
   10977.279     0.999829       904945      5851.43
   10985.471     0.999854       904958      6826.67
   11018.239     0.999878       904981      8192.00
   11034.623     0.999890       904994      9102.22
   11051.007     0.999902       905005     10240.00
   11067.391     0.999915       905013     11702.86
   11083.775     0.999927       905027     13653.33
   11100.159     0.999939       905035     16384.00
   11116.543     0.999945       905042     18204.44
   11124.735     0.999951       905047     20480.00
   11132.927     0.999957       905050     23405.71
   11149.311     0.999963       905056     27306.67
   11165.695     0.999969       905063     32768.00
   11173.887     0.999973       905068     36408.89
   11173.887     0.999976       905068     40960.00
   11182.079     0.999979       905071     46811.43
   11190.271     0.999982       905074     54613.33
   11198.463     0.999985       905077     65536.00
   11198.463     0.999986       905077     72817.78
   11198.463     0.999988       905077     81920.00
   11206.655     0.999989       905079     93622.86
   11214.847     0.999991       905080    109226.67
   11231.231     0.999992       905082    131072.00
   11231.231     0.999993       905082    145635.56
   11239.423     0.999994       905083    163840.00
   11247.615     0.999995       905084    187245.71
   11247.615     0.999995       905084    218453.33
   11255.807     0.999996       905086    262144.00
   11255.807     0.999997       905086    291271.11
   11255.807     0.999997       905086    327680.00
   11255.807     0.999997       905086    374491.43
   11255.807     0.999998       905086    436906.67
   11263.999     0.999998       905087    524288.00
   11263.999     0.999998       905087    582542.22
   11263.999     0.999998       905087    655360.00
   11263.999     0.999999       905087    748982.86
   11263.999     0.999999       905087    873813.33
   11272.191     0.999999       905088   1048576.00
   11272.191     1.000000       905088          inf
#[Mean    =     6784.695, StdDeviation   =     1792.427]
#[Max     =    11264.000, Total count    =       905088]
#[Buckets =           27, SubBuckets     =         2048]
----------------------------------------------------------
  1493513 requests in 30.00s, 2.91GB read
  Socket errors: connect 0, read 4294, write 12, timeout 0
Requests/sec:  49781.00
Transfer/sec:     99.41MB
```
</details>

<details>
<summary>wrk bench against cohttp-mirage pinned to commit 06d2127e0fb197c2549038a36a9787fedf9027cc (the new parser) </summary>

```
Initialised 2 threads in 0 ms.
Running 30s test @ http://localhost:8080
  2 threads and 1000 connections
  Thread calibration: mean lat.: 841.765ms, rate sampling interval: 3358ms
  Thread calibration: mean lat.: 841.993ms, rate sampling interval: 3358ms
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.50s   934.19ms   5.17s    57.79%
    Req/Sec    32.60k    44.57    32.66k    40.00%
  Latency Distribution (HdrHistogram - Recorded Latency)
 50.000%    3.50s 
 75.000%    4.31s 
 90.000%    4.80s 
 99.000%    5.10s 
 99.900%    5.14s 
 99.990%    5.16s 
 99.999%    5.17s 
100.000%    5.17s 

  Detailed Percentile spectrum:
       Value   Percentile   TotalCount 1/(1-Percentile)

    1847.295     0.000000            1         1.00
    2217.983     0.100000       114641         1.11
    2535.423     0.200000       228475         1.25
    2854.911     0.300000       342500         1.43
    3178.495     0.400000       456837         1.67
    3500.031     0.500000       570689         2.00
    3661.823     0.550000       627566         2.22
    3823.615     0.600000       684763         2.50
    3987.455     0.650000       742185         2.86
    4151.295     0.700000       799156         3.33
    4313.087     0.750000       855709         4.00
    4395.007     0.775000       884270         4.44
    4476.927     0.800000       913599         5.00
    4558.847     0.825000       942403         5.71
    4640.767     0.850000       971157         6.67
    4722.687     0.875000       999184         8.00
    4763.647     0.887500      1013301         8.89
    4804.607     0.900000      1027709        10.00
    4845.567     0.912500      1041969        11.43
    4886.527     0.925000      1055969        13.33
    4927.487     0.937500      1070119        16.00
    4947.967     0.943750      1077293        17.78
    4968.447     0.950000      1084506        20.00
    4988.927     0.956250      1091739        22.86
    5009.407     0.962500      1098939        26.67
    5029.887     0.968750      1106206        32.00
    5038.079     0.971875      1109132        35.56
    5050.367     0.975000      1113496        40.00
    5058.559     0.978125      1116363        45.71
    5070.847     0.981250      1120601        53.33
    5079.039     0.984375      1123380        64.00
    5087.231     0.985938      1126084        71.11
    5091.327     0.987500      1127401        80.00
    5095.423     0.989062      1128677        91.43
    5103.615     0.990625      1131101       106.67
    5107.711     0.992188      1132259       128.00
    5111.807     0.992969      1133367       142.22
    5115.903     0.993750      1134435       160.00
    5119.999     0.994531      1135452       182.86
    5124.095     0.995313      1136386       213.33
    5128.191     0.996094      1137288       256.00
    5128.191     0.996484      1137288       284.44
    5132.287     0.996875      1138087       320.00
    5132.287     0.997266      1138087       365.71
    5136.383     0.997656      1138792       426.67
    5136.383     0.998047      1138792       512.00
    5140.479     0.998242      1139407       568.89
    5140.479     0.998437      1139407       640.00
    5140.479     0.998633      1139407       731.43
    5144.575     0.998828      1139916       853.33
    5144.575     0.999023      1139916      1024.00
    5144.575     0.999121      1139916      1137.78
    5148.671     0.999219      1140293      1280.00
    5148.671     0.999316      1140293      1462.86
    5148.671     0.999414      1140293      1706.67
    5152.767     0.999512      1140550      2048.00
    5152.767     0.999561      1140550      2275.56
    5152.767     0.999609      1140550      2560.00
    5152.767     0.999658      1140550      2925.71
    5152.767     0.999707      1140550      3413.33
    5156.863     0.999756      1140712      4096.00
    5156.863     0.999780      1140712      4551.11
    5156.863     0.999805      1140712      5120.00
    5156.863     0.999829      1140712      5851.43
    5156.863     0.999854      1140712      6826.67
    5160.959     0.999878      1140802      8192.00
    5160.959     0.999890      1140802      9102.22
    5160.959     0.999902      1140802     10240.00
    5160.959     0.999915      1140802     11702.86
    5160.959     0.999927      1140802     13653.33
    5165.055     0.999939      1140854     16384.00
    5165.055     0.999945      1140854     18204.44
    5165.055     0.999951      1140854     20480.00
    5165.055     0.999957      1140854     23405.71
    5165.055     0.999963      1140854     27306.67
    5165.055     0.999969      1140854     32768.00
    5165.055     0.999973      1140854     36408.89
    5165.055     0.999976      1140854     40960.00
    5169.151     0.999979      1140877     46811.43
    5169.151     0.999982      1140877     54613.33
    5169.151     0.999985      1140877     65536.00
    5169.151     0.999986      1140877     72817.78
    5169.151     0.999988      1140877     81920.00
    5169.151     0.999989      1140877     93622.86
    5169.151     0.999991      1140877    109226.67
    5169.151     0.999992      1140877    131072.00
    5169.151     0.999993      1140877    145635.56
    5169.151     0.999994      1140877    163840.00
    5169.151     0.999995      1140877    187245.71
    5169.151     0.999995      1140877    218453.33
    5169.151     0.999996      1140877    262144.00
    5169.151     0.999997      1140877    291271.11
    5169.151     0.999997      1140877    327680.00
    5169.151     0.999997      1140877    374491.43
    5169.151     0.999998      1140877    436906.67
    5169.151     0.999998      1140877    524288.00
    5173.247     0.999998      1140879    582542.22
    5173.247     1.000000      1140879          inf
#[Mean    =     3503.824, StdDeviation   =      934.188]
#[Max     =     5169.152, Total count    =      1140879]
#[Buckets =           27, SubBuckets     =         2048]
----------------------------------------------------------
  1890298 requests in 30.00s, 3.69GB read
Requests/sec:  63009.22
Transfer/sec:    125.83MB
```
</details>

PR #824, opened 2022-01-07
http.bench: extra test and faster benchmark

mseri:
For the current benchmark, from tests on a recent and an old computer, picking a 1s quota introduces a little variability, but picking 2s is practically indistinguishable from picking 10s.


mseri:
Updated to 5s, on my old desktop I can also measure a noticeable difference for shorter times.

PR #831, opened 2022-01-10
fix: deprecate the flush field

rgrinberg:


mseri:
Why this change? [Here](https://github.com/mirage/ocaml-cohttp/blob/4198a5c1e47dfa6430950b9eba4139e4871c9ac9/cohttp/src/response.mli#L20) it seems that there was a justification for introducing it

mseri:
Thanks!

mseri (approved):

avsm:
I agree with the deprecation, but not the message at all. It would be much more user friendly to add the deprecation with a pointer to what the user should migrate to.  What's someone going to do when they get this message right now in their code base? There's nowhere to migrate to yet.

mseri:
Yes. I had the same comment in a separate deprecation. The plan is to have the replacement for all the deprecations before the new master is ready to be released

rgrinberg:
I opened #833 to address this and improved the message.

rgrinberg:
Leave it unset and use an explicit flush function IIRC

PR #838, opened 2022-01-15
lwt: new server implementation

### LINKED ISSUE #821: Conduit Free Servers
Conduit on the server side of cohttp is a performance bottleneck and a source of complexity (dependencies, larger api surface). On the other hand, it is also rarely useful. The standard way to deploy web servers is behind a proxy that handles SSL termination. Nginx, HAProxy, Envoy are common examples of such proxies. Therefore conduit is mostly useful in environments where self contained servers that do their own SSL termination are necessary. Mirage is the obvious example here.

My proposal is therefore to provide a version of cohttp-async and cohttp-lwt-unix that is free of conduit. In particular, I propose the following two new packages:

* cohttp-server-lwt-unix
* cohttp-server-async

My reasoning is that this option is minimally disruptive to end users. In particular, users that use the existing cohttp-async and cohttp-lwt-unix packages for the http clients will not be disturbed at all. As conduit remains very useful for http clients, there's no justification to break their existing code.

rgrinberg:
Fix #821 

cc @anuragsoni


rgrinberg:
As discussed with Anurag, the proposed interface is a little more primitive on purpose. We allow users to initialize the server how they want so we don't have to keep up with the various server construction options provided by Lwt.

mseri:
I was going to suggest to just move this into cohttp-lwt-unix and use it to implement the server there, exposing a V1 and V2, but this is much lighter dependency-wise so I think it is useful to have them separate.

Can this be used as a dependency to cohttp-lwt-unix to replace part of the implementation there or you think it is not worth for some reason?

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
```diff
@@ -0,0 +1,111 @@
...
+  type 'a t = 'a Lwt.t
+
+  let ( >>= ) = Lwt.bind
+  let return = Lwt.return
+
+  type ic = Input_channel.t
+  type oc = Lwt_io.output_channel
+  type conn = unit
+
+  let wrap_read f ~if_closed =
```
mseri, line 66: Should we just bump up `lwt` requirements and use `is_closed` (also in cohttp-lwt-unix)? It was released in 4.2.0 in 2019. If so, I can open an issue and this can be done on a separate PR.

rgrinberg: Yeah that sounds fine to me. Definitely out of scope for this PR.

rgrinberg:
@anuragsoni could you give this one a bench? Pretty sure it should be similar to how you removed Conduit, but let's just make sure.

anuragsoni:
> could you give this one a bench? Pretty sure it should be similar to how you removed Conduit, but let's just make sure.

I ran wrk benchmarks using the same inputs as benchmarks for the new parser pull request
https://gist.github.com/anuragsoni/765c486aac01aa7d75fd0b3036801ac5 

Benchmarks for the parser pull-request were at https://gist.github.com/anuragsoni/eb095a7f4c3aec1f4ba65b9866d9f9d3

The results here seem to be about the same as the conduit branch which is what I'd expect considering in both scenarios we end up with an `Lwt_io` with the same default buffer size (4096 bytes I believe?)

rgrinberg:
And now there's a simple sanity test as well.

rgrinberg:
@anuragsoni i added a skeleton implementation that that doesn't use Cohttp_lwt (but still uses the existing IO api)

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.mli
```diff
@@ -0,0 +1,28 @@
...
+
+  val request : t -> Http.Request.t
+  val read_body : t -> string Lwt.t
+  val discard_body : t -> unit Lwt.t
+  val respond : t -> Http.Response.t -> Body.t -> unit Lwt.t
+end
+
+type t
+
+val create : ?on_exn:(exn -> unit) -> (Context.t -> unit Lwt.t) -> t
```
rgrinberg, line 25: This kind of callback allows us to send a response and then continue doing something. That's not possible with a `request -> response` style of callback.

rgrinberg:
@anuragsoni I got rid of the existing IO layer for writing in favor of something simpler.

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
```diff
@@ -0,0 +1,300 @@
...
+      | Some s -> s
+      | None -> Encoding.Fixed (Int64.of_int (String.length s))
+    in
+    (encoding, `String s)
+
+  let stream ?(encoding = Encoding.Chunked) f = (encoding, f)
+  let chunk_size = 4096
+
+  let write_chunk oc (sub : Substring.t) =
+    let* () = Lwt_io.write_int oc sub.len in
```
anuragsoni, line 50: The chunk length needs to be written in hexadecimal

anuragsoni:
For reference, this is the lwt program I was using to run a comparison benchmark with the conduit lwt implementation:

```ocaml
open Lwt.Infix

let length = 2053
let text = String.make length 'a'
let headers = Cohttp.Header.of_list [ ("content-length", Int.to_string length) ]

let server_callback ctx =
  Cohttp_server_lwt_unix.Context.discard_body ctx >>= fun () ->
  Cohttp_server_lwt_unix.Context.respond ctx
    (Http.Response.make ~headers ())
    (Cohttp_server_lwt_unix.Body.string text)

let server = Cohttp_server_lwt_unix.create server_callback

let main () =
  let listen_address = Unix.(ADDR_INET (inet_addr_loopback, 8080)) in
  Lwt.async (fun () ->
      Lwt_io.establish_server_with_client_address ~backlog:11_000 listen_address
        (fun _addr ch -> Cohttp_server_lwt_unix.handle_connection server ch)
      >>= fun _server -> Lwt.return_unit);
  let forever, _ = Lwt.wait () in
  forever

let () =
  Printexc.record_backtrace true;
  Logs.set_level (Some Info);
  Logs.set_reporter (Logs_fmt.reporter ());
  ignore (Lwt_main.run (main ()))
```

mseri:
Let’s add this to the benchmarks binaries as well

rgrinberg:
already added

mseri:
Oh, sorry, I had missed it, thanks!

rgrinberg:
It's there now. I forgot to push.

rgrinberg:
Okay, this is is ready. I'll talk to Anurag about what to do about Lwt_io in subsequent work.

cohttp-server-lwt-unix/bench/bench.ml
```diff
@@ -0,0 +1,33 @@
+open Lwt.Syntax
```
mseri, line 1: Can you move this in https://github.com/mirage/ocaml-cohttp/tree/master/bench and update https://github.com/mirage/ocaml-cohttp/blob/master/bench/latency.sh and the dune file to have alsot the latency tests as part of the `cohttp-bench` package?

mseri (approved):

PR #839, opened 2022-01-15
Allow creating cohttp-async servers with existing reader/writers

anuragsoni:
I'm not sure about the naming scheme so I'm happy to have suggestions for that. I put these in a separate module to avoid a clash with the existing create functions. There won't be a change in performance compared to existing conduit based servers, but this will let people use the full set of configuration options within async's Tcp.Server modules which might be desirable.

### DIFF
--- a/cohttp-async/src/cohttp_async.ml
+++ b/cohttp-async/src/cohttp_async.ml
@@ -1,7 +1,8 @@
 module Body = Body
 module Body_raw = Body [@@deprecated "Use Body"]
 module Client = Client
-module Io = Io [@@deprecated "This module is not for public consumption"]
+module Io = Io.IO
+module Input_channel = Input_channel
 module Request = Cohttp.Request [@@deprecated "Use Cohttp.Request directly"]
 module Response = Cohttp.Response [@@deprecated "Use Cohttp.Response directly"]
 module Server = Server
--- a/cohttp-async/src/cohttp_async.ml
+++ b/cohttp-async/src/cohttp_async.ml
@@ -1,8 +1,7 @@
 module Body = Body
 module Body_raw = Body [@@deprecated "Use Body"]
 module Client = Client
-module Io = Io.IO
-module Input_channel = Input_channel
+module Io = Io [@@deprecated "This module is not for public consumption"]
 module Request = Cohttp.Request [@@deprecated "Use Cohttp.Request directly"]
 module Response = Cohttp.Response [@@deprecated "Use Cohttp.Response directly"]
 module Server = Server
--- a/cohttp-async/src/server.mli
+++ b/cohttp-async/src/server.mli
@@ -105,19 +105,30 @@ val create :
   ('address, 'listening_on) t Async_kernel.Deferred.t
 (** Build a HTTP server, based on the [Tcp.Server] interface *)
 
-val create_tcp_handler :
-  'addr ->
-  Async_unix.Reader.t ->
-  Async_unix.Writer.t ->
-  (body:Body.t -> 'addr -> Http.Request.t -> response Async_kernel.Deferred.t) ->
-  unit Async_kernel.Deferred.t
+module Expert : sig
+  val create :
+    (body:Body.t -> 'addr -> Http.Request.t -> response Async_kernel.Deferred.t) ->
+    'addr ->
+    Async_unix.Reader.t ->
+    Async_unix.Writer.t ->
+    unit Async_kernel.Deferred.t
+  (** [create] accepts a user provided cohttp handler, and creates a server
+      callback that works with user provided socket address,
+      [Async_unix.Reader.t] and [Async_unix.Writer.t]. This can be useful if
+      there is a need for more control over how the Reader and Writer get
+      created. *)
 
-val create_expert_tcp_handler :
-  'addr ->
-  Async_unix.Reader.t ->
-  Async_unix.Writer.t ->
-  (body:Body.t ->
-  'addr ->
-  Http.Request.t ->
-  response_action Async_kernel.Deferred.t) ->
-  unit Async_kernel.Deferred.t
+  val create_expert :
+    (body:Body.t ->
+    'addr ->
+    Http.Request.t ->
+    response_action Async_kernel.Deferred.t) ->
+    'addr ->
+    Async_unix.Reader.t ->
+    Async_unix.Writer.t ->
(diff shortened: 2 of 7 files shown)


cohttp-async/src/cohttp_async.ml
```diff
@@ -1,11 +1,8 @@
 module Body = Body
 module Body_raw = Body [@@deprecated "Use Body"]
 module Client = Client
-module Io = Io [@@deprecated "This module is not for public consumption"]
+module Io = Io.IO
```
mseri, line 4: Why removed this deprecation?

cohttp-async/src/cohttp_async.ml
```diff
@@ -1,11 +1,8 @@
...
+module Input_channel = Input_channel
 module Request = Cohttp.Request [@@deprecated "Use Cohttp.Request directly"]
 module Response = Cohttp.Response [@@deprecated "Use Cohttp.Response directly"]
 module Server = Server
-
-module Private = struct
```
mseri, line 9: Why you needed to make this exposed again? It does not seem to make a difference for the implementation and can be still referred to as a private module

cohttp-async/src/cohttp_async.ml
anuragsoni: We forward the new IO type (reader + buffer) in the expert response callback. This IO module will be the only way to perform any read operations there so leaving it deprecated doesn’t seem ideal.

cohttp-async/src/cohttp_async.ml
anuragsoni: Similar to the IO module. This gets passed to the expert response callback so it probably is better to keep it part of the public api?

cohttp-async/src/cohttp_async.ml
anuragsoni: I thought about recreating a new reader and forward that to the expert callback, but that has an edge case where if the user decides to use the expert mode for any reason other than connection upgrades (to websocket etc) then eventually the control will come back to Cohttp and we’ll need to continue processing more requests and would need to recreate the new input channel again.

cohttp-async/src/cohttp_async.ml
anuragsoni: I ended up taking a different approach. The only place we need this is for the expert response within the servers, so I've moved a restricted view of Input_channel within the server module.

cohttp-async/src/cohttp_async.ml
anuragsoni: I brought back the deprecation. I moved to a different approach so instead of making the entire IO module public, I just moved a restricted Input_channel module within the server module as its only useful there when working with expert responses.

anuragsoni:
I'd like to come back to this after a decision has been reached about #842

cohttp-async/src/server.mli
```diff
@@ -1,3 +1,18 @@
...
+  val read : t -> int -> string Async_kernel.Deferred.t
+  val refill : t -> [> `Eof | `Ok ] Async_kernel.Deferred.t
+
+  val with_input_buffer :
+    t -> f:(string -> pos:int -> len:int -> 'a * int) -> 'a
+
+  val is_closed : t -> bool
+  val close : t -> unit Async_kernel.Deferred.t
+  val close_finished : t -> unit Async_kernel.Deferred.t
+end
```
rgrinberg, line 14: Could you add some docs to this module?

anuragsoni: Once #842 is merged, this module will be removed as we won't need to access the Input_channel module outside of cohttp-async again.

rgrinberg: It is now merged.

cohttp-async/src/server.mli
```diff
@@ -102,3 +117,25 @@ val create :
...
 (** Build a HTTP server, based on the [Tcp.Server] interface *)
+
+module V2 : sig
+  val create :
+    'addr ->
+    Async_unix.Reader.t ->
+    Async_unix.Writer.t ->
+    (body:Body.t ->
+    'addr ->
+    Cohttp.Request.t ->
```
rgrinberg, line 128: We prefer Http.Request.t now.

cohttp-async/src/server.mli
anuragsoni: I've synced with the changes from #842 and the diff now only contains the 2 additional functions that will allow using the library with user provided reader/writer

mseri:
Can you also add an entry on the changelog for this and explain the behaviour/api change?

anuragsoni:
> Can you also add an entry on the changelog for this and explain the behaviour/api change?

Good catch! I've updated the changelog

cohttp-async/src/server.mli
```diff
@@ -104,3 +104,20 @@ val create :
...
+
+val create_expert_tcp_handler :
+  'addr ->
+  Async_unix.Reader.t ->
+  Async_unix.Writer.t ->
+  (body:Body.t ->
+  'addr ->
+  Http.Request.t ->
+  response_action Async_kernel.Deferred.t) ->
+  unit Async_kernel.Deferred.t
```
rgrinberg, line 123: I  think our convention elsewhere is to add an `Expert` submodule.

Btw, some minor docstrings for these functions would be appreciated.

anuragsoni: I like the idea of using `Expert` here. I've switched to using that. Also added some basic docs about why these functions exist, and who might find it useful.

cohttp-async/src/server.mli
```diff
@@ -104,3 +104,31 @@ val create :
...
+    Async_unix.Reader.t ->
+    Async_unix.Writer.t ->
+    unit Async_kernel.Deferred.t
+  (** [create] accepts a user provided cohttp handler, and creates a server
+      callback that works with user provided socket address,
+      [Async_unix.Reader.t] and [Async_unix.Writer.t]. This can be useful if
+      there is a need for more control over how the Reader and Writer get
+      created. *)
+
+  val create_expert :
```
rgrinberg, line 121: Sorry to nitpick, but this name is a bit tautological. What about `create_with_expert_response`?

anuragsoni: Good point! I renamed it to `create_with_response_action` as the user can use both a regular response and an expert response here.

rgrinberg:
Squashed & merged thanks.

PR #842, opened 2022-01-16
Switch back cohttp-async expert response to use reader

anuragsoni:
This allows us to remove Input_channel from the public api altogether. One interesting change here is the change in behavior in the server module where we no longer attempt to continue the server loop after the user provided `Expert` handler returns. I think this behavior is okay for use-cases where the expert handler is used for connection upgrades (web sockets etc), and I am not sure if there are other common uses of the expert response_action outside of connection upgrades.

With this change I think we'll maintain compatibility with any existing cohttp-async consumer that was using the expert response (ex: for websockets), so I think the additional complication of performing the conversion to a reader, and the switch in behavior of not attempting to continue the server loop after the expert handler returns is worth it


rgrinberg (approved):
Looks good. Will merge once CI approves.

PR #845, opened 2022-01-17
refactor: remove has_body from io functor

rgrinberg:


rgrinberg:
@anuragsoni @mseri perhaps this function is useful enough to promote to `http`?

anuragsoni:
> perhaps this function is useful enough to promote to `http`?

Promoting it to the `http` library makes sense to me. We have other similar functions there as well (keep-alive checks).

mseri:
I agree, seems to fit well with the api

mseri:
Can you add an entry in the changleog describing the breaking change?

PR #848, opened 2022-01-18
refactor(cohttp): simplify chunk parsing

rgrinberg:


cohttp/src/transfer_io.ml
```diff
@@ -38,10 +38,11 @@ module Make (IO : S.IO) = struct
...
            > more chunk extensions, immediately following the chunk-size
         *)
-        try String.sub chunk_size_hex 0 (String.index chunk_size_hex ';')
-        with _ -> chunk_size_hex
+        match String.index chunk_size_hex ';' with
+        | exception Not_found -> chunk_size_hex
+        | s -> String.sub chunk_size_hex 0 s
       in
-      try Some (Int64.of_string ("0x" ^ hex)) with _ -> None
+      Int64.of_string_opt ("0x" ^ hex)
```
anuragsoni, line 45: We can think about using the new parser within the http library as well:

```ocaml
match Http.Private.parse_chunk_length chunk_size_hex with
| Ok (chunk_length, consumed) -> assert (consumed = String.length chunk_size_hex); chunk_length
| Error Partial -> assert false (* this branch will never be reached here since we feed the full line *)
| Error (Msg msg) -> failwith msg
```

rgrinberg: I gave it a try. It's a big ugly because we need to add the newline at the end of the input.

cohttp/src/transfer_io.ml
```diff
@@ -63,12 +52,17 @@ module Make (IO : S.IO) = struct
...
-            match parse_chunksize chunk_size_hex with
-            | None -> return Done
-            | Some 0L ->
+            match
+              Http.Private.Parser.parse_chunk_length (chunk_size_hex ^ "\r\n")
+            with
+            | Error Partial ->
+                assert false
+                (* this branch will never be reached here since we feed the full line *)
+            | Error (Msg _) -> return Done
```
mseri, line 61: Why we want to ignore any error here? There are things like invalid characters and so on

cohttp/src/transfer_io.ml
anuragsoni: Hmm true. Forgot about the new line.

cohttp/src/transfer_io.ml
mseri: I see that the previous version wasn't failing either, just wondering why errors are ok to be ignored here.

anuragsoni: > just wondering why errors are ok to be ignored here.

Also curious about this. `Done` here makes sense to me with respect to informing the IO layers that there won't be any new data in the stream, but maybe we should also report a  parsing failure instead of silently suppressing it so we can respond with a bad_request etc? In the current form, it seems like we are banking on the next request cycle to result in a parse error as we attempt to parse a http request, but that seems less than ideal.

anuragsoni: I'll also add that i'm not proposing making a behavior change in this PR, as keeping that as a separate PR will make the most sense.

cohttp/src/transfer_io.ml
rgrinberg: It's still better this way I think. We can always tweak `parse_chunk_length` to not eat the newline

cohttp/src/transfer_io.ml
rgrinberg: I don't mind turning this (and other failures) like it into real failures that abort the request. I'm not sure how feasible this is with the current implementation, but feel free to try this.

anuragsoni: > I'm not sure how feasible this is with the current implementation, but feel free to try this.

this will definitely need more experimentation, so keeping it as a separate PR for this is definitely the way to go.

mseri (approved):

PR #851, opened 2022-01-22
Enable js-test

smorimoto:


dune
```diff
@@ -0,0 +1 @@
+(dirs :standard \ node_modules)
```
smorimoto, line 1: We have to tell dune to include `node_modules` since it is excluded from the build directory as the cause of the current problem.

mseri: Super, thanks!

Issue #790, opened 2021-11-10
Request timeout with Cohttp_lwt_unix.Client.get


vouillon:
The documentation explains how to set a timeout while waiting for a response from the server, but I don't see how one can set up a timeout while reading the response body without getting a file descriptor leak.

In particular, as far as I can see, with the following code, the file descriptor is not closed as long as the server does not close its side of the connection, which can well never happen if it stops responding.
```ocaml
  Lwt.pick
    [
      (let%lwt _, body =
         Cohttp_lwt_unix.Client.get
           (Uri.make ~scheme:"http" ~host:"localhost" ~port:8080 ())
       in
       let%lwt () = Cohttp_lwt.Body.drain_body body in
       Lwt.return_unit);
      (let%lwt () = Lwt_unix.sleep 6. in
       Lwt.return_unit);
    ]
```

mseri:
You are correct and I don’t have a good answer, I need to think about it

PR #853, opened 2022-01-25
Explicit pipelining and connection cache

madroach:
I implemented explicit pipelining and a connection cache. The explicit pipeline improves on former `callv` in that `EOF` is handled more gracefully. Requests that were lost in flight will receive proper notice `Retry` that they may safely be retried (whether idempotent or not).
The connection cache is just that. A connection cache. Pipeline depth, parallel connections and keep alive time are configurable. A `no_cache` dummy connection cache is there for use by the convenience `call`, `get`… functions, which won't use the connection cache. I'm not sure whether I put `cohttp/src/conn_cache.ml` at the right place.
I would like to change the uri parameter of `Pipeline.create` and `callv` to something like `Conduit.endp`, because a single endpoint might serve several _virtual_ hostnames.

Feedback is welcome.


mseri:
Thanks a lot! it will take a few days before I can have a proper look, but this sounds really good

cohttp-lwt/src/client.ml
```diff
@@ -12,33 +14,340 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
...
+        | Failed _ -> ()
+        | Connecting _ | Closed -> assert false
+        end;
+        finalise pipeline
+      | `Invalid reason ->
+        let e = Failure ("Cohttp_lwt failed to read response: " ^ reason) in
+        pipeline.conn <- Failed e;
+        finalise pipeline
+
+    (* in shutdown and process I rely on mutex waiters receiving the lock in the
```
mseri, line 148: Is there risk of this changing in the future?

cohttp/src/conn_cache.ml
```diff
@@ -0,0 +1,10 @@
+module type S = sig
```
mseri, line 1: This signature should be in `cohttp-lwt`.

Could we call it `Connection`, as in https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-async/src/client.mli#L30? I understand that here we are not exactly doing the same, but is there a way to make the interfaces similar?

mseri (commented):
A few preliminary comments, before I can look more carefully at the implementation

cohttp-lwt/src/client.ml
```diff
@@ -12,33 +14,340 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
...
+              Pipeline.shutdown pipeline
+          end;
+        res
+    end
+
+    (*
+     * Select from existing connections by hostname, not IP.
+     * One hostname may resolve to multiple IP addresses.
+     * In such a case we want only one connection.
+     * The downside is that some resolver redirections may break.
```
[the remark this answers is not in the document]
madroach: Indeed. If you don't use the connection cache this doesn't matter. But as I wrote initially, I suspect using a `Conduit.endp` instead of an hostname for opening connections is better.
At the moment I do not think there is much to gain from making it configurable. If you have special needs in that you want multiple connections to the very same IP you are probably better of directly using the pipeline.

cohttp-lwt/src/client.ml
madroach: I suspect it won't change, but that's why I added the comment after all.

cohttp/src/conn_cache.ml
madroach: Ah, indeed. But I'd rather rename
- `Pipeline` → `Connection`
- `Pipeline.process` → `Connection.request`
- `Pipeline.shutdown` → `Connection.close`
- `Connection.is_closed` could be implemented by using the already present condition variable.
Last time I looked there was no connection cache in cohttp-async.

I'll do the renaming. But not tonight.

mseri: Great!

cohttp-lwt/src/client.ml
mseri: Do you reckon we have a way of doing it without relying on an implementation detail?

cohttp-lwt/src/client.ml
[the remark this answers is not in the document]
mseri: I don’t have a strong opinion right now, let’s see if somebody else has something to say about it

cohttp-lwt/src/client.ml
madroach: Yes, by making the FIFO `Queue` explicit and not relying on `Lwt_mutex` FiFo behaviour.
Now that I'm thinking about it, there indeed is a problem with the current implementation in that the length of the waiting requests ignores requests in the `Lwt_mutex` queue.

cohttp-lwt/src/s.ml
```diff
@@ -37,6 +37,27 @@ end
...
+    val length : t -> int
+
+    val request :
+      t ->
+      ?body:Body.t ->
+      Cohttp.Request.t ->
+      (Cohttp.Response.t * Body.t) Lwt.t
+  end
+
   module Conn_cache : sig
```
mseri, line 61:
```suggestion
  module Connection_cache : sig
```
A few more characters do not harm.
The module with the signature will need to be documented before this can be merged

cohttp/src/conn_cache.ml
```diff
@@ -5,6 +5,6 @@ module type S = sig
 
   val set_default : t -> unit
 
-  val process : ?conn_cache:t -> ?body:body -> Request.t ->
+  val request : ?conn_cache:t -> ?body:body -> Request.t ->
```
mseri, line 8:
```suggestion
  val request : ?cache:t -> ?body:body -> Request.t ->
```
`t` is already a `Connection_cache.t`

cohttp-lwt/src/client.ml
mseri: Ok, if this is the case, then I think we should make it implementation independent and have an explicit queue

cohttp/src/conn_cache.ml
madroach: OK. Renaming is finished. But I'm reluctant to use the parameter ordering of `Cohttp_async.Client.Connection.request`:
```
val Cohttp_async.Client.Connection.request :
  ?body:Body.t ->
  t ->
  Http.Request.t ->
  (Http.Response.t * Body.t) Async_kernel.Deferred.t
```
But rather keep my ordering:
```
val Cohttp_lwt.Client.Connection.request :
  t ->
  ?body:Body.t ->
  Cohttp.Request.t ->
  (Cohttp.Response.t * Body.t) Lwt.t
```
since you might very well want to do a partial application of just `t` and be able to supply the body later. While there is no use in eliminating the `?body` optional parameter early.

cohttp-lwt/src/client.ml
```diff
@@ -12,33 +14,348 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
...
+            Lwt.return_unit
+          in
+          let connect () =
+            let connection = Connection.connect ~finalise ?ctx addr
+            and timeout = ref Lwt.return_unit in
+            let rec busy () =
+              Lwt.cancel !timeout;
+              if Connection.length connection = 0 then begin
+                timeout :=
+                  Time.sleep_ns Int64.(mul (of_int keep) 1_000_000L) >>= fun () ->
```
mseri, line 284: Do we really need this sleep 0.001s call here?

mseri: Would be nice if we can avoid threading through that. Otherwise we should restructure the code to add explicit timeouts to the calls, which could also solve https://github.com/mirage/ocaml-cohttp/issues/790

cohttp/src/conn_cache.ml
mseri: I don't mind keeping your ordering, I always thought that with labeled parameter it did not matter to be honest. If this is the case it may be worth discussing if we should swap the async call (ping @anuragsoni) as well

anuragsoni: Hmm, I also thought that with optional arguments it did not impact partial application in this case since we do have a `Request` that follows `t`.

cohttp-lwt/src/client.ml
madroach: This is a configureable timeout on idle connections. `keep` is the number of idle milliseconds after which a connection is closed.

Regarding #790 I also had a timeout in my `Connection` module, but removed it. I could add it again (but probably better in a separate pull request). The timeout needs to be tickled after every successful read of a response header or body stream chunk (via `Lwt_stream.map`).

cohttp-lwt/src/client.ml
```diff
@@ -1,7 +1,9 @@
 open Lwt.Infix
 module Header = Cohttp.Header
 
-module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
+exception Retry
+
+module Make (IO : S.IO) (Net : S.Net with module IO = IO) (Time : Mirage_time.S) = struct
```
rgrinberg, line 6: Instead of requiring the Mirage_time signature, what about specifying only the part that we need? This will make it both more flexible to users and make allow us to drop the mirage dependency.

madroach: `Mirage_time` is
```
(** Sleep operations. *)
module type S = sig
  val sleep_ns: int64 -> unit Lwt.t
  (** [sleep_ns n] Block the current thread for [n] nanoseconds, treating
      the [n] unsigned.  *)
end
```
There are no other parts to pick from. `mirage-time-unix` depends on `duration`. This dependency could easily be dropped. What do you propose?

rgrinberg: If the signature is so simple then I definitely suggest inlining it.

dinosaure: Note that this choice becomes incompatible with `functoria` which passes implementation _via_ _functors_. Not sure that we want to break that. We can pass that by a value `~time` (as I did into `dream`) but at the end, a _functor_ is required.

madroach: uh, let me push what I did…

cohttp-lwt/src/client.ml
```diff
@@ -53,21 +53,38 @@ module Make (IO : S.IO) (Net : S.Net with module IO = IO) (Time : Mirage_time.S)
...
+        (length connection)
+        (match connection.state with
+        |Connecting _ -> "Connecting"
+        |Full _ -> "Full"
+        |Closing _ -> "Closing"
+        |Half _ -> "Half"
+        |Closed -> "Closed"
+        |Failed _ -> "Failed")
+
+    let log pipeline s = prerr_endline (to_string pipeline ^ ": " ^ s)
```
mseri, line 75: cohttp-lwt uses logs for logging (it is already linking logs and logs.lwt), we should stick to that

cohttp-lwt/src/client.ml
```diff
@@ -3,7 +3,7 @@ module Header = Cohttp.Header
 
 exception Retry
 
-module Make (IO : S.IO) (Net : S.Net with module IO = IO) (Time : Mirage_time.S) = struct
+module Make (IO : S.IO) (Net : S.Net with module IO = IO) (Sleep : S.Sleep) = struct
```
mseri, line 6: @dinosaure this should be fine for functoria, no?

cohttp-lwt/src/client.ml
madroach: What about 4efebd7 ?
I inlined the signature of `Mirage_time.S` in `Cohttp_lwt` and called it `Sleep`, but kept the dependency on `mirage-time` in `Cohttp_mirage`:
```
module Cohttp_mirage.Client.Make : sig
    (P : Mirage_clock.PCLOCK)
    (T : Mirage_time.S)
    (R : Resolver_mirage.S)
    (S : Conduit_mirage.S) : sig
```
while in unix I do
```
module Sleep = struct
  let sleep_ns ns =
    Lwt_unix.sleep @@
    let open Int64 in
    if ns >= zero
    then to_float ns
    else (shift_right_logical ns 1 |> to_float) *. 2.
end

include Cohttp_lwt.Make_client (Io) (Net) (Sleep)
```

Is this ok with `functoria` @dinosaure ?
Is it ok to keep the dependency of `cohttp-mirage` on `mirage-time` @rgrinberg ?

cohttp-lwt/src/s.ml
```diff
@@ -29,6 +29,10 @@ module type Net = sig
   val close : IO.ic -> IO.oc -> unit
 end
 
+module type Sleep = sig
+  val sleep_ns : int64 -> unit Lwt.t
```
mseri, line 33: If we are functorizing it this way, and we only need milliseconds, should we have `sleep_ms` here instead, so that it is also clear why it is different from mirage-time?

cohttp-lwt/src/client.ml
dinosaure: Hmmhmm, I'm not sure. I mean, for `mirage`/`functoria`, we already have a mapping between a `time` device and the `module type Mirage_time.S`. You can do that but that mostly means that CoHTTP devs must keep in their mind that this interface should **always** be equal to `Mirage_time.S` (and on the other side, mirage devs should be aware that any upgrade of `Mirage_time.S` implies an update of `Cohttp_lwt.S.Sleep`).

If we keep some packages as interfaces (like `mirage-time`, `mirage-block`, etc.), it's to be sure to provide a coherent ecosystem and a simple update of these interfaces should be systematic along our `mirage/*` ecosystem.

Then, the release cycle of `Mirage_time.S` is not a real problem - I mean, the interface did not change for a long time. About dependencies, I still think that it's fine. `mirage-time` is just an interface which can be used across projects and, again, ensure a _certain_ homogeneity across our multi-repositories organization. I'm not sure that anyone want to take this hidden responsibility between what the MirageOS expects and what CoHTTP wants.

cohttp-lwt/src/s.ml
madroach: I thought a bit about the necessary order of magnitude of the idle timeout. I believe sensible values range from days = 2¹⁶ s to about 50μs = 2⁻¹⁴s (roundtrip to localhost). This range barely fits into a 31bit integer, so I'm afraid we'll need a full `int64`. Now with 64bits order of magnitude won't matter that much. I would choose μs, but since `mirage-time` already uses ns, I would just stick with that and also expose that in the `Connection_cache`.

cohttp-lwt/src/client.ml
dinosaure: See this [comment](https://github.com/mirage/ocaml-cohttp/pull/853#discussion_r800618552). Again, I'm not really against that and it can be fine for `cohttp-lwt` but, for `cohttp-mirage`, it must be a `functor` (and it should be `Mirage_time.S`). This is what I did about `dream` which expects a `~time` argument to launch the server and be free about the operating system. Then, `dream-mirage` ensure that it expects a `Mirage_time.S` and we pass the function `Time.sleep_ns` to the runner then.

cohttp-lwt/src/s.ml
mseri: Fair enough, if the range is so large I agree

cohttp-lwt/src/client.ml
madroach: note that the application of the `Cohttp_lwt.Make` functor in `cohttp-mirage/src/client.ml:76` enforces compatibility between `Cohttp_lwt.S.Sleep` and `Mirage_time.S`.

Nevertheless _I_ would prefer using `Mirage_time.S` in `Cohttp_lwt.Client`, adding dependency on `mirage-time`.

cohttp-lwt/src/client.ml
mseri: Thanks for looking into it

At least `cohttp-mirage` will depend on (and use) `mirage-time`, so we have that package as a check for things to remain compatible.

The current code is already functorized, so that should be fine unless I misunderstood something

cohttp-lwt/src/s.ml
madroach: there seems to be a reason, why `Unix.sleep` expects `float`.

mseri: No idea why is like that historically.

Afaik in glibc, `sleep` only works on seconds and you need `nanosleep` for fractions of seconds.
I don't know why in ocaml's `Unix` module this is reflected by `sleep` which takes an integer (in seconds) and `sleepf` which, instead, takes a float to allow for fractions of seconds. Maybe this is to work also on older 32 bits systems?
My guess is that in `Lwt_unix.sleep` they just wanted to remain close to the `Unix` module.

cohttp-lwt-unix/src/client.ml
```diff
@@ -1,3 +1,12 @@
-include Cohttp_lwt.Make_client (Io) (Net)
+module Sleep = struct
+  let sleep_ns ns =
+    Lwt_unix.sleep @@
+    let open Int64 in
+    if ns >= zero
+    then to_float ns
```
mseri, line 6: Here you need to divide by a million, no?

cohttp-lwt-unix/src/client.ml
```diff
@@ -1,3 +1,12 @@
...
+    else (shift_right_logical ns 1 |> to_float) *. 2.
```
mseri, line 7: Do we really need to support this? You were saying that 2¹⁶ should be enough

cohttp-lwt/src/client.ml
madroach: So keep it like it like it is now or may I remove `Cohttp_lwt.S.Sleep` again and use `Mirage_time.S` in both `cohttp-lwt` and `cohttp-mirage` instead ? I would not depend on `mirage-time-unix`, but keep my inlined implementation of it.

mseri: I don't have a strong opinion. Let's wait for @rgrinberg

cohttp-lwt/src/client.ml
madroach: done.

cohttp-lwt-unix/src/client.ml
madroach: oh, indeed. need do divide by 10⁹.

cohttp-lwt/src/client.ml
rgrinberg: I insist on dropping the dependency. If our goal is to maintain compatibility with some signature coming from a mirage package, we can do it in the cohttp-mirage package.

avsm: At a higher level, I don't quite understand why the Client module needs to add a Time dependency at all. Can't the connection cache be built _using_ the Client module, and adding in timeouts and connection pooling there?  This way existing Client consumers aren't affected.

madroach:
That's a great idea, @avsm. I'll do just that and use an inlined signature for sleep_ns. I believe it can even be made independent of lwt. Should I then put it into cohttp so that async may use it, too?

madroach:
After all I did it the other way round and implemented the `Client` functor using my `Connection` and `S.Connection_cache` functors, which now live in separate files. The `Client` functor still takes `Io` and `Net` modules as arguments although the `Io` module is not used. It has shrunk to not much more than convenience functions. This way users may now choose to use only `Connection` or `Connection_cache` to process their requests.

cohttp-lwt/src/client.ml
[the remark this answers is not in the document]
madroach: In my latest changes I indeed changed the type of the connection cache keys to `endp`. This indeed only makes a difference in case you are using a connection cache.

mseri:
I have been sick. I am starting reviewing this but it will take a while. I'd welcome help from anybody interested  (and help with mirage)

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+  type t =
+    { mutable cache : Connection.t list Endp_map.t
+    ; ctx : ctx
+    ; keep : int64
+    ; retry : int
+    ; parallel : int
+    ; depth : int
+    }
+
+  (*
```
mseri, line 67: Should this comment go next to `get_connection`?

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+
+  let create ?(ctx = Net.default_ctx) ?(keep=60_000_000_000L) ?(retry=2) ?(parallel=4) ?(depth=100) () =
+    { cache = Endp_map.empty
+    ; ctx ; keep ; retry ; parallel ; depth
+    }
+
+  let rec get_connection self endp =
+    let finalise connection =
+      let conns = Endp_map.find endp self.cache in
+      let conns = List.filter ((!=) connection) conns in
```
mseri, line 82:
```ocaml
      let conns = Endp_map.find endp self.cache
                  |> List.filter ((!=) connection)
      in
```
feels more in line with the rest of the style but I don't have a very strong opinion about this

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+struct
+  module Net = Connection.Net
+  module IO = Net.IO
+  open IO
+
+  type ctx = Net.ctx
+
+  exception Retry = S.Retry
+
+  module Endp_map = Map.Make (Net.Endp)
```
mseri, line 55: Why you opted for the map in a mutable record field instead of an Hashtable?

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+        | _ :: tl ->
+          search (length + 1) tl
+        | [] -> assert false
+      in
+      match search 0 conns with
+      | shallowest, _ when Connection.length shallowest = 0 ->
+        Lwt.return shallowest
+      | _, length when length < self.parallel ->
+        let connection = create () in
+        self.cache <- self.cache |> Endp_map.add endp (connection :: conns);
```
mseri, line 123:
```suggestion
        self.cache <- Endp_map.add endp (connection :: conns) self.cache;
```
Why the indirection?

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+        Lwt.return connection
+      | shallowest, _ when Connection.length shallowest < self.depth ->
+        Lwt.return shallowest
+      | _ ->
+        Lwt.try_bind
+          (fun () -> conns |> List.map Connection.notify |> Lwt.choose)
```
mseri, line 129:
```suggestion
          (fun () -> Lwt.choose (List.map Connection.notify conns))
```
Why the indirection?

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+          (fun _ -> get_connection self endp)
+          (fun _ -> get_connection self endp)
+
+  let request self ?body req =
+    Net.resolve ~ctx:self.ctx (Cohttp.Request.uri req) >>= fun endp ->
+    let rec request retry =
+      get_connection self endp >>= fun conn ->
+      Lwt.catch (fun () -> Connection.request conn ?body req)
+        begin function
```
mseri, line 138:
```ocaml
        begin function
        | Retry ->
          begin match body with
          | Some `Stream _ -> Lwt.fail Retry
          | None | Some `Empty | Some `String _ | Some `Strings _ ->
            if retry <= 0
            then Lwt.fail Retry
            else request (retry - 1)
          end
        | e -> Lwt.fail e
        end
```

Why `Retry as e`?

mseri (changes requested):
I have a few comments, mostly minors or requests to document the interfaces and add examples and some tests.

cohttp-lwt/src/connection_cache.ml
madroach: I wanted it to read similar `self.cache += endp`. I'm not feeling very attached to it, so I can change it if its stupid.

cohttp-lwt/src/s.ml
```diff
@@ -22,13 +22,62 @@ module type Net = sig
 
   type ctx [@@deriving sexp_of]
 
+  module Endp : Map.OrderedType
+
   val default_ctx : ctx
-  val connect_uri : ctx:ctx -> Uri.t -> (IO.conn * IO.ic * IO.oc) Lwt.t
+  val resolve : ctx:ctx -> Uri.t -> Endp.t IO.t
```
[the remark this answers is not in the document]
madroach: `resolve` is just `Resolver_lwt.resolve_uri ~uri ctx.resolver` of `Conduit_lwt`. So if you want a more specific name, maybe `resolve_uri`?

cohttp-lwt-unix/src/net.mli
```diff
@@ -33,6 +33,15 @@ val init : ?ctx:Conduit_lwt_unix.ctx -> ?resolver:Resolver_lwt.t -> unit -> ctx
     {!Resolver_lwt.t} with a different name resolution strategy (for instance to
     override a hostname to point it to a Unix domain socket). *)
 
+module Endp : Map.OrderedType with type t = Conduit.endp
+
+val resolve : ctx:ctx -> Uri.t -> Endp.t Lwt.t
+
+val connect_endp :
```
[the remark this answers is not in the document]
madroach: I added a small description to both of them. Honestly I don't understand what the current docstring of `connect_uri` is talking about. Reading the code `ctx` is not modified. Therefore I don't repeat this stuff in the documentation of `connect_endp`. Maybe it should be removed from `connect_uri`, too?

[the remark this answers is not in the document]
mseri: `uri` is in the type, we can probably leave it as is

cohttp-lwt/src/s.ml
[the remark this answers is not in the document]
mseri: It's ok, let's leave it as resolve, the `uri` is already in the type

cohttp-lwt/src/connection_cache.ml
mseri: I don't have such a strong opinion, but I think in ocaml < 4.13 this is slower than function application, and I think it reads more easily the non piped way

cohttp-lwt/src/connection_cache.ml
mseri: This one I prefer direct, without piping

cohttp-lwt/src/connection_cache.ml
madroach: I suspect I thought it easier to export a compare function from cohttp-lwt-unit than a hash function, but missed that `Hashtbl` is polymorphic anyway. I now use `Hashtbl`.

cohttp-lwt/src/s.ml
```diff
@@ -37,6 +86,12 @@ end
 module type Client = sig
   type ctx
 
+  module Connection : Connection with type Net.ctx = ctx
```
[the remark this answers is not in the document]
madroach: I added documentation. I will work on some examples.

cohttp-lwt/src/client.ml
```diff
@@ -1,94 +1,45 @@
...
     | `DELETE -> false
     | _ -> true
 
-  let call ?(ctx = Net.default_ctx) ?headers ?(body = `Empty) ?chunked meth uri
-      =
+  let cache = ref No_cache.(request (create ()))
+
+  let set_cache c = cache := c
+
+  let call ?ctx ?headers ?(body = `Empty) ?chunked meth uri
```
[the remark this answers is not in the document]
madroach: done

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+    | Connecting of (IO.ic * IO.oc) Lwt.t
+    | Full of (IO.ic * IO.oc)
+    | Closing of (IO.ic * IO.oc)
+    | Half of IO.ic
+    | Closed
+    | Failed of exn
+  type req_resr =
+    { req :Request.t
+    ; body :Body.t
+    ; res_r :(Response.t * Body.t) Lwt.u
```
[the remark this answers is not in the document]
madroach: Oh, I don't need the `Lwt.t` anymore. I removed it.

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+
+  let src = Logs.Src.create "cohttp.lwt.client" ~doc:"Cohttp Lwt client"
+
+  module Log = (val Logs.src_log src : Logs.LOG)
+
+  type state =
+    | Connecting of (IO.ic * IO.oc) Lwt.t
+    | Full of (IO.ic * IO.oc)
+    | Closing of (IO.ic * IO.oc)
+    | Half of IO.ic
```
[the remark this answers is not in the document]
madroach: done

cohttp-lwt-unix/src/client.mli
```diff
@@ -1,10 +1,22 @@
 (** The [Client] module implements the full UNIX HTTP client interface,
     including the UNIX-specific functions defined in {!C}. *)
 
-include Cohttp_lwt.S.Client with type ctx = Net.ctx
+include Cohttp_lwt.S.Client with type ctx := Net.ctx
+
+module Connection_cache : sig
```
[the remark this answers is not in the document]
madroach: done

cohttp-lwt-unix/src/client.mli
```diff
@@ -1,10 +1,22 @@
 (** The [Client] module implements the full UNIX HTTP client interface,
     including the UNIX-specific functions defined in {!C}. *)
 
-include Cohttp_lwt.S.Client with type ctx = Net.ctx
+include Cohttp_lwt.S.Client with type ctx := Net.ctx
```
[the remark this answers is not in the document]
madroach: Oh, we don't. I revert the change.

cohttp-lwt/src/client.ml
```diff
@@ -1,94 +1,45 @@
...
 
-module Make (IO : S.IO) (Net : S.Net with module IO = IO) = struct
-  module IO = IO
-  module Response = Make.Response (IO)
-  module Request = Make.Request (IO)
-
-  let src = Logs.Src.create "cohttp.lwt.client" ~doc:"Cohttp Lwt client"
-
-  module Log = (val Logs.src_log src : Logs.LOG)
+module Make (IO : S.IO) (Net : S.Net) =
```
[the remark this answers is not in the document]
madroach: `IO` is unused here since it is included in `Net`. We may still keep the constraint though.

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+    Queue.iter (fun {res_r; _} -> Lwt.wakeup_later_exn res_r e) q;
+    Queue.clear q;
+    Lwt_condition.broadcast_exn connection.condition e
+
+  let close connection =
+    match connection.state with
+    | Connecting channels ->
+      connection.state <- Closed;
+      Lwt.cancel channels;
+      Lwt.on_success channels (fun (ic, oc) -> Net.close ic oc);
```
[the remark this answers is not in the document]
madroach: It doesn't really matter. Either one will be a no-op. One cannot cancel a fulfilled promise.

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+      connection.state <- Closed;
+      Lwt_condition.broadcast connection.condition ()
+    | Closed | Failed _ -> ()
+
+  let shutdown connection =
+    match connection.state with
+    | Connecting channels ->
+      Lwt.async @@ fun () ->
+      channels >>= fun channels ->
+      (* need to loose possible race with writer *)
```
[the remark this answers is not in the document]
madroach: correct

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+      Lwt.pause () >>= fun () ->
```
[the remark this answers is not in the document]
madroach: Hmm. It seems there is no race anymore. What I was thinking about is this:
```
open Connection
let connection = create endpoint in
let response = request connection request in
shutdown connection
(* now both, shutdown and the request wait for the channels to become ready *)
(* request would fail if shutdown won the race *).
```
But now `request` returns immediately without handling control to Lwt having enqueued the request.
The writer won't be bothered by the `Closing` state, since it sees the full queue.

I may need to have another look. But right now I can't find the race.

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+      Lwt.return_unit
+    | Full channels ->
+      connection.state <- Closing channels;
+    | Closing _ | Half _ | Closed | Failed _ -> ()
+
+  let is_closed connection =
+    match connection.state with
+    | Full _ -> false
+    | Connecting _ -> false
+    | Closing _ | Half _ -> true (* really ? *)
```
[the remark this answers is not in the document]
madroach: Yes. I removed that comment. What people will use that function for is "May I still enqueue requests". You may not on a `Closing` connection.

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+      | `Eof ->
+        begin match connection.state with
+        | Full (_, oc) | Closing (_, oc) ->
+          Net.close_out oc;
+          connection.state <- Closed
+        | Half _ ->
+          connection.state <- Closed
+        (* Failed is no special case since we read proper EOF. *)
+        | Closed | Failed _ -> ()
+        | Connecting _ -> assert false
```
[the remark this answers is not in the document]
madroach: ok. But what do you put in such a place to avoid non-exaustive pattern matching warnings ?

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+      end;
+    connection
+
+  let connect ?finalise ?persistent ?ctx uri =
+    let connection = create ?finalise ?persistent ?ctx uri in
+    match connection.state with
+    | Connecting channels ->
+      channels >>= fun _ -> Lwt.return connection
+    | _ -> Lwt.return connection
+end
```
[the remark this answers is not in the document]
madroach: I'll work on an example.

cohttp-lwt/src/connection_cache.ml
```diff
@@ -0,0 +1,150 @@
...
+  module IO = Net.IO
+  open IO
+
+  type ctx = Net.ctx
+
+  exception Retry = S.Retry
+
+  module Endp_map = Map.Make (Net.Endp)
+
+  (* type t = < request : ?body:Body.t -> Cohttp.Request.t -> (Cohttp.Response.t * Body.t) IO.t > *)
```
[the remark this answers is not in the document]
madroach: It's gone.

cohttp-lwt/src/connection_cache.ml
madroach: This comment is gone and included in the proper documentation.

cohttp-lwt/src/connection_cache.ml
madroach: To make it explicit I'm raising the very same exception. But I removed it as you suggest.

cohttp-lwt/src/connection.ml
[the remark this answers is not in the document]
mseri: I would consider that an incorrect use though, I would rather use Lwt.finally and call shutdown there, no?

madroach:
Hi, thanks for taking the time to review. In commit d8f3d70 I further gutted the `Client` module, since the functionality of encoding selection is indeed needed in `Connection`. The other changes apply the changes you requested or fix stuff I found while adding documentation and reading your review.

cohttp-lwt/src/connection.ml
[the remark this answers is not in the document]
mseri: Here I think it is fine, I don’t think it is worth coming out with some specific error that will never occur and I don’t think we can use `-> .`

cohttp-lwt/src/connection.ml
```diff
@@ -0,0 +1,264 @@
...
+          queue_fail connection connection.waiting Retry;
+          Lwt.return_unit
+        end
+      >>= fun () ->
+      writer connection
+    | Failed e ->
+      queue_fail connection connection.waiting e;
+      Lwt.return_unit
+    | Half _ | Closed -> Lwt.return_unit
+    | Connecting _ -> assert false
```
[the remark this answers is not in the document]
mseri: Same here. I think it is fine

cohttp-lwt/src/connection.ml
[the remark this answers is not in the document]
madroach: No, I actually want that to work. It allows one to send `EOF` before even reading the response, so the remote end knows we are finished as early as possible.

cohttp-lwt/src/s.ml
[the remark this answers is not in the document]
madroach: I added tests meant to serve as usage examples, too:
https://github.com/mirage/ocaml-cohttp/pull/853/files#diff-913a2fc30c256ffaba94e5029edda2ef06ef625b6e7f58a9ecb5a8371805b5bf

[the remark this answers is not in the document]
mseri: Fantastic! Thanks a lot

mseri (approved):
Thanks a lot. This is good to be merged in my opinion.
Can some of the other reviewers also have a look?

mseri:
Oh no, I need to take it back. We need to fix first

```
File "cohttp-mirage/src/client.ml", line 79, characters 10-51:
79 |   include Cohttp_lwt.Make_client (HTTP_IO) (Net_IO)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: The functor application is ill-typed.
       These arguments:
         HTTP_IO Net_IO
       do not match these parameters:
         functor (IO : Cohttp_lwt__.S.IO) (Net : $T2) -> ...
       1. Module HTTP_IO matches the expected module type Cohttp_lwt__.S.IO
       2. Modules do not match:
 [...]
```
and
```
File "cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.ml", line 1:
Error: The implementation cohttp-lwt-jsoo/src/cohttp_lwt_jsoo.pp.ml
       does not match the interface cohttp-lwt-jsoo/src/.cohttp_lwt_jsoo.objs/byte/cohttp_lwt_jsoo.cmi:
        ... ...
       The module `Connection' is required but not provided
       File "cohttp-lwt/src/s.ml", line 165, characters 2-56:
         Expected declaration
       The value `set_cache' is required but not provided
       File "cohttp-lwt/src/s.ml", lines 171-173, characters 2-8:
         Expected declaration
       In module Make_client_async:
       The value `request' is required but not provided
       File "cohttp-lwt/src/s.ml", lines 179-181, characters 2-74:
         Expected declaration
```

cohttp-lwt-unix/src/net.mli
```diff
@@ -33,6 +33,19 @@ val init : ?ctx:Conduit_lwt_unix.ctx -> ?resolver:Resolver_lwt.t -> unit -> ctx
     {!Resolver_lwt.t} with a different name resolution strategy (for instance to
     override a hostname to point it to a Unix domain socket). *)
 
+type endp = Conduit.endp
```
rgrinberg, line 36: Why are re-exporting this alias?

madroach: because it is added in `Cohttp_lwt.S.Net` as abstract type that needs to be satisfied. I changed cohttp-unix/src/net.mli to `include Cohttp_lwt.S.Net with type endp = Conduit.endp` to deduplicate code and documentation.

The `endp` type is needed in `Connection_cache`.

madroach:
Now `cohttp-lwt-jsoo` and `cohttp-mirage` are building fine. I did some restructuring of the cohttp-mirage module structure to make it more similar to cohttp-lwt-unix.
I would actually like to do more into this direction by moving `Cohttp_mirage.Client.(ctx, with_authenticator)` into the new `net.ml`, too. But this would break backwards-compatibility. Then the `Client.Make` functor could be made to accept a single `Connection` module as parameter like in the Cohttp_lwt_unix.Client functor.

mseri:
The new release of cohttp will be breaking stuff anyway. If mirage Devs are fine with it, I would not mind. Maybe it is better to postpone this to a followup PR and have a separate discussion

mseri:
Ping @hannesm @samoht

madroach:
Hi, had second thoughts about my choice for the request handler type. I used
``` ocaml
?body:Body.t -> Cohttp.Request.t -> (Cohttp.Response.t * Body.t) Lwt.t
```
But although looking concise and simple this complicates things. Mostly it is not very clear how the transfer encoding is supposed to be passed around. Is it chosen based on the `Body.t` variant? Is taken from `request.headers` ? Or from `request.encoding`? Those are non-trivial questions the user would need to be aware of. `Request.t` is not a good fit to be exposed in the client interfaces.

Therefore I now changed it to match `Client.call` without the `?chunked` parameter:
``` ocaml
?headers:Http.Header.t -> ?body:Body.t -> Http.Method.t -> Uri.t -> (Cohttp.Response.t * Body.t) Lwt.t
```

This no longer looks as concise, but is actually simpler to use because the user does no longer need to construct a `Request.t`. The implementation is simpler, too. It no longer has to take `request.encoding` and the heuristics built into `Request.make` into account when selecting the transfer encoding.

The `?chunked` parameter of `Client.call` is actually obsolete now. The same effect can be achieved by using the appropriate `Body.t` variant. `String` will result in Fixed encoding, `Strings` and `Stream` will result in `Chunked` encoding.

I'm sorry for this additional big change.

mseri:
Don’t be sorry, sounds like a good change. I will take some time to review it tomorrow evening in the train.

btw there is some new breakage:
```
File "cohttp-lwt-unix/test/test_sanity.ml", line 176, characters 59-61:
[18](https://github.com/mirage/ocaml-cohttp/runs/6155743686?check_suite_focus=true#step:6:18)
176 |         Request.write (fun _writer -> Lwt.return_unit) req oc >>= fun () ->
[19](https://github.com/mirage/ocaml-cohttp/runs/6155743686?check_suite_focus=true#step:6:19)
                                                                 ^^
[20](https://github.com/mirage/ocaml-cohttp/runs/6155743686?check_suite_focus=true#step:6:20)
Error: This expression has type Net.IO.oc
[21](https://github.com/mirage/ocaml-cohttp/runs/6155743686?check_suite_focus=true#step:6:21)
       but an expression was expected of type
[22](https://github.com/mirage/ocaml-cohttp/runs/6155743686?check_suite_focus=true#step:6:22)
         oc = Lwt_io.output Lwt_io.channel
```

thanks a lot for your efforts with this PR, it is really appreciated!

mseri:
```
Fatal error: exception Unix.Unix_error(Unix.EINVAL, "select", "")
```

cohttp-lwt/src/connection.ml
```diff
@@ -195,31 +198,23 @@ struct
...
+          begin match Body.transfer_encoding body with
+          | Fixed _ as e -> Lwt.return (e, body)
+          | Chunked as e when connection.persistent = `True ->
+            Lwt.return (e, body)
+          | Chunked (* connection.persistent <> `True *) ->
+            (* We don't know yet whether chunked encoding is supported.
+             * Therefore use fixed length encoding. *)
+            Body.length body >>= fun (length, body) ->
+            Lwt.return (Cohttp.Transfer.Fixed length, body)
+          | Unknown -> assert false
```
mseri, line 217: Should we raise an exception in this case?

cohttp-lwt/src/s.ml
```diff
@@ -185,28 +187,19 @@ module type Client = sig
...
-      [body] to {!val:request}.
 
       @return [(response, response_body)] Consume [response_body] in a
-      timely fashion. Please see {!val:requester} about how and why.
+      timely fashion. Please see {!val:call} about how and why.
 
       @param chunked use chunked encoding if [true]. The default is
       [false] for compatibility reasons.
-      @param ctx If provided, {!val:request} is not used, but
+      @param ctx If provided, no connection cache is not used, but
```
mseri, line 202: I am not sure about the wording of this parameters, I don't understand what you mean

cohttp-lwt-unix/test/test_client.ml
```diff
@@ -53,7 +53,7 @@ let test_server tests =
     >|= Cohttp_lwt_unix_test.response
   in
   Cohttp_lwt_unix_test.test_server_s
-    ~name:"Mutable ressources"
+    ~name:"mutable ressources"
```
mseri, line 56:
```suggestion
    ~name:"mutable resources"
```

cohttp-lwt-unix/test/test_sanity.ml
```diff
@@ -84,6 +84,7 @@ let check_logs test () =
 
 let ts =
   Cohttp_lwt_unix_test.test_server_s server (fun uri ->
+      let uri = Uri.with_host uri (Some "localhost") in
```
mseri, line 87: This was not a problem before, why does it need change?

cohttp-lwt/src/connection.ml
```diff
@@ -97,8 +97,8 @@ struct
 
   let rec reader connection =
     match connection.state with
-    | Connecting _ -> assert false
-    | Closed | Failed _ -> assert false
+    | Connecting _ | Failed _ -> assert false
```
mseri, line 100: If it may trigger, we should actually raise an exception instead

cohttp-lwt/src/connection.ml
madroach: `Body.transfer_encoding` does not return `Unknown`. I can't see how any exception would be more helpful than `Assert_failure`.

cohttp-lwt-unix/test/test_sanity.ml
madroach: the uri is something like `http://0.0.0.0:6394`. Listening on `0.0.0.0` is fine, but connecting to `0.0.0.0` / `ANY_ADDRESS` is a linuxism. I'm on OpenBSD.

cohttp-lwt/src/connection.ml
madroach: I don't believe it can trigger. `reader` is the only function setting `state` to `Failed _`. `writer` will only close the out channel, so that the `reader` may still receive responses. If this assertion still does trigger I'd like to know that and fix my understanding of the code.

cohttp-lwt-unix/test/test_sanity.ml
mseri: Ok but then let’s fix it here: https://github.com/mirage/ocaml-cohttp/blob/7aa9a33ab1a9fa9fe767b5fe259251436e41fd02/test_helpers/cohttp_lwt_unix_test/src/cohttp_lwt_unix_test.ml#L32

cohttp-lwt/src/connection.ml
mseri: One more reason to raise a meaningful exception, imho, to make sure that we can track down where it comes from. Assert false are usually a nightmare to pinpoint

madroach: This is how it failed on me:
```
Fatal error: exception File "cohttp-lwt/src/connection.ml", line 104, characters 16-22: Assertion failed
Raised at Cohttp_lwt__Connection.Make.reader in file "cohttp-lwt/src/connection.ml", line 104, characters 16-28
Called from Lwt.Sequential_composition.bind.create_result_promise_and_callback_if_deferred.callback in file "src/core/lwt.ml", line 1860,
characters 23-26
```
In that case it was pretty specific, which assertion failed. So I don't really get what you are missing here. What do you propose?

mseri: Oh, ok this is much more informative than I remember. Was this improved recently? In this case, let's leave it like that!

cohttp-lwt-unix/test/test_sanity.ml
madroach: done.

madroach:
> ```
> (cd _build/default/cohttp-lwt-unix/test && ./test_sanity.exe)
> [DEBUG][application]: Cohttp debugging output is active
> Fatal error: exception Unix.Unix_error(Unix.EINVAL, "select", "")
> ```

Its good to know this fails for you too and this is no problem of OpenBSD `select`

mseri:
It is still failing:
```
Warning: [cohttp.lwt.server]: Client connection error Connection reset by peer: read("")
Warning: [cohttp.lwt.server]: Client connection error Connection reset by peer: read("")
Fatal error: exception Unix.Unix_error(Unix.EINVAL, "select", "")
```

madroach:
Yes. It was an fd-leak on both out and in channels. I did assume the in channel would close automatically after producing EOF (I was wrong as it seems). The out channel wasn't closed because I set `state` to `Half ic` without closing `oc`. This prevented the `reader` and the `close` routine from closing it.

mseri:
```
File "cohttp-lwt-unix/test/dune", line 50, characters 0-86:
[31](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:31)
50 | (rule
[32](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:32)
51 |  (alias runtest)
[33](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:33)
52 |  (package cohttp-lwt-unix)
[34](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:34)
53 |  (action
[35](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:35)
54 |   (run ./test_client.exe)))
[36](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:36)
(cd _build/default/cohttp-lwt-unix/test && ./test_client.exe)
[37](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:37)
[INFO][application]: Running high-level interface
[38](https://github.com/mirage/ocaml-cohttp/runs/6179650289?check_suite_focus=true#step:7:38)
Fatal error: exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")

File: cohttp-lwt/src/s.ml [hunk 42]
```diff
@@ -20,23 +20,193 @@ end
...
+  val length : t -> int
+
+  (** Request notification on change of [length] and on closing. *)
+  val notify : t -> unit Net.IO.t
+
+  (** Queue a request. Please see {!type:requester}. *)
+  val call : t -> call
+end
+
+module type Connection_cache = sig
```
rgrinberg, line 175: This module type needs a documentation string.

cohttp-lwt/src/s.ml
```diff
@@ -20,23 +20,193 @@ end
...
+  module IO : IO
```
rgrinberg, line 176: Why do we need to include the `IO` module here?

cohttp-lwt-unix/src/net.mli
```diff
@@ -16,14 +16,13 @@
...
 type ctx = { ctx : Conduit_lwt_unix.ctx; resolver : Resolver_lwt.t }
 [@@deriving sexp_of]
 
-val default_ctx : ctx
-(** [default_ctx] is the default network context. It uses
-    [Conduit_lwt_unix.default_ctx] and [Resolver_lwt_unix.system]. *)
+include Cohttp_lwt.S.Net
+  with module IO = Io
+   and type ctx := ctx
+   and type endp = Conduit.endp
```
rgrinberg, line 25: Why don't we use a destructive substitution for `endp` as well?

cohttp-lwt-unix/test/dune
```diff
@@ -41,3 +41,14 @@
...
+(executable
+ (modules test_client)
+ (name test_client)
+ (libraries cohttp_lwt_unix_test cohttp-lwt-unix))
+
+(rule
+ (alias runtest)
+ (package cohttp-lwt-unix)
+ (action
+  (run ./test_client.exe)))
```
rgrinberg, line 54: You can rewrite this in a single stanza:

```
(test
 (name test_client)
 (modules test_client)
 (package ..)
 (libraries ...))
```

cohttp-lwt/src/s.ml
madroach: We don't since I changed the type of the handler function to ` ... -> (Response.t, Body.t) Lwt.t` It used to be ` ... -> (Response.t, Body.t) IO.t`.
I removed it.

cohttp-lwt-unix/src/net.mli
madroach: Because the interface missing a `endp` type would no longer satisfy `Cohttp_lwt.S.Net`. I can and need to do the destructive substitution only because left and right-hand-side bear the same name.

cohttp-lwt-unix/test/dune
mseri: That’s some legacy idiom, I’d say let’s leave it as is, and change it in a separate PR later

mseri:
I rebased on master and cleaned up the history. Let's check with the CI that things are fine. If there are issues and no further comments by next week, I'll merge this as well

mseri:
Thanks a lot! If there are comments, there is still scope to update this, but I think it is a nice and welcome feature to have

mseri:
I have left the history more extensive, since this touches many existing components and I think it is useful to still be able to track why some changes have happened, and why some changes were made to other packages. @rgrinberg is it ok if we merge with this history or you think it should shrink more?

mseri:
There has been plenty of time for comments, time to get some user feedback on this :)
Thanks a lot for the effort and the patience

PR #854, opened 2022-01-26
RFC : Port cohttp-lwt-unix to eio using lwt_eio 

bikallem:
This is a WIP port of `cohttp-lwt-unix` to eio using the `lwt_eio` (https://github.com/talex5/lwt_eio/) library. `lwt_eio` is an `eio` backend for `lwt` and allows existing lwt libraries to used together with `eio`. 

Asides: 
1. Mirage and jsoo packages are disabled for now since both mirage and jsoo don't yet have eio backend(AFAICT).
2. This resulting API lives in `cohttp-lwt-unix` and it is not backwards-compatible with existing codebases. Should this live off new package altogether?
3. I haven't migrated tests just yet.

The gist of the PR is the new `Cohttp_lwt_unix.Client` and `Cohttp_lwt_unix.Server` API which doesn't expose any `'a Lwt.t` in the module interface signature. This allows a direct style of coding where the `lwt` monad and its artefacts(`>>=`, `>|=`, bind, map etc) are less pronounced, visible or necessary.

The purpose of this Draft PR is to get a feeler for what the maintainers think of this PR? Is there an appetite for this work?


mseri:
Wow, thanks a lot. I will need a few day to look at it and play with it. From my point of view is great to have this, but will deserve some thinking. For example, this would make it only available on ocaml 5.0+, right?

bikallem:
> this would make it only available on ocaml 5.0+, right?
Correct. Also mirage and jsoo packages will be unavailable since jsoo and mirage don't yet support ocaml 5.0 effects.

avsm:
Thanks for this @bikallem! As a counterpoint, do you also have the WIP port of cohttp directly to eio? (without Lwt_eio). That would perhaps give us a feel for what a totally separate implementation of cohttp without Lwt looks like.

mseri:
If needed, I have a fork of cohttp and conduit lying around using only unix (mostly worked but was single threaded still, I have never found the time to fix it)

bikallem:
> Thanks for this @bikallem! As a counterpoint, do you also have the WIP port of cohttp directly to eio? (without Lwt_eio). That would perhaps give us a feel for what a totally separate implementation of cohttp without Lwt looks like.

Sure, I will create another PR with pure eio cohttp as a counter point.

smorimoto:
Am I the only one who is curious about how eio performs under cohttp as well?

mseri:
No. I can’t wait to test it!

bikallem:
I have opened a draft PR for eio-only cohttp - https://github.com/mirage/ocaml-cohttp/pull/857

rgrinberg:
> I have opened a draft PR for eio-only cohttp

Thanks. To proceed with this PR, I think we need a separate `cohttp-lwt-eio` package. It's not possible to break code for everyone who is using the current packages.

avsm:
Let's focus on getting #857 in first. This adapter will be easier to figure out once we have the pure eio version in

bikallem:
closing this since #857 adds cohttp-eio

PR #857, opened 2022-02-01
RFC: eio-only version of cohttp

bikallem:
This is a WIP PR which implements an eio-only version of cohttp. Therefore the API is necessarily different as compared to other cohttp packages (lwt, async). 

1. The HTPP protocol implemented is 1.1
2. It is developed against 4.12+domains switch.
3. It supports running the server in a multiple domains(cores) setting. It utilizes socket sharding (https://www.nginx.com/blog/socket-sharding-nginx-release-1-9-1/) so it is automatically load balanced among the cores/domains in linux. Load balancing needs to be investigated for other OSes. 
4. Only the server API done for now. A sample server - server1.ml - in included in the PR. It can also be viewed here - https://github.com/bikallem/ocaml-cohttp/blob/eio-3/cohttp-eio/examples/server1.ml. 
5. The server API aims to be composable. To precisely define composable, here composable means `Server.middleware` and `Server.handler` values can be defined by other external users and exist in various opam packages. And that such packages can be reused and composed together to form a web application. The demo - server1.ml - demonstrates one such use case of composing together `Server.handler` values. Next, `Server.middleware` is designed to cater to scenarios where external users can define web application functionality such as `cookie based sessions`, `sql db based sessions`, `authentication` and the like and a web application developer can per-use such packages. I intend to update the examples with such a demo middleware. The design of `Server.handler` and `Server.middleware` is mostly inspired by F# web library - https://github.com/giraffe-fsharp/Giraffe. However, it can also be found in another ocaml web library (https://github.com/aantron/dream).

I am opening the PR to get early feedback and to gauge maintainer appetite for this work. 

Latest benchmarking results:
1 Core performance:
image

Multicore performance: (24 cores)
image


mseri:
While the lwt implementation is in a delicate situation, requiring a big break with the past and the ecosystem, this one could be easily added as cohttp subpackage already now imho.

We could add an extra lower bound on 5.0.0 just for this package (leaving jsoo enabled: we know it does not yet support 5.0.0, but they can coexist, we just need to figure out the api).

It could be interesting then to evolve this and the conduit-free implementation of lwt-server together so that they api remains consistent.

Fwiw, I think we can even leave the client out for the time being. And focus on the server to start with.

avsm:
I'm in favour of @mseri's proposal. I think we can get an eio client in there too though -- it's a fairly small amount of code _vs_ the server.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -0,0 +1,158 @@
...
+end
+
+(** [Response] is a HTTP/1.1 response. *)
+module Response : sig
+  type t
+
+  and body =
+    [ `String of Cstruct.t
+    | `Chunked of write_chunk
+    | `Custom of Faraday.t -> unit
```
mseri, line 99: Neat

cohttp-eio/src/parser.ml
```diff
@@ -0,0 +1,228 @@
+open Angstrom
```
mseri, line 1: The new parser in the Http module is pretty fast, is there a reason you needed to reimplement it?

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -0,0 +1,158 @@
...
+        length : int;
+        extensions : chunk_extension list;
+      }
+    | Last_chunk of chunk_extension list
+
+  and chunk_extension = { name : string; value : string option }
+end
+
+(** [Request] is a HTTP/1.1 request. *)
+module Request : sig
```
mseri, line 55: There are signatures for these modules in the http and cohttp packages, is there a reason you are re-defining them all?

mseri:
It looks like there is a lot of re-implementation. I think, if we call it cohttp-*, it should re-use signature from the `http` module (I would suggest also the parser). And, if anything needs being extended or improved, we do it at those libraries level. Ideally `cohttp-eio` would depend on `eio` and `cohttp` and get the relevant signatures from there, so that as much as possible of the types and the interfaces are shared.

cohttp-eio/src/cohttp_eio.mli
mseri: can we extend that instead? Like `include Cohttp.Request` and add any extra needed function? I am just thinking of ways to have as compatible interfaces as possible

rgrinberg: You could just provide one function that returns the underlying `Http.Request.t`. This would allow users to re-use whatever functions they have working on existing types.

rgrinberg:
> The server API aims to be composable. To precisely define composable, here composable means Server.middleware and Server.handler values can be defined by other external users and exist in various opam packages.  ...

I don't think we should be introducing web framework components into cohttp. Especially since this was already implemented outside of cohttp without loss of functionality.

rgrinberg:
Why did you decide to use cstructs? Anil mentioned that strings are the way to go with multicore.

.envrc
```diff
@@ -0,0 +1,2 @@
+use nix shell-eio.nix 
+eval $(opam env --shell=bash --inplace-path --set-switch)
```
rgrinberg, line 2: this shouldn't be committed. not every is using direnv.

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,260 @@
...
+      | `String cs ->
+          Http.Header.add_unless_exists headers hdr
+            (string_of_int @@ Cstruct.length cs)
+      | `Chunked _ -> Http.Header.remove headers hdr
+      | `Custom _ -> headers
+      | `None -> Http.Header.add_unless_exists headers hdr "0"
+    in
+    let headers =
+      let date = Unix.time () |> Unix.gmtime in
+      Http.Header.add_unless_exists headers "Date" (datetime_to_string date)
```
rgrinberg, line 131: Do we really need this? I don't remember `Date` being mandatory.

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,260 @@
...
+      let status_phrase = Http.Status.reason_phrase_of_code status in
+      Printf.sprintf "%s %d %s\r\n" version status status_phrase
+    in
+    Faraday.write_string faraday status_line;
+    let headers = Http.Response.headers res in
+    let headers =
+      (*--- Don't cache set-cookie headers in browsers and proxies. ---*)
+      if Http.Header.mem headers "set-cookie" then
+        Http.Header.add headers "Cache-Control" {|no-cache="Set-Cookie"|}
+      else headers
```
rgrinberg, line 117: We can just let downstream frameworks take care of this for us.

dune-project
```diff
@@ -113,27 +113,27 @@ should also be fine under Windows too.
...
+;   (http (= :version))
+;   (cohttp (= :version))
+;   (cohttp-lwt (= :version))
+;   logs
+;   (lwt (>= 3.0.0))
+;   (lwt_ppx :with-test)
+;   (conf-npm :with-test)
+;   (js_of_ocaml (>= 3.3.0))
+;   (js_of_ocaml-ppx (>= 3.3.0))
+;   (js_of_ocaml-lwt (>= 3.5.0))))
```
rgrinberg, line 136: Why is this commented out?

dune
```diff
@@ -0,0 +1 @@
+(data_only_dirs cohttp-lwt-jsoo cohttp_lwt_jsoo_test)
```
rgrinberg, line 1: We don't need this dune file.

dune-project
```diff
@@ -230,3 +230,19 @@ online at https://hdrhistogram.github.io/HdrHistogram/plotFiles.html.")
...
+ (synopsis "CoHTTP implementation with eio backend")
+ (description "A CoHTTP server and client implementation based on `eio` library. `cohttp-eio`\
+              features a multicore capable HTTP 1.1 server. The library promotes and is built \
+              with direct style of coding as opposed to a monadic.")
+ (depends
+  eio
+  eio_main
+  cstruct
+  logs
+  (fmt (>= 0.8.7))
```
rgrinberg, line 245: I could be wrong, but I don't see `fmt` or `logs` being used anywhere.

cohttp-eio/src/parser.ml
anuragsoni: If bigstrings will make it easier for `eio` to consume the parser from the Http module, we can revisit the choice of using bytes there, and update everything to work with big strings. Lwt/Async shouldn't have any issues working with big strings.

tmattio:
Disclaimer: not a maintainer of Cohttp or Eio, so I'm probably missing a bit of context.

What's the reason for adding this as a new `cohttp-eio` package?
With effects allowing us to write APIs without monads, the use of Eio becomes an implementation detail that doesn't leak to the package interface (apart from the `Reader.create` function apparently, but I'm guessing that's something that could be changed). So I was expecting this would allow package maintainers to stop having to maintain multiple packages for each IO backend.

Is there something preventing us from implementing the functionality in `cohttp` and have the `lwt` and `async` backends "wrap" the `cohttp` package to keep the API compatibility?

rgrinberg:
Event if we could completely drop support for 4.x, the Lwt and Async packages have their own event loops while Eio has its own. I doubt people want to use a cohttp-{lwt,async} package that brings its own event loop.

dune-project
bikallem: jsoo doesn't build on 4.12+domains switch.

dune
bikallem: just so that dune doesn't build jsoo packages.

bikallem:
> Disclaimer: not a maintainer of Cohttp or Eio, so I'm probably missing a bit of context.
> 
> What's the reason for adding this as a new `cohttp-eio` package? With effects allowing us to write APIs without monads, the use of Eio becomes an implementation detail that doesn't leak to the package interface (apart from the `Reader.create` function apparently, but I'm guessing that's something that could be changed). So I was expecting this would allow package maintainers to stop having to maintain multiple packages for each IO backend.
> 
> Is there something preventing us from implementing the functionality in `cohttp` and have the `lwt` and `async` backends "wrap" the `cohttp` package to keep the API compatibility?

`eio` sits atop effects, uring and luv and has its own IO loop which works a bit differently from lwt/async. If you want direct style code in a lwt codebase, there is `lwt_eio` package which does that.  Here is another PR which attempts to do just that (https://github.com/mirage/ocaml-cohttp/pull/854). Perhaps going forward if `eio` gains traction and becomes the "default"/blessed io package, then library authors will probably stop maintaining lwt/async packages.

bikallem:
> > The server API aims to be composable. To precisely define composable, here composable means Server.middleware and Server.handler values can be defined by other external users and exist in various opam packages.  ...
> 
> I don't think we should be introducing web framework components into cohttp. Especially since this was already implemented outside of cohttp without loss of functionality.

Agreed. It won't. `cotthp-eio` will provide types and maybe some glue functions to help downstream web frameworks/components provide "web framework/component" like functionalities.

cohttp-eio/src/server.ml
bikallem: According to this spec (https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1.2), `Date` seems to be mandatory if the server is capable. To quote `An
   origin server MUST send a Date header field in all other cases.`

tmattio:
> Even if we could completely drop support for 4.x, the Lwt and Async packages have their own event loops while Eio has its own. I doubt people want to use a cohttp-{lwt,async} package that brings its own event loop.

Couldn't eio offer an API to provide your own event loop? So people could use the same package but call `Eio_lwt.init_event_loop ()` or similar to tell Eio to use Lwt (or Async's) event loop.

cohttp-eio/src/server.ml
bikallem: golang http lib also seems to add this header - https://github.com/golang/go/blob/master/src/net/http/server.go#L1416-L1417

bikallem:
> Why did you decide to use cstructs? Anil mentioned that strings are the way to go with multicore.

`Cstruct.t` is the default/chosen data structure in `eio`, so we use Cstruct.t also to save the hassle of  of data structure conversion/allocation/gc cost. With regards to cstruct/bigstring, AIUI Cstruct.t/bigstring allows zero copy/no allocation when data is moved from C memory space to OCaml managed memory space. Furthermore, on linux, uring removes the need to do allocation/copying when moving data from OS kernel space to user space. `io_uring` which `eio` uses has been designed to explicitly exploit these functionalities. `cohttp-eio` aims to also exploit these to the fullest possible. Perhaps @talex5 can elucidate.

cohttp-eio/src/server.ml
rgrinberg: That's a good point, but I'd still go against it. For a few reasons:

* We don't do it in the lwt/async. It's inconsistent with respect to our own libs.
* The implementation we have here is naive. Getting the current time is takes a syscall and should be cached for high performance servers.
* It's just as easy to implement outside of the basic library. It could just as easily be done through middleware. Or even a simple response generation function which would work for all backends.

dune
rgrinberg: I see. Note that you could always build just the package you want with `dune build cohttp-eio.install`.

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,260 @@
...
+      {
+        Client_connection.flow;
+        addr;
+        switch = sw;
+        reader = Reader.create (flow :> Eio.Flow.read);
+        oc = (flow :> Eio.Flow.write);
+      }
+  done
+
+let create ?(socket_backlog = 10_000) ?(domains = cpu_core_count) ~port
```
rgrinberg, line 234: What do you think about making the default for `domains` to be `1`? To me seems like the safest option since the user's code might not necessarily be thread safe.

rgrinberg:
> > Why did you decide to use cstructs? Anil mentioned that strings are the way to go with multicore.
> 
> `Cstruct.t` is the default/chosen data structure in `eio`, so we use Cstruct.t also to save the hassle of of data structure conversion/allocation/gc cost. With regards to cstruct/bigstring, AIUI Cstruct.t/bigstring allows zero copy/no allocation when data is moved from C memory space to OCaml managed memory space. Furthermore, on linux, uring removes the need to do allocation/copying when moving data from OS kernel space to user space. `io_uring` which `eio` uses has been designed to explicitly exploit these functionalities. `cohttp-eio` aims to also exploit these to the fullest possible. Perhaps @talex5 can elucidate/correct the above if my understanding is incorrect.

Thanks for response. This still needs some clarification. AFAIK, normal strings aren't going to be moved by the GC in 5.0 so they should be usable for io_uring and not require extra copies anyway.

rgrinberg:
> Agreed. It won't. cotthp-eio will provide types and maybe some glue functions to help downstream web frameworks/components provide "web framework/component" like functionalities.

In which case, we should drop the middleware alias. You have a noble goal in mind, but I doubt it will have the outcome we all desire. Who is to say that the various OCaml web frameworks popular today are going to use this type? I don't yet see any interest from them their maintainers, therefore I doubt that this is going to be adopted. To get some unification here, we need to get the various maintainers on board first and *only then* start introducing stuff to smooth out compatibility. Otherwise, we're just going to introduce the nth+1 framework.

cohttp-eio/src/server.ml
avsm: Agreed. We can drop Date for now and put it in the cohttp-consuming libraries

dinosaure:
> Thanks for response. This still needs some clarification. AFAIK, normal strings aren't going to be moved by the GC in 5.0 so they should be usable for io_uring and not require extra copies anyway.

Currently, from what @kayceesrk explained to me, we should not assume that a large `string` will never move. It's currently the case because compaction is not yet implemented. But the only true _caml_ object which should never moves (for other reasons than `malloc`) is a `bigarray`.

cohttp-eio/src/server.ml
bikallem: Removed now.

cohttp-eio/src/server.ml
bikallem: Removed now.

talex5 (commented):
Thanks for looking into this! Comments on the code inline.

To reply to some of the comments here:

> Why did you decide to use cstructs? Anil mentioned that strings are the way to go with multicore.

There's an issue on Eio about deciding which to use: https://github.com/ocaml-multicore/eio/issues/140

> What's the reason for adding this as a new cohttp-eio package? [...] Is there something preventing us from implementing the functionality in cohttp and have the lwt and async backends "wrap" the cohttp package to keep the API compatibility?

For Lwt at least this should be quite easy using Lwt_eio, which allows running eio fibres from lwt, and awaiting lwt promises from eio.

> Couldn't eio offer an API to provide your own event loop?

It does, and we have both luv and uring event loops already. However, Lwt_eio does it the other way around. That has the advantage of supporting multiple domains for Eio code (even if Lwt code can only run in one of them), and allows using the faster io_uring for Eio operations.

cohttp-eio/src/parser.ml
```diff
@@ -0,0 +1,207 @@
...
+      let field_content =
+        let c2 =
+          optional
+            (let+ c1 = skip_many1 (space <|> htab) *> vchar in
+             Format.sprintf " %c" c1)
+          >>| function
+          | Some s -> s
+          | None -> ""
+        in
+        lift2 (fun c1 c2 -> Format.sprintf "%c%s" c1 c2) vchar c2
```
mseri, line 35: We did a number of benchmarks some time ago. Using `String.concat` or `Buffer` can be orders of magnitude faster than `Printf.sprintf`. Of course, here we are using `%X`, so we may not have a an alternative without relying on extra code or library.

I still think that you should try to use our parser from http, where needed we can adjust it or abstract it over bytes/bigstrings. It feels wrong to ship two different http parsers.

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,188 @@
...
+          match value with None -> "" | Some v -> Printf.sprintf "=%s" v
+        in
+        Printf.sprintf ";%s%s" name v |> Buffer.add_string buf)
+      exts;
+    Buffer.contents buf
+  in
+  let write = function
+    | Chunk.Chunk { size; data; extensions = exts } ->
+        let buf =
+          Printf.sprintf "%X%s\r\n%s\r\n" size (extensions exts)
```
mseri, line 55: The `%X` mentioned in the other comment is used here

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,216 @@
...
+  let sockaddr = `Tcp (Eio.Net.Ipaddr.V4.loopback, t.port) in
+  let ssock =
+    Eio.Net.listen ~reuse_addr:true ~reuse_port:true ~backlog:t.socket_backlog
+      ~sw net sockaddr
+  in
+  let on_accept_error exn =
+    Printf.fprintf stderr "Error while accepting connection: %s"
+      (Printexc.to_string exn)
+  in
+  while not (Atomic.get t.closed) do
```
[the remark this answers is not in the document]
mseri: Thanks. This should go in the module documentation then

shell-eio.nix
```diff
@@ -0,0 +1,29 @@
...
+    libev
+    hidapi
+    libffi
+    jq
+    zlib
+    rustup
+    opam
+    bc
+    httpie
+  ];
```
mseri, line 26: I don't think we need most of these packages

cohttp-eio/src/parser.ml
bikallem: Let me get back to you on this issue a bit later.

shell-eio.nix
bikallem: Yes, agreed. The whole file `shell-eio.nix` is there because I don't know how to use/install trunk ocaml 5.00 via the nix package system. trunk ocaml 5.00 because it has multicore/effect support. I am a newly minted newbie on nixos so the nix packaging system, nix language are all alien to me now.

cohttp-eio/src/client_connection.ml
```diff
@@ -0,0 +1,9 @@
+type t = {
+  flow : < Eio.Flow.two_way ; Eio.Flow.close >;
+  switch : Eio.Std.Switch.t;
+  addr : Eio.Net.Sockaddr.t;
+  reader : Reader.t;
+  oc : Eio.Flow.write;
```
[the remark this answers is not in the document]
bikallem: removed now.

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,216 @@
...
+    Eio.Net.accept_sub ~sw ssock ~on_error:on_accept_error
+    @@ fun ~sw flow addr ->
+    handle_request t
+      {
+        Client_connection.flow;
+        addr;
+        switch = sw;
+        reader = Reader.create (flow :> Eio.Flow.read);
+        oc = (flow :> Eio.Flow.write);
```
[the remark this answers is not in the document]
bikallem: removed now.

cohttp-eio/src/server.ml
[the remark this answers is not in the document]
talex5: (in fact, you don't even need to do this in each domain, as cancelling the main `Server.run` one will cancel all the sub-domains it spawned)

mseri:
You can probably easily port the tests (especially the fuzzy ones) from Http, Cohttp and cohttp-lwt-Unix to test this PR automatically

cohttp-eio/src/server.ml
```diff
@@ -0,0 +1,216 @@
...
+      }
+  done
+
+let create ?(socket_backlog = 10_000) ?(domains = cpu_core_count) ~port
+    request_handler : t =
+  { socket_backlog; domains; port; request_handler; closed = Atomic.make false }
+
+(* wrk2 -t10 -c400 -d30s -R2000 http://localhost:3000 *)
+let run (t : t) =
+  Eio_main.run @@ fun env ->
```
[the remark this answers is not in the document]
bikallem: It is removed now. Thanks.

cohttp-eio/src/reader.ml
```diff
@@ -0,0 +1,104 @@
+(* Based on https://github.com/inhabitedtype/angstrom/blob/master/lib/buffering.ml *)
```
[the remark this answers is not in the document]
bikallem: The reader here uses a different buffering strategy than the one implemented for Angstrom.

cohttp-eio/src/server.ml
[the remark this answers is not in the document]
bikallem: do you mean by calling `Switch.fail sw` ?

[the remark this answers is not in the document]
talex5: Yes, that will do it.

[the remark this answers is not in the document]
bikallem: Also I'd rather go for graceful stop/shutdown, i.e. stop accepting new connections and close the connection after sending any outstanding responses.

cohttp-eio/src/parser.ml
bikallem: @mseri `cohttp-eio` now uses the parser included in `http` package.

mseri: 🥳 thanks! Did you notice any problem with it performance-wise?

bikallem: Performance was good, slightly better than with angstrom based parser. Do you think if using bigstring as the base type in parser(instead of string) and thus eliminating bigstring to string conversion cost will increase in further performance boost?

bikallem: I just did a multi-core benchmark of the `http` parser against `httpaf-eio` and `nethttp-go`, It seems the  parser doesn't scale linearly as compared to the other two. It starts plateauing in/around 300K req/second.
image
The branch is here if your are interested. https://github.com/bikallem/ocaml-cohttp/blob/eio-3-multi-core/cohttp-eio/src/reader.ml

bikallem: It seems `eio_linux` with polling plateaus around 300k, so I changed it back to `eio_main`. With that change we don't have any scaling issue. The latest multicore benchmark (48 cores) is a below. It scales very nicely without ever plateauing. 
image

mseri: Wow. Impressive

bikallem:
@mseri how do we handle the conflicts? This PR requires the jsoo packages to be disabled/not compiled since 4.12+domains switch is not compatible with jsoo and jsoo packages. Should we wait for ocaml 5.00?

mseri:
Re enable it. It will not be available in 5.0.0 due to jsoo unavailability, as this package which will only be available when “base-domains” is true (add a dependency on that please). We can deal with the full extent of the CI separately

cohttp-eio/src/parser.ml
talex5: > The latest multicore benchmark (48 cores) is below. It scales very nicely without ever plateauing.

Perhaps this is true, but the graph doesn't show that:
- With 1 core, it shows it's possible to do about 100,000 requests per second
- With 48 cores it shows the new code doing about 400,000 requests per second

We don't know how fast it goes with 48 cores because the x-axis only goes up to 400,000 and all the multicore versions can handle that with no problem, as expected. We need to know where each one levels off.

It would be more useful to graph the maximum speed against the number of cores being used. Possibly replacing `wrk2` with `wrk` would make that easier.

bikallem:
@mseri This server API now ready for review. The parser is optimized specific to eio (i.e. it is not usable outside of OCaml 5.0) and implements the HTTP chunk encoding spec (https://datatracker.ietf.org/doc/html/rfc7230#section-4.1) including the chunk trailer part(https://datatracker.ietf.org/doc/html/rfc7230#section-4.1.2) and decoding chunked part algorithm(https://datatracker.ietf.org/doc/html/rfc7230#section-4.1.3).

mseri:
@talex5 is there any way to make the OCaml ci be more selective when building? Like don’t build cohttp-eio unless we are on 5.00?

talex5:
> @talex5 is there any way to make the OCaml ci be more selective when building? Like don’t build cohttp-eio unless we are on 5.00?

There's an issue for it here: https://github.com/ocurrent/ocaml-ci/issues/297

bikallem:
I have looked at the ocaml-ci issue, but I am not sure if I understand the resolution. Do we wait for the resolution of the issue before we can merge this PR? 

@mseri is https://github.com/ocaml/setup-ocaml an option as the main ci as opposed to ocaml-ci? 
/cc @smorimoto

.github/workflows/workflow.yml
```diff
@@ -1,51 +1,99 @@
...
+          opam-local-packages: ${{ matrix.local-packages }}
 
       - run: yarn --frozen-lockfile
         working-directory: cohttp_lwt_jsoo_test
 
       - run: echo "PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig" >>"$GITHUB_ENV"
         if: ${{ matrix.os == 'macos-latest' }}
 
-      - run: opam install . --deps-only --with-test
+      - run: opam depext -i re uri uri-sexp sexplib0 ppx_sexp_conv stringext fmt alcotest conf-libev lwt cmdliner conduit-lwt-unix magic-mime logs ounit js_of_ocaml-lwt js_of_ocaml-ppx async async_unix base core core_unix conduit-async mirage-crypto ipaddr conduit-mirage mirage-kv astring base_quickcheck ppx_assert ppx_compare ppx_here core_bench crowbar ocurl mirage-channel lwt_ppx core_kernel.v0.14.2
```
mseri, line 52: I'd like to avoid having to spell everything out by hand just for core-kernel or fix that. I can look at it in a separate PR if you want

.github/workflows/nix.yml
```diff
@@ -1,16 +0,0 @@
-name: "Nix"
```
mseri, line 1: Even if it is currently brittle, we should not remove this CI test

.github/workflows/nix.yml
```diff
@@ -1,16 +0,0 @@
...
-    runs-on: ubuntu-latest
-    steps:
-    - name: Checkout code
-      uses: actions/checkout@v2
-      with:
-        submodules: true
-    - uses: cachix/install-nix-action@v13
-      with:
-        nix_path: nixpkgs=channel:nixos-unstable
-    - run: nix-shell --command "dune runtest"
```
mseri, line 16:
```
--command "dune runtest bench cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-jsoo cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top cohttp_async_test cohttp_lwt_jsoo_test cohttp_lwt_unix_test cohttp_server cohttp_test http --profile release"
```
should do the trick (once the ocamlformat issue is fixed in a separate PR)

mseri:
Besides some minor points commented above, this looks good to me.

I'd like to hear some more voices before merging though

.github/workflows/workflow.yml
mseri: This should be fixed by https://github.com/mirage/ocaml-cohttp/pull/869
Once that is merged, you should be able to just use `- run: opam install . --deps-only --with-test`

.envrc
mseri: @bikallem this file still needs to be removed

.github/workflows/workflow.yml
mseri: By the way, why using depext worked while using install broke? Opam 2.1 should install the depexts automatically, no?

.github/workflows/workflow.yml
```diff
@@ -1,51 +1,99 @@
...
 
-      - run: opam exec -- make build
+      - run: opam exec -- dune build -p http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-lwt-jsoo,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top --profile release
```
mseri, line 54: Dune build also builds the examples, it was done on purpose

mseri: I understand why it is needed now, I think we should instead run:
```
- run: |
    opam exec -- dune build --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-lwt-jsoo,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top --profile=dev
    opam exec -- dune build examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
```
So that it builds the extra binaries and it fails on warnings

.github/workflows/workflow.yml
mseri: I also see why you want this, otherwise we are already installing plenty of cohttp packages already... argh

.github/workflows/workflow.yml
```diff
@@ -1,51 +1,99 @@
...
 
-      - run: opam exec -- make test
+      - run: opam exec -- dune runtest bench cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-jsoo cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top cohttp_async_test cohttp_lwt_jsoo_test cohttp_lwt_unix_test cohttp_server cohttp_test http --profile release
```
mseri, line 56:
```suggestion
      - run: opam exec -- dune runtest cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-jsoo cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top cohttp_async_test cohttp_lwt_jsoo_test cohttp_lwt_unix_test cohttp_server cohttp_test http --profile release
```
We don't want to run the benchmarks in the CI, just build them

mseri:
Sorry for the complication, I have fixed the CI yesterday and there are now some conflicts to resolve

.github/workflows/workflow.yml
```diff
@@ -24,10 +24,6 @@ jobs:
           - ocaml-variants.4.12.1+options,ocaml-option-afl
           - ocaml-variants.4.13.1+options,ocaml-option-afl
           - ocaml-variants.4.14.0+options,ocaml-option-afl
-        local-packages:
```
mseri: This you still need

.github/workflows/workflow.yml
```diff
@@ -42,11 +41,11 @@ jobs:
       - run: yarn --frozen-lockfile
         working-directory: cohttp-lwt-jsoo/test
 
-      - run: opam depext -i re uri uri-sexp sexplib0 ppx_sexp_conv stringext fmt alcotest conf-libev lwt cmdliner conduit-lwt-unix magic-mime logs ounit js_of_ocaml-lwt js_of_ocaml-ppx async async_unix base core core_unix conduit-async mirage-crypto ipaddr conduit-mirage mirage-kv astring base_quickcheck ppx_assert ppx_compare ppx_here core_bench crowbar ocurl mirage-channel lwt_ppx core_kernel.v0.14.2
+      - run: opam install . --deps-only --with-test
```
mseri, line 44: You should do what I did for macos in here, otherwise it fails trying to install eio:
```
- run: opam install --with-test --deps-only http cohttp cohttp-lwt cohttp-lwt-unix cohttp-server-lwt-unix cohttp-mirage cohttp-async cohttp-curl-async cohttp-curl-lwt cohttp-curl cohttp-top
      - run: |
          opam exec -- dune build --profile=dev --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top @install
          opam exec -- dune build -j1 bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
      - run: opam exec -- dune runtest --profile=dev cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top http
```

.github/workflows/workflow.yml
```diff
@@ -54,6 +50,38 @@ jobs:
...
+          opam-depext: true
+          opam-depext-flags: --with-test
+
+      - run: echo "PKG_CONFIG_PATH=$(brew --prefix openssl)/lib/pkgconfig" >>"$GITHUB_ENV"
+
+      - run: opam install --with-test --deps-only http cohttp cohttp-lwt cohttp-lwt-unix cohttp-server-lwt-unix cohttp-mirage cohttp-async cohttp-curl-async cohttp-curl-lwt cohttp-curl cohttp-top
+
+      - run: |
+          opam exec -- dune build --profile=dev --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top @install
+          opam exec -- dune build -j1 bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
```
mseri, line 81: `-j1` here is my mistake, I should have removed it

dune-project
```diff
@@ -266,37 +288,56 @@ conduit for SSL.")
...
+ (synopsis "CoHTTP implementation with eio backend")
+ (description
+  "A CoHTTP server and client implementation based on `eio` library. `cohttp-eio`features a multicore capable HTTP 1.1 server. The library promotes and is built with direct style of coding as opposed to a monadic.")
+ (depends
+  base-domains
+  eio
+  cstruct
+  bigstringaf
+  fmt
+  (http
```
mseri, line 342:
```suggestion
  (eio_main :with-test)
  (http
```

mseri: Is there a way to run the tests without `ncat`? Otherwise we need to add `conf-ncat` to `opam-repository` and add it among the test dependencies

.github/workflows/workflow.yml
```diff
@@ -79,6 +79,45 @@ jobs:
...
+      - name: Use OCaml ${{ matrix.ocaml-compiler }}
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+          dune-cache: ${{ matrix.os == 'ubuntu-latest' }}
+          opam-local-packages: ${{ matrix.local-packages }}
+
+      - run: opam repo add alpha git+https://github.com/kit-ty-kate/opam-alpha-repository.git && opam update
+
+      - run: opam install cstruct eio_main bigstringaf fmt mdx
```
mseri, line 119:
```suggestion
      - run: opam install --with-test --deps-only cohttp-eio
```

.github/workflows/workflow.yml
```diff
@@ -79,6 +79,45 @@ jobs:
...
+            cohttp-eio.opam
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - run: sudo apt-get install ncat
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+
+      - run: brew install nmap
+        if: ${{ matrix.os == 'macos-latest' }}
```
mseri, line 105: `ncat/nmap/nc` either should come with a `conf-` package added as a test dependency or we should find a different way to run the tests

.github/workflows/workflow.yml
bikallem: Indeed, setup-ocaml defaults to using opam 2.0 where --with-deps --with-test doesn't seem to work when we have `http {=version}` in the opam files.

.github/workflows/nix.yml
bikallem: Added the nix.yml back

.github/workflows/workflow.yml
bikallem: removed.

.github/workflows/workflow.yml
bikallem: This won't work with opam files that have  `(http (= :version))` constraint since setup-ocaml uses opam 2.0.

dune-project
bikallem: Done.

bikallem: Not at the moment, `ncat` in the nmap is the only cross platform `ncat` that works consistently across macos and linux. I will send a pr to create `conf-nmap` in the opam-repository shortly.

.github/workflows/workflow.yml
bikallem: Done.

.github/workflows/nix.yml
bikallem: Added the nix.yml back.

.github/workflows/workflow.yml
bikallem: reverted back.

.github/workflows/workflow.yml
bikallem: Done.

.github/workflows/workflow.yml
bikallem: Will create a PR to add `conf-nmap.opam` to opam-repository.

bikallem:
I have merged the macos/ubuntu jobs section into one since the ci build steps were nearly identical except one step.

.github/workflows/workflow.yml
```diff
@@ -35,53 +39,27 @@ jobs:
...
 
       - run: |
           opam exec -- dune build --profile=dev --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top @install
-          opam exec -- dune build -j1 bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
+          opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
+      
       - run: opam exec -- dune runtest --profile=dev cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top http
 
+      - run: opam exec -- make js-test
+
```
mseri, line 62: You need
```suggestion
      - if: ${{ matrix.os == 'ubuntu-latest' }} 
        run: |
          opam depext -i cohttp-lw-jsoo
          opam exec -- make js-test
```
Or something like it

dune-project
mseri: Ok, perfect, thanks!

.github/workflows/workflow.yml
bikallem: add cohttp-lwt-jsoo to `opam  install`

bikallem:
@mseri @rgrinberg Is there anything outstanding/to-be-done on this PR?

.github/workflows/workflow.yml
```diff
@@ -34,51 +39,64 @@ jobs:
...
-      - run: opam exec -- make test
+      - run: |
+          opam exec -- dune build --profile=dev --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top @install
+          opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
+      
+      - run: opam exec -- dune runtest --profile=dev cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top http
 
       - run: opam exec -- make js-test
 
-  build-test-macos:
```
mseri, line 52: If we remove this, then probably we should rename the build above to `build-test-default` or something like that

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
-- Remove unused values to help ci(@bikallem #873)
+- Address warning 32 errors (@bikallem #876)
+- Remove unused values. (@bikallem #873)
```
mseri, line 3: Please remove these two, and instead modify line 14/15 into:
```
- Various internal refactors (rgrinberg, mseri, bikallem #876, #873, #820, #800, #799, #797)
```

mseri: Up to you if you want to use the `@`, I had started removing them since a number of people were annoyed by the continuous emails from github due to CHANGELOG mentions

cohttp-eio/src/writer.ml
```diff
@@ -0,0 +1,49 @@
+(* Original https://github.com/talex5/httpaf/blob/eio/lib/serialize.ml *)
```
mseri, line 1: If this comes from somewhere else, we need to retain the copyright header:
```
https://github.com/talex5/httpaf/blob/eio/lib/serialize.ml#L1-L32
```

.github/workflows/workflow.yml
bikallem: Done. renamed to `build-test-default`

CHANGES.md
bikallem: Done.

cohttp-eio/src/writer.ml
bikallem: Added copyright notice.

.github/workflows/workflow.yml
```diff
@@ -47,6 +48,12 @@ jobs:
       - run: yarn --frozen-lockfile
         working-directory: cohttp-lwt-jsoo/test
 
+      - run: |
+          sudo apt update
+          sudo apt upgrade
```
mseri, line 53:
```suggestion
```

I just made one more test, it works also without this step (and it is slightly faster)

mseri: I am testing if we can also drop the next one

bikallem: @mseri could we do these small tweaks on the main branch itself after this PR is merged?

mseri: Sure. I can probably also push in this branch myself. I am currently waiting a bit longer to see if I can get extra feedback

mseri:
I wanto to rebase and cleanup the history before merging this. This will take some time but I should be able to do it next week, please bear some more patience 🙏

mseri:
Thanks. Some commits, like 15261416fd19b6e4b915b1ef71330ebf2ecc6d06 and 43bed482756afabf2132bc8b2bd80223350c3f8e, contain changes to other packages that are unrelated with this PR (like cohttp-async, cohttp-curl* and the new cohttp server. As well as adding and removing .envrc. We should clean those away before merging. I'll try to do it in the next few days

cohttp-eio/src/reader.ml
```diff
@@ -0,0 +1,322 @@
...
+  match end_of_input t with
+  | true -> Stdlib.raise_notrace End_of_file
+  | false ->
+      let meth = p_meth t in
+      let resource = p_resource t in
+      let version = p_version t in
+      let headers = http_headers t in
+      let encoding = Http.Header.get_transfer_encoding headers in
+      commit t;
+      { Http.Request.meth; resource; version; headers; scheme = None; encoding }
```
mseri, line 322: Since we were talking about scheme, why here we set it to `None`? We don't have a way to populate it?
Is the warning due to the use of encoding?

mseri:
I have cleaned the commits from the spurious changes inherited with the merges. Just had a final question, for the rest I am waiting for the CI to make sure I did not screw up anything

cohttp-eio/src/reader.ml
bikallem: > Since we were talking about scheme, why here we set it to None? We don't have a way to populate it?

I am not sure if there is a reliable way to determine a value for the scheme. For example, a web app using cohttp-eio could be behind a proxy server where the ssl/tls terminate or it could be a normal socket connection without ssl/tls.

> Is the warning due to the use of encoding?

Yes, dune gives a build error during dev without the attribute.

mseri:
The tests on nix are failing with
```
File "cohttp-eio/src/dune", line 4, characters 12-15:
[35667](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35668)
4 |  (libraries eio cstruct http bigstringaf fmt))
[35668](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35669)
                ^^^
[35669](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35670)
Error: Library "eio" not found.
[35670](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35671)
Hint: try:
[35671](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35672)
  dune external-lib-deps --missing -j 1 @runtest
[35672](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35673)
File "cohttp-eio/tests/dune", line 6, characters 23-31:
[35673](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35674)
6 |  (libraries cohttp_eio eio_main fmt))
[35674](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35675)
                           ^^^^^^^^
[35675](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35676)
Error: Library "eio_main" not found.
[35676](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35677)
Hint: try:
[35677](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35678)
  dune external-lib-deps --missing -j 1 @runtest
[35678](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35679)
File "cohttp-eio/tests/dune", line 13, characters 23-31:
[35679](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35680)
13 |  (libraries cohttp_eio eio_main fmt))
[35680](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35681)
                            ^^^^^^^^
[35681](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35682)
Error: Library "eio_main" not found.
[35682](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35683)
Hint: try:
[35683](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35684)
  dune external-lib-deps --missing -j 1 @runtest
[35684](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35685)
File "cohttp-eio/tests/dune", line 22, characters 0-50:
[35685](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35686)
22 | (mdx
[35686](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35687)
23 |  (package cohttp-eio)
[35687](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35688)
24 |  (packages cohttp-eio))
[35688](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35689)
Error: Program ocaml-mdx not found in the tree or in PATH
[35689](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35690)
 (context: default)
[35690](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35691)
Hint: opam install mdx
[35691](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35692)
File "cohttp-eio/tests/dune", line 27, characters 9-25:
[35692](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35693)
27 |  (deps %{bin:test-server} %{bin:test-chunk-server} %{bin:crlf}))
[35693](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35694)
              ^^^^^^^^^^^^^^^^
[35694](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35695)
Error: Program test-server not found in the tree or in PATH
[35695](https://github.com/mirage/ocaml-cohttp/runs/6801084258?check_suite_focus=true#step:4:35696)
 (context: default)
 ```

bikallem:
I am not familiar with nix but I think nix needs to not run the cohttp-eio package and tests for now. If nix is to run on cohttp-eio then it needs to run a version of OCaml with effects and domains, 4.12.0+domains or 5.0.0+trunk. @rgrinberg

bikallem (commented):
shell-eio.nix only installs/provides system dependencies when working in nixos. It will not setup the required OCaml version (4.12.+domains or 5.0.0). For that I believe it needs a nix package to be submitted to nixpkgs repo in github. I am not familiar enough with nixpkgs to be able to package trunk OCaml. Best leave the eio part out of nix.yml for now.

mseri:
The last two commits should fix the CI, if also nix is green I will merge

mseri:
It is all green now. Time to take a stance a merge. The next cohttp release will be a alpha o beta release anyway, so extra changes will have time to be made

mseri:
Thanks a lot for the patience and the effort

PR #858, opened 2022-02-02
Cookies: case-insensitive check of set-cookies

### LINKED ISSUE #855: Cookie extraction should use case-insensitive comparison for headers
The HTTP headers are not always lower cased. For instance, when working with the mediawiki Docker container I'm getting a `Set-Cookie` field instead of `set-cookie`, which `Cohttp.Cookie.Set_cookie_hdr.extract` assumes, and this makes a difference:
```ocaml
utop # #require "cohttp";;
utop # Cohttp.Header.of_list ["set-cookie", "my_wiki_session=bq02atpavmnv00b2mn6j62sc5li43cu3; path=/; HttpOnly"] |> Cohttp.Cookie.Set_cookie_hdr.extract;;
- : (string * Cohttp.Cookie.Set_cookie_hdr.t) list =
[("my_wiki_session",
  {Cohttp.Cookie.Set_cookie_hdr.cookie =
    ("my_wiki_session", "bq02atpavmnv00b2mn6j62sc5li43cu3");
   expiration = `Session; domain = None; path = Some "/"; secure = false;
   http_only = true})]
utop # Cohttp.Header.of_list ["Set-Cookie", "my_wiki_session=bq02atpavmnv00b2mn6j62sc5li43cu3; path=/; HttpOnly"] |> Cohttp.Cookie.Set_cookie_hdr.extract;;
- : (string * Cohttp.Cookie.Set_cookie_hdr.t) list = []
```
I am thinking it could be passed through `String.lowercase_ascii`, but I'm not sure if the best place would be when parsed, on-demand in the iterators, or on the consumer side (in `exctract`).

I'm using Cohttp 5.0.0.

mseri:
Should fix #855
I am not looking into fixing the rest of the cookies module right now

TODO:

- [ ] tests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,14 +17,15 @@
     when deciding whether to use keep-alive. This allows a handler to decide to
     close a connection even if the client requested a keep-alive in the
     request.
-- async(server): allow creating a server without using conduit (anuragsoni, #839)
+- async(server): allow creating a server without using conduit (anuragsoni #839)
   + Add `Cohttp_async.Server.Expert.create` and
     `Cohttp_async.Server.Expert.create_with_response_action`that can be used to
     create a server without going through Conduit. This allows creating an
     async TCP server using the Tcp module from `Async_unix` and lets the user
     have more control over how the `Reader.t` and `Writer.t` are created.
-- http(header): faster `to_lines` and `to_frames` implementation (mseri, #847)
+- http(header): faster `to_lines` and `to_frames` implementation (mseri #847)
 - ci: add changelog check (mseri #850)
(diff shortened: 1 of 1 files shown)


cohttp/src/cookie.ml
```diff
@@ -135,11 +135,12 @@ module Set_cookie_hdr = struct
 
   (* TODO: check dupes+order *)
   let extract hdr =
+    let (=) = Http.Header.Private.caseless_equal in
```
rgrinberg, line 138: Can you inline this definition? It's confusing to read as `=` usually refers to polymorphic comparison

rgrinberg (approved):

PR #860, opened 2022-02-09
chore: add bench for cohttp-lwt-unix

rgrinberg:
@anuragsoni I'm benchmarking this and getting the same issues as #838 wrt to perf.

### DIFF
--- a/cohttp-lwt-unix/bench/bench.ml
+++ b/cohttp-lwt-unix/bench/bench.ml
@@ -1,0 +1,20 @@
+open Cohttp_lwt_unix
+
+let handler _ req _ =
+  let uri = Cohttp.Request.uri req in
+  match Uri.path uri with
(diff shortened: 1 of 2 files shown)


mseri:
Why not updating https://github.com/mirage/ocaml-cohttp/blob/master/bench/lwt_unix_server.ml ?

mseri:
I had added all of them in the same bench folder to contain the "bench" optional package

anuragsoni:
> I'm benchmarking this and getting the same issues as #838 wrt to perf.

The perf numbers for the conduit based lwt server and the one in #838 were very similar in my tests as well. I think as first seen on the parser PR, we hit an upper bound with the current `Lwt_io` based server runtime where beyond a certain throughput (it was 80,000 RPS on my laptop) the latency numbers degrade to be greater than multiple seconds. I had to drop Lwt_io and use a different buffering strategy (using the byte buffer + Lwt_unix.file_descr) to see further improvements.

PR #861, opened 2022-02-09
simplify server read

rgrinberg:


cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
```diff
@@ -98,11 +98,7 @@ module Input_channel = struct
   type t = { buf : Bytebuffer.t; chan : Lwt_io.input_channel }
 
   let refill ic buf ~pos ~len =
-    Lwt.catch
-      (fun () ->
-        Lwt_io.read_into ic buf pos len >|= fun c ->
-        if c > 0 then `Ok c else `Eof)
-      (function Lwt_io.Channel_closed _ -> Lwt.return `Eof | exn -> raise exn)
+    Lwt_io.read_into ic buf pos len >|= fun c -> if c > 0 then `Ok c else `Eof
```
mseri, line 101: Are we sure that we will not get unexpected `Channel_closed _` failures?

rgrinberg: Yes indeed this is still possible

PR #864, opened 2022-03-28
Use package core_unix in examples

bcc32:
Similar to #791, but for the examples directory, which is a dependency
of @runtest for cohttp-async.

Signed-off-by: Aaron L. Zeng <me@bcc32.com>


mseri:
Thanks

PR #865, opened 2022-03-29
chore: update to latest cmdliner

rgrinberg:


mseri:
I am going to merge but we need to figure out how to fix the ci, both nix and the macos builders are broken

PR #866, opened 2022-04-16
Add missing dependency on Unix

dra27:
Silences the alert proposed in ocaml/ocaml#11198, but this PR can be merged regardless.


mseri:
Thanks

Issue #633, opened 2018-09-04
Build error on Arch Linux


basus:
I'm getting the following error when trying to build `cohttp-async` on Arch Linux with OCaml 4.06.1:

```
=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] The compilation of cohttp-async failed at "jbuilder build -p cohttp-async -j 1".

#=== ERROR while installing cohttp-async.1.0.2 ================================#
# opam-version 1.2.2
# os           linux
# command      jbuilder build -p cohttp-async -j 1
# path         /home/basus/.opam/4.06.1/build/cohttp-async.1.0.2
# compiler     4.06.1
# exit-code    1
# env-file     /home/basus/.opam/4.06.1/build/cohttp-async.1.0.2/cohttp-async-691-1872ec.env
# stdout-file  /home/basus/.opam/4.06.1/build/cohttp-async.1.0.2/cohttp-async-691-1872ec.out
# stderr-file  /home/basus/.opam/4.06.1/build/cohttp-async.1.0.2/cohttp-async-691-1872ec.err
### stderr ###
# [...]
# collect2: error: ld returned 1 exit status
# File "caml_startup", line 1:
# Error: Error during linking
#     ocamlopt cohttp-async/bin/cohttp_server_async.exe (exit 2)
[...]

PR #869, opened 2022-04-23
async: add upper bound on core_unix when testing

mseri:
The deprecation of Core_unix otherwise breaks the CI

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,7 @@
   + This new implementation does not depend on conduit and has a simpler and
     more flexible API
 - async: Adapt cohttp-curl-async to work with core_unix.
+- async: Add upper bound to core_kernel when testing to fix the CI
 - *Breaking changes*
   + refactor: move opam metadata to dune-project (rgrinberg #811)
   + refactor: deprecate Cohttp_async.Io (rgrinberg #807)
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -33,6 +33,7 @@
   + This new implementation does not depend on conduit and has a simpler and
     more flexible API
 - async: Adapt cohttp-curl-async to work with core_unix.
+- async: Add upper bound to core_kernel when testing to fix the CI
```
rgrinberg, line 36: Do we usually include test related changes?

rgrinberg (approved):

CHANGES.md
mseri: Sometimes we did, sometimes not.
In any case I just noticed that this is fixed by #633

mseri: I don't know if this PR is still needed, I'll try to test locally first

rgrinberg: I guess I'm more used to thinking about the change log as something user facing. So I would prefer to leave uninteresting information out.

mseri: This was user-facing though, since if you install with or without tests, it changes the dependencies

mseri:
No longer needed, thanks to the better fix in #633

PR #870, opened 2022-04-23
chore: update nix build plan

rgrinberg:
disable ocamlformat until it's available in nixpkgs

PR #871, opened 2022-04-24
Refactor Github Actions to avoid SIGKILL on Mac OS and speed up tests

mseri:
The `-j1` is needed to avoid incurring in what seems a macos bug (see the comments in https://github.com/mirage/ocaml-cohttp/pull/870 for more details).

Spelling out all packages in the macos tests is done to avoid having to rebuild and reinstall `cohttp-lwt-jsoo` and all the relevant javascript packages. We will need to to it anyways soon for all CI jobs to have usable tests for `cohttp-eio`, see https://github.com/mirage/ocaml-cohttp/pull/857


mseri:
I have re-run the tests a dozen of times on my repository. It turns out that `jobs=1` reduces the frequency of failures but does not eliminate it...

mseri:
If I disable the dune cache on the macos worker, I can run the CI without the job limitation and was not able to re-trigger the failures. This happens to be faster then using j=1 with the cache enabled

PR #875, opened 2022-04-26
chore: change log cleanup

rgrinberg:
The change log is for end users for the library. Therefore, we should
only keep stuff that is relevant to them.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,17 +1,17 @@
 ## Unreleased
-- Remove unused values to help ci(@bikallem #873)
 - New curl based clients (@rgrinberg #813)
   + cohttp-curl-lwt for an Lwt backend
   + cohttp-curl-async for an Async backend
 - Completely new Parsing layers for servers (@anuragsoni #819)
   + Cohttp now uses an optimized parser for requests.
-  + The new parser produces much less temporary buffers during read operations in servers.
+  + The new parser produces much less temporary buffers during read operations
+    in servers.
 - Faster header comparison (gasche #818)
-- Introduce http package containing common signatures and structures useful for compatibility with cohttp - and no dependencies (rgrinberg #812)
-- Faster Request/Response comparison (rgrinberg #814)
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,15 +1,14 @@
...
   + cohttp-curl-async for an Async backend
 - Completely new Parsing layers for servers (@anuragsoni #819)
   + Cohttp now uses an optimized parser for requests.
-  + The new parser produces much less temporary buffers during read operations in servers.
+  + The new parser produces much less temporary buffers during read operations
+    in servers.
 - Faster header comparison (gasche #818)
-- Introduce http package containing common signatures and structures useful for compatibility with cohttp - and no dependencies (rgrinberg #812)
-- Faster Request/Response comparison (rgrinberg #814)
-- Use raise_notrace in header short circuiting exceptions (rgrinberg #802)
```
mseri, line 12: Why not leaving the optimizations?

rgrinberg: You mean the one about `raise_notrace`? Honestly it's so minor it's hardly worth mentioning. If you'd like, I can bring it back though.

mseri: I meant both,
```
- Faster Request/Response comparison (rgrinberg #814)
- Use raise_notrace in header short circuiting exceptions (rgrinberg #802)
```
but it could also go, more compactly below

CHANGES.md
```diff
@@ -1,15 +1,14 @@
...
+- Introduce http package containing common signatures and structures useful for
+  compatibility with cohttp - and no dependencies (rgrinberg #812)
 - async(server): allow reading number of active connections (anuragsoni #809)
 - Various internal refactors (rgrinberg, mseri #820, #800, #799, #797)
```
mseri, line 13: For example added here:
```suggestion
- Various internal refactors and optimizations (rgrinberg, mseri #820, #814, #802, #800, #799, #797)
```

rgrinberg: Done

PR #876, opened 2022-04-27
Fix ci build master

bikallem:
Address some more 32 warnings.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -45,8 +45,6 @@ jobs:
 
       - run: opam exec -- make build
 
-      - run: opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
-
       - run: opam exec -- make test
 
       - run: opam exec -- make js-test
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -45,6 +45,8 @@ jobs:
 
       - run: opam exec -- make build
 
+      - run: opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
+
       - run: opam exec -- make test
 
       - run: opam exec -- make js-test
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## Unreleased
-- Remove unused values to help ci(@bikallem #873)
+- Address warning 32 errors (@bikallem #876)
+- Remove unused values. (@bikallem #873)
 - New curl based clients (@rgrinberg #813)
   + cohttp-curl-lwt for an Lwt backend
   + cohttp-curl-async for an Async backend
--- a/cohttp-lwt-unix/test/test_parser.ml
+++ b/cohttp-lwt-unix/test/test_parser.ml
@@ -44,15 +44,6 @@ let basic_res_content =
    \r\n\
    home=Cosby&favorite+flavor=flies"
 
-let post_req =
-  "POST /path/script.cgi HTTP/1.0\r\n\
-   From: frog@jmarshall.com\r\n\
-   User-Agent: HTTPTool/1.0\r\n\
-   Content-Type: application/x-www-form-urlencoded\r\n\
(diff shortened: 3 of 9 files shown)


bikallem:
@mseri I tracked the issue with `dune build` giving error/warning 32 in ci builds. It seems dune 3.0/3.1 raises this error while dune < 2.9 seems to not surface this error. eio-3 branch uses dune version 3.0 which is why the ci was failing in the branch but passing in the master. This PR addresses this issue with dune version and removes all instances of warning 32 in the code base.

.github/workflows/workflow.yml
```diff
@@ -45,6 +45,8 @@ jobs:
 
       - run: opam exec -- make build
 
+      - run: opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
```
rgrinberg, line 48: What additional targets does this add over `make build`?

CHANGES.md
```diff
@@ -1,4 +1,5 @@
 ## Unreleased
+- Address warning 32 errors in ci (@bikallem #876)
```
rgrinberg, line 2: We don't to describe CI changes in the change log.

.github/workflows/workflow.yml
bikallem: removed.

CHANGES.md
bikallem: removed references to ci

cohttp/test/test_header.ml
```diff
@@ -438,11 +431,6 @@ let link_ext_star () =
       ]
     (H.get_links headers)
 
-let trim_ws () =
```
mseri, line 441: Why not keeping this test and running it instead?

examples/lwt_unix_doc/docker_lwt.ml
```diff
@@ -1,14 +1,6 @@
 open Lwt.Infix
 open Cohttp_lwt_unix
 
-let ctx =
```
mseri, line 4: We should leave this here and instead replace `t` with
```ocaml
let t =
  Cohttp_lwt_unix.Client.get ~ctx (Uri.of_string "http://docker/version")
  >>= fun (resp, body) ->
  let open Cohttp in
  let code = resp |> Response.status |> Code.code_of_status in
  Printf.printf "Response code: %d\n" code;
  Printf.printf "Headers: %s\n" (resp |> Response.headers |> Header.to_string);
  body |> Cohttp_lwt.Body.to_string >|= fun body ->
  Printf.printf "Body of length: %d\n" (String.length body);
  print_endline ("Received body\n" ^ body)
```
This other body was probably an oversight.
Or at least add a flag to select one of the other behaviour.

These examples were meant to build the examples in the README.
There is a ticket to move to `mdx`, but until that is implemented, we should try and keep them for now.

cohttp/test/test_header.ml
bikallem: Done. Added to the test suite.

examples/lwt_unix_doc/docker_lwt.ml
bikallem: Done.

cohttp-lwt-unix/test/test_parser.ml
```diff
@@ -44,15 +44,6 @@ let basic_res_content =
...
    home=Cosby&favorite+flavor=flies"
 
-let post_req =
-  "POST /path/script.cgi HTTP/1.0\r\n\
-   From: frog@jmarshall.com\r\n\
-   User-Agent: HTTPTool/1.0\r\n\
-   Content-Type: application/x-www-form-urlencoded\r\n\
-   Content-Length: 32\r\n\
-   \r\n\
-   home=Cosby&favorite+flavor=flies"
```
rgrinberg, line 54: What about parsing this request as well?

rgrinberg:
@smorimoto there's some shared cache failures. could you take a look? The shared cache lives on a different FS so hardlinks can't work. We need to either copy the artifacts or use the same FS.

rgrinberg:
Merged.

PR #877, opened 2022-05-10
Require higher `crowbar` version

Leonidas-from-XIV:
As described in https://github.com/ocaml/opam-repository/pull/21344#discussion_r868912162 the signature of `Crowbar.range` changed between 0.1 and 0.2 and Cohttp requires the newer interface.


mseri:
Thanks

PR #878, opened 2022-06-08
http: add Request.make

bikallem:
Add `Request.make` so that we don't have to adorn `[@warning "-3"]` when
creating `Request.t`.

### DIFF
--- a/cohttp/src/request.ml
+++ b/cohttp/src/request.ml
@@ -49,15 +49,9 @@ let scheme t = t.scheme
 let resource t = t.resource
 let version t = t.version
 let encoding t = t.encoding
-let fixed_zero = Transfer.Fixed Int64.zero
 
-let guess_encoding ?(encoding = fixed_zero) headers =
-  match Header.get_transfer_encoding headers with
-  | Transfer.(Chunked | Fixed _) as enc -> enc
-  | Unknown -> encoding
-
-let make ?(meth = `GET) ?(version = `HTTP_1_1) ?encoding ?headers uri =
-  let headers = match headers with None -> Header.init () | Some h -> h in
+let make ?(meth = `GET) ?(version = `HTTP_1_1) ?(encoding = Transfer.Unknown)
+    ?(headers = Header.init ()) uri =
   let headers =
     Header.add_unless_exists headers "host"
       (match Uri.scheme uri with
@@ -81,15 +75,14 @@ let make ?(meth = `GET) ?(version = `HTTP_1_1) ?encoding ?headers uri =
         Header.add_authorization headers auth
     | _, _, _ -> headers
   in
-  let encoding = guess_encoding ?encoding headers in
-  {
-    meth;
-    version;
-    headers;
-    scheme = Uri.scheme uri;
-    resource = Uri.path_and_query uri;
-    encoding;
-  }
+  let scheme = Uri.scheme uri in
+  let resource = Uri.path_and_query uri in
+  let encoding =
+    match Header.get_transfer_encoding headers with
+    | Transfer.Unknown -> encoding
+    | encoding -> encoding
+  in
+  { headers; meth; scheme; resource; version; encoding }
 
 let is_keep_alive t = Http.Request.is_keep_alive t
 
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -763,6 +763,11 @@ module Request = struct
     | `DELETE | `POST | `PUT | `PATCH | `OPTIONS | `Other _ ->
         Transfer.has_body req.encoding
 
+  let make ?(meth = `GET) ?(version = `HTTP_1_1) ?(headers = Header.empty)
+      ?scheme resource =
+    let encoding = Header.get_transfer_encoding headers in
+    { headers; meth; scheme; resource; version; encoding }
+
   let pp fmt t =
     let open Format in
     pp_open_vbox fmt 0;
@@ -800,12 +805,8 @@ module Response = struct
     | i -> i
 
   let make ?(version = `HTTP_1_1) ?(status = `OK) ?(flush = false)
-      ?(encoding = Transfer.Chunked) ?(headers = Header.empty) () =
-    let encoding =
-      match Header.get_transfer_encoding headers with
-      | Transfer.(Chunked | Fixed _) as enc -> enc
-      | Unknown -> encoding
-    in
+      ?(headers = Header.empty) () =
+    let encoding = Header.get_transfer_encoding headers in
     { encoding; headers; version; flush; status }
 
   let headers t = t.headers
--- a/http/src/http.mli
+++ b/http/src/http.mli
@@ -349,7 +349,12 @@ module Header : sig
 
   val get_content_range : t -> Int64.t option
   val get_connection_close : t -> bool
+
   val get_transfer_encoding : t -> Transfer.encoding
+  (** [get_transfer_encoding h] checks the "content-length", "content-range" and
+      "transfer-encoding" headers to infer the transfer encoding. Uses Unknown
+      if nothing is found.*)
+
   val add_transfer_encoding : t -> Transfer.encoding -> t
   val connection : t -> [ `Keep_alive | `Close | `Unknown of string ] option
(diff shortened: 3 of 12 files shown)


http/src/http.ml
```diff
@@ -755,6 +755,15 @@ module Request = struct
     | `GET | `HEAD | `CONNECT | `TRACE -> `No
     | `DELETE | `POST | `PUT | `PATCH | `OPTIONS | `Other _ ->
         Transfer.has_body req.encoding
+
+  let[@warning "-3"] make ?(meth = `GET) ?(version = `HTTP_1_1)
```
mseri, line 759: This seems a regression forcing the scheme to be `None` and ignoring all the changes in https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/src/request.ml#L59 

Why does the warning 3 happen? You can get rid of it just by usign Cohttp.Private.Make instead of Cohttp.Make

bikallem: Ah, I wasn't aware there was make in `cohttp`.  That function is what I was after. 

I am using `http` from eio client/server so am not using `cohttp` package. Should the make in https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/src/request.ml#L59 be moved/copied to `http` package itself since it is useful to users who consume `http` only?

mseri: Ping @rgrinberg

bikallem: @mseri Ah sorry, I hadn't refreshed the browser window when you made the comment above so didn't see it before posing mine. I have now made the `make` even more simple, i.e. not even trying to guess the `encoding` since that attribute is marked as deprecated.

mseri: Like we do for response: https://github.com/mirage/ocaml-cohttp/blob/b2c3198658fda65cab7f5275bcd2ea329a146b9f/http/src/http.ml#L783

bikallem: Done.

http/src/http.ml
```diff
@@ -755,6 +755,10 @@ module Request = struct
...
```
mseri, line 759:
```suggestion
  let make ?(meth = `GET) ?(version = `HTTP_1_1) 
```

http/src/http.ml
```diff
@@ -755,6 +755,10 @@ module Request = struct
     | `GET | `HEAD | `CONNECT | `TRACE -> `No
     | `DELETE | `POST | `PUT | `PATCH | `OPTIONS | `Other _ ->
         Transfer.has_body req.encoding
+
+  let[@warning "-3"] make ?(meth = `GET) ?(version = `HTTP_1_1) 
+      ?(headers = Header.empty) ?scheme resource =
+    { headers; meth; scheme; resource; version; encoding = Transfer.Unknown }
```
mseri, line 761: I would still keep the encoding and document it as in `Response.Header`, we will remove it once we remove the field:

```
has an encoding value
      determined from the content of [headers] or if no proper header is
      present, using the value of [encoding]. 
```

mseri: Just so that we are on the same page, this is what I am talking about: https://github.com/mirage/ocaml-cohttp/blob/b2c3198658fda65cab7f5275bcd2ea329a146b9f/http/src/http.mli#L417-L429
I think it is more consistent if we do the same thing in both places, and we should let the deprecation stay until the next next version

mseri:
The CI seems to be stuck right now, anyways it will fail badly because the javascript tests are completely broken. One failure, with `conf-npm` seems to be due to incompatible versions of node and npm, the other is about not finding the library for pretend xmlrpc. I thin it may make sense to disable them for now. I will do it in a separate PR later on, after merging eio

http/src/http.ml
```diff
@@ -781,12 +790,7 @@ module Response = struct
     | i -> i
 
   let make ?(version = `HTTP_1_1) ?(status = `OK) ?(flush = false)
```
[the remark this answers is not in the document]
bikallem: Argh, my bad. Seems to be a holdover from an earlier attempt at simplification. It is reverted now.

bikallem:
I followed mainly `Response.make` for the default value of `encoding`. However, shouldn't the default value of encoding be `Transfer.Unknown` rather than `Transfer.Chunked` ?

mseri:
I don't know. In `cohttp` [we actually use `Transfer.Fixed Int64.zero` as default](https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/src/request.ml#L54). For the response `Chunked` was a safe option, not knowing the body. I think we should stick to backward compatibility in this case (encoding will be removed in any case soon)

http/src/http.ml
```diff
@@ -755,6 +755,15 @@ module Request = struct
     | `GET | `HEAD | `CONNECT | `TRACE -> `No
     | `DELETE | `POST | `PUT | `PATCH | `OPTIONS | `Other _ ->
         Transfer.has_body req.encoding
+
+  let make ?(meth = `GET) ?(version = `HTTP_1_1) ?(encoding = Transfer.Chunked)
```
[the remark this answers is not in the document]
mseri: To keep compatibility with https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/src/request.ml#L59

http/src/http.mli
```diff
@@ -383,6 +383,23 @@ module Request : sig
...
+    string ->
+    t
+  (** [make resource] is [t]. The default values if not specified are as
+      follows:
+
+      - [meth] is [`GET]
+      - [version] is [`HTTP_1_1]
+      - [headers] is [Header.empty]
+      - [encoding] is [Transfer.Chunked]
+      - [scheme] is [None] *)
```
mseri, line 402:
```suggestion
      - [scheme] is [None] 
      
      The request has an encoding value determined from the content of [headers]
      or if no proper header is present, using the value of [encoding]. Checked headers
      are "content-length", "content-range" and "transfer-encoding". The default
      value of [encoding] is `fixed 0`.
      *)
```

mseri:
I am going to make one more change and push over it. Then let's see also what @rgrinberg thinks about it

bikallem:
@rgrinberg is this PR in a state to be merged?

mseri:
We have discussed this in the afternoon. I am making some further changes before it can be merged, essentially we will drop encoding from the constructor, document its behaviour and update cohttp to use it.

cohttp/src/request.ml
```diff
@@ -77,7 +78,12 @@ let make ?(meth = `GET) ?(version = `HTTP_1_1) ?encoding ?headers uri =
   in
   let scheme = Uri.scheme uri in
   let resource = Uri.path_and_query uri in
-  Http.Request.make ~meth ~version ~headers ?scheme ?encoding resource
+  let headers =
+    match Header.get_transfer_encoding headers with
+    | Transfer.(Chunked | Fixed _) -> headers
+    | Unknown -> Header.add_transfer_encoding headers encoding
```
mseri, line 84: I kept `Cohttp.{Request,Response}.make` backward compatible, @rgrinberg do you reckon we should just drop encoding also from here?

mseri:
For some reason that I don't really understand, the `cohttp-lwt-unix` tests seem to deadlock... or at least they get stuck forever. Any clue?

mseri:
Ok, I made it as small as possible. The `cohttp` functions `{Request,Response}.make` remain backward compatible, and the `Http` ones are saner and drop the encoding parameter. The code is updated in a backward-compatible way, the only thing that needed fixing was one test, since now the Response encoding defaults to `Unknown` so the responses need to explicitly set a encoding header if they use the dumb `Http` constructor (or they need to rely to the old one in `Cohttp`)

mseri:
I'd like both your opinions on this one before merging.
I agree with @rgrinberg on having the `http` constructors as dumb as possible and do nothing fancy, but I could not modify the `cohttp` ones in a backward compatible way since it would require changing the headers and did end up in some weird corner cases (like response 304 or 1xx/204 having a `Fixed 0` transfer-encoding instead of `Unknown`)

rgrinberg (approved):

bikallem:
LGTM.

PR #879, opened 2022-06-09
cohttp-eio Client module

bikallem:
This the client implementation of cohttp eio backend. Roughly, the PR aims to achieve the following (some are optional and/or outside the remit of this PR):

- [x] cohttp-eio client module (`Cohttp_eio.Client`)
- [x] add client tests, (add/update server tests)
- [x] ~domain name resolving using [ocaml-dns](https://github.com/mirage/ocaml-dns) (https://github.com/mirage/ocaml-dns/pull/312)~
- [x] ~TLS support so that client can connect to server which mandate `https` only connection (optional)~

A couple of tasks above are optional since the PRs/dependencies are in other repositories. I will update the task with PRs and status as they are merged/accepted.

EDIT: ocaml-tls and ocaml-dns for cohttp-eio client tls and domain name resolution functionality will be resolved in future PRs when respective PRs to support eio backend are merged.


cohttp-lwt-jsoo/test/src/dune
```diff
@@ -11,5 +11,5 @@
  (action
   (setenv
    NODE_PATH
-   "%{project_root}/cohttp-lwt-jsoo/test/node_modules"
+   "%{project_root}/../../cohttp-lwt-jsoo/test/node_modules"
```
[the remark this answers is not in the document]
bikallem: To compensate for the dune path computation on version > 2.9, I think.

bikallem:
>Please rebase it on master. There seem to be spurious changes from an old eio-3 branch (like the shell-eio..nix file

Do you have a suggestion on how to go about the rebase? I tried `git rebase master` and there were lots of conflicts which I couldn't get my head around. So I just merged master on the branch.

mseri:
I can probably di it in a couple of days if that is ok.

I don’t know if it is the best way, but I usually do it manually by running interactive rebase. For each commit I reset, stage the relevant changes (for partial chunks, should not be the case here, there is `gjt add -p`). Then I drop everything else with `git restore .` and continue the rebase

bikallem:
I managed to rebase on master and remove all the unnecessary commits. `git rebase -i master` seems to be the trick in this case.

patricoferris (commented):
Thanks for this @bikallem, I'm using a slightly modified version over in https://github.com/geocaml/carbon-intensity :))

cohttp-eio/src/client.ml
```diff
@@ -0,0 +1,105 @@
...
+      let () = (reason_phrase *> crlf *> return ()) reader in
+      let headers = http_headers reader in
+      commit reader;
+      Http.Response.make ~version ~status ~headers ()
+
+(* Generic HTTP call *)
+
+let call ?(meth = `GET) ?(version = `HTTP_1_1) ?(headers = Http.Header.init ())
+    ?(body = Body.Empty) env sw stream uri =
+  let sock = Net.connect ~sw (Stdenv.net env) stream in
```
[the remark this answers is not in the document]
bikallem: This is done in the latest commit.

bikallem:
@mseri This is ready for review.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -90,3 +92,75 @@ module Server : sig
...
+    Uri.t ->
+    response
+  (** [body_disallowed_call] denotes HTTP client calls where a request is not
+      allowed to have a request body. *)
+
+  type 'a body_allowed_call =
+    ?version:Http.Version.t ->
+    ?headers:Http.Header.t ->
+    ?body:Body.t ->
+    (< env ; .. > as 'a) ->
```
patricoferris, line 115: Does it need to be passed as an object? I meant that the argument would be of type `Eio.Net.t`

bikallem: Yes, at the moment, the object just needs `net`. However, this object will likely have `sys_clock`, `mono_clock` and `secure_random` to accomodate ocaml-dns, ocaml-tls dependency requirements.

talex5 (commented):
At the moment, `get`, `post`, etc take a URI and an open-ended set of resources that might be needed to connect to it. This is not very flexible, since it requires cohttp to define all possible connection methods. I think we should split this into two parts:

1. A plain HTTP API where you pass in an already-connected `Eio.Flow.two_way` (plus the URI, for e.g. the "host" header).
2. A high-level API for resolving URIs to resources. This will know about networks, TLS, DNS, etc. It might follow redirects too.

It's not clear to me that (2) should even be in this library, as it should also handle other protocols too (`file://`, `ftp://`, `gemini://`, etc). In Python, the `urllib.request` module handles that.

Perhaps we should put a `Resolver` in Eio, and let cohttp populate it, e.g. something like:

```ocaml
let r = Eio.Resolver.make () in
Resolver.register r "http"  @@ Cohttp_eio.http_resolver env;
Resolver.register r "https" @@ Cohttp_eio.https_resolver env;
let data = Resolver.resolve r uri in
```

There could also be a convenience function to do this in one go, of course.

bikallem:
@mseri The CI Mac failing seems not related to this PR.

bikallem:
@talex5 Eio.Client makes do with `Eio.Flow.two_way` now.

talex5 (commented):
Thanks! I made a few comments on the interface inline.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -90,3 +92,74 @@ module Server : sig
...
+  type resource_path = string
+  (** Represents the HTTP resource path. *)
+
+  type 'a body_disallowed_call =
+    ?version:Http.Version.t ->
+    ?headers:Http.Header.t ->
+    Eio.Switch.t ->
+    Eio.Flow.two_way ->
+    resource_path ->
+    response
```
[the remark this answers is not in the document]
bikallem: Done

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -90,3 +92,74 @@ module Server : sig
...
+
+  (** {1 Generic HTTP call} *)
+
+  val call :
+    ?meth:Http.Method.t ->
+    ?version:Http.Version.t ->
+    ?headers:Http.Header.t ->
+    ?body:Body.t ->
+    Eio.Switch.t ->
+    Eio.Flow.two_way ->
```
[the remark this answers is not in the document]
bikallem: Done

cohttp-eio/examples/client1.ml
```diff
@@ -0,0 +1,25 @@
...
+    let he = Unix.gethostbyname host in
+    let addr = `Tcp (Eio_unix.Ipaddr.of_unix he.h_addr_list.(0), port) in
+    (Net.connect ~sw (Stdenv.net env) addr :> Eio.Flow.two_way)
+  in
+  (flow, path)
+
+let () =
+  Eio_main.run @@ fun env ->
+  Switch.run @@ fun sw ->
+  let flow, path = connect_info env sw "http://www.reddit.com/" in
```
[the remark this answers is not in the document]
bikallem: Updated the example.

cohttp-eio/src/client.ml
```diff
@@ -0,0 +1,107 @@
...
+      let version = Parser.(version <* space) buf_read in
+      let status = Parser.(status_code <* space) buf_read in
+      let () = Parser.(reason_phrase *> crlf *> return ()) buf_read in
+      let headers = Parser.http_headers buf_read in
+      Http.Response.make ~version ~status ~headers ()
+
+(* Generic HTTP call *)
+
+let call ?(meth = `GET) ?(version = `HTTP_1_1) ?(headers = Http.Header.init ())
+    ?(body = Body.Empty) sw flow uri =
```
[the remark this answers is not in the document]
bikallem: Done. This PR now use `Eio.Buf_write` module to replace home grown `Writer` module. As a result of which `sw` argument is no longer required.

cohttp-eio/examples/client1.ml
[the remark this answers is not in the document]
bikallem: It seems the `Host` header is quite important for HTPP clients to work correctly, therefore I think the library should add it. 
cc /@mseri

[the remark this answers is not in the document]
bikallem: Client now add `Host` header to outgoing requests.

mseri:
There is some race in the tests:
```
+++ b/_build/.sandbox/fcaf2ed2da7147d68504b5c372cfd680/default/cohttp-eio/tests/test_client.t.corrected
@@ -1,6 +1,7 @@
 Test Client.get
 
   $ test-server &
+  Fatal error: exception Eio_luv.Luv_error(EADDRINUSE) (* address already in use *)
   $ running_pid=$!
   $ test-client get
   meth: GET
@@ -9,10 +10,13 @@ Test Client.get
   headers: Header { Accept = "application/json"; Host = "localhost:8080" }
 
   $ kill ${running_pid}
+  kill: (44075) - No such process
+  [1]
 
 Test Client.post
 
   $ test-server &
+  Fatal error: exception Eio_luv.Luv_error(EADDRINUSE) (* address already in use *)
   $ running_pid=$!
   $ test-client post
   meth: POST
@@ -25,6 +29,8 @@ Test Client.post
   hello world!
 
   $ kill ${running_pid}
+  kill: (44098) - No such process
+  [1]
 
 Test Client.invalid_uri
```

cohttp-eio/src/client.ml
```diff
@@ -0,0 +1,127 @@
...
+  let host_name =
+    match Uri.host uri with
+    | Some h -> h
+    | None -> raise (invalid_arg "uri: host missing")
+  in
+  let scheme =
+    match Uri.scheme uri with
+    | Some "http" -> `HTTP
+    | Some "https" -> `HTTPS
+    | Some _ -> raise (invalid_arg "uri: invalid uri scheme")
```
mseri, line 78: In the current cohttp clients, you can override the resolvers to for example do http over sockets, see e.g. https://github.com/mirage/ocaml-cohttp#creating-custom-resolver-a-docker-socket-client-example can you do this here?

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -90,3 +92,94 @@ module Server : sig
 
   val not_found_handler : handler
 end
+
+module Client : sig
+  type response = Http.Response.t * Eio.Buf_read.t
```
mseri, line 97: To check the response headers or the status with this client I should do something like
```ocaml
ler res = Client.get "whatever" in
let status = Http.Response.status (fst res) in 
...
```
right? Can we add it to the example or the documentation?

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -90,3 +92,94 @@ module Server : sig
...
+  val patch : 'a body_allowed_call
+
+  (** {1 Response Body} *)
+
+  val read_fixed : response -> string option
+  (** [read_fixed (response,reader)] is [Some bytes], where [bytes] is of length
+      [n] if "Content-Length" header is a valid integer value [n] in [response].
+      [reader] is updated to reflect that [n] bytes was read. *)
+
+  val read_chunked : response -> (Body.chunk -> unit) -> Http.Header.t option
```
mseri, line 174: Can we add a default handler and also have a read function invoking the correct read with a default handler (a dumb one filling a buffer then returning is contents should do, we just need to document that it could hang forever).

mseri: To clarify my comment. I like the explicitness here but we start getting different incompatible APIs between the clients. It would be nice to have a `Body.read` for eio that just gives you the string, while keeping these two functions there or in a `Lowlevel` submodule.

bikallem: Chunked reading is quite different (behavior  and HTTP RFC spec) from fixed body reading; which is why they are a different/separate API. Just as an example, in chunked encoding we return a set of headers after chunk reading is done. However, this is not the case with `fixed` body read. 

Chunked read is also meant to be read in chunks so as to be memory efficient, therefore - except perhaps as a quick demo - I think using a buffer is not to be recommended. 

Finally, the cohttpe-eio client module keeps the read/write API consistent with the server where there is a separate read for fixed and chunked.

cohttp-eio/src/cohttp_eio.mli
bikallem: Ack. I will add it to the tests.

cohttp-eio/src/client.ml
bikallem: Do you specifically mean connecting to a UNIX domain socket? Otherwise the example client and tests already provides a demo of how to create a connection. In any case I think connecting to a domain socket would a good test case so I will add it to the test case.

mseri: Yes, I meant unix socket. Two common questions I had in the past are how to timeout requests and how to connect to unix domain sockets. Since with ocaml5+eio it is done differently, I think we should document it either in the interface, in an example to link to the README or in the README, so that we avoid answering the same questions all the time.

cohttp-eio/src/cohttp_eio.mli
mseri: I understand. As I said I lilke the lowlevel interface exposing the difference, and in the past we needed to add an expert mode to cohttp to actually be able to do something similar in a as backward compatible way as possible.

However, for most cases you get few bytes/kilobytes of body to read no matter the transfer encoding, and you just want to get it out as a string, so I think that we should have an inefficient read function, a sort of equivalent of `Cohttp.Body.read`, that hides this detail from you and justg gets the string for you, does not metter if it is not efficient (who wants efficiency can call read_chunked or read_fixed)

cohttp-eio/src/cohttp_eio.mli
bikallem: Added docker_client.ml which demonstrates the usage.

cohttp-eio/src/client.ml
bikallem: Added docker_client.ml which mimics the cohttp-lwt-unix example.

cohttp-eio/src/cohttp_eio.mli
bikallem: I can certainly understand the allure of such a `read` function. I myself tried to shoehorn the two read functions into one. I mean why have two `read` functions when one will do. However, the resulting `read` API turned out to be incorrect, imprecise and confusing(to the user). 

For example, 
1. `read_chunked` returns an updated set of headers as specified in https://datatracker.ietf.org/doc/html/rfc7230#section-4.1.3. However, this is not the case with `read_fixed` which just returns a string blob. We could recommend in the API documentation to ignore such returned value when the response is fixed body. However why do the very thing that you recommend against. `read_fixed` is the recommended API without any such adulteration. 
2. Each chunk may have chunk extensions (https://datatracker.ietf.org/doc/html/rfc7230#section-4.1.1) which can contain metadata about how such a chunk maybe interpreted/consumed. How should we interpret such an extension? One option could be to silently ignore such extensions and pretend they don't exist. However, IIUC OCaml eschews such practices and generally does not recommend it. 

The current API of `read_fixed` and `read_chunked` are the result of above consideration.

mseri: Then we should clearly document how to use it properly. Can you add to the documentation an example showing how to check which function to use and a simple example of use of both functions?

bikallem: Ack.

bikallem: Added demo/test for reading and writing chunk encoding.

writing chunk-encoding - https://github.com/bikallem/ocaml-cohttp/blob/eio-client/cohttp-eio/tests/test_client.ml#L39
reading chunk-encoding - https://github.com/bikallem/ocaml-cohttp/blob/eio-client/cohttp-eio/tests/test_client.ml#L90

cohttp-eio/examples/client1.ml
```diff
@@ -0,0 +1,16 @@
+open Eio
+open Cohttp_eio
+
+let conn env sw () =
+  let hostname, port = ("www.example.org", 80) in
+  let he = Unix.gethostbyname hostname in
+  let addr = `Tcp (Eio_unix.Ipaddr.of_unix he.h_addr_list.(0), port) in
+  let flow = (Net.connect ~sw (Stdenv.net env) addr :> Eio.Flow.two_way) in
```
talex5, line 8: Are these casts needed? Probably it indicates that the API is too strict if users need to do this.

cohttp-eio/src/body.ml
```diff
@@ -44,29 +45,24 @@ let pp_chunk fmt = function
...
-open Reader
-open Eio.Buf_read
-
 let read_fixed t headers =
-  match Http.Header.get headers "Content-length" with
-  | Some v ->
-      let content_length = int_of_string v in
-      let content = take content_length t in
-      content
-  | None -> raise @@ Invalid_argument "Request is not a fixed content body"
```
talex5, line 56: If there's no content-length then we should use `Buf_read.take_all` I think.

cohttp-eio/src/client.ml
```diff
@@ -0,0 +1,112 @@
...
+  Buf_read.take_while (function
+    | '\x21' .. '\x7E' | '\t' | ' ' -> true
+    | _ -> false)
+
+(* https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.2 *)
+let response buf_read =
+  let open Buf_read.Syntax in
+  match Buf_read.at_end_of_input buf_read with
+  | true -> Stdlib.raise_notrace End_of_file
+  | false ->
```
talex5, line 58: Is this needed? It will raise `End_of_file` anyway when you try to read the version.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -96,3 +100,94 @@ module Server : sig
...
+
+  type host = string * int option
+  (** Represents a server host domain name and port, e.g. www.example.org:8080,
+      www.reddit.com *)
+
+  type resource_path = string
+  (** Represents HTTP request resource path, e.g. "/shop/purchase",
+      "/shop/items", "/shop/categories/" etc. *)
+
+  type 'a conn = unit -> (host * Eio.Flow.two_way as 'a)
```
talex5, line 115: This type isn't actually polymorphic as written. Also, does it need to be a function? It looks like everything that takes it just calls the function once immediately at the start.
```suggestion
  type 'a conn = host * (#Eio.Flow.two_way as 'a)
```

cohttp-eio/src/cohttp_eio.mli
talex5: It seems that `read_fixed` could be considered as a special case of `read_chunked`, where there is only one chunk and there are no extra headers.

So, perhaps `read_chunked` should work on non-chunked bodies too, just calling the chunk handler once. Might need renaming (e.g. to just `read`).

cohttp-eio/src/rwer.ml
```diff
@@ -0,0 +1,65 @@
+(* This modules encapsulates refactored - common - readers and writers
+   used by the Client and Server modules.
+
+    rwer.ml => (R)eader (W)riter + er
+*)
+
+module Buf_read = Eio.Buf_read
+module Buf_write = Eio.Buf_write
+
+let return v _ = v
```
talex5, line 10: Note: we have `Eio.Buf_read.return` now.

cohttp-eio/src/server.ml
```diff
@@ -65,26 +60,44 @@ let internal_server_error_response =
...
+
+let resource =
+  let open Eio.Buf_read.Syntax in
+  Rwer.(take_while1 (fun c -> c != ' ') <* space)
+
+let[@warning "-3"] http_request t =
+  let open Eio.Buf_read.Syntax in
+  match Buf_read.at_end_of_input t with
+  | true -> Stdlib.raise_notrace End_of_file
+  | false ->
```
talex5, line 89: Don't think this does anything.

cohttp-eio/tests/test_client.ml
```diff
@@ -0,0 +1,124 @@
+module Net = Eio.Net
+module Stdenv = Eio.Stdenv
+module Switch = Eio.Switch
+open Cohttp_eio
+
+let conn env sw port () =
+  let addr = `Tcp (Net.Ipaddr.V4.loopback, port) in
+  let flow = (Net.connect ~sw (Stdenv.net env) addr :> Eio.Flow.two_way) in
+  let host = ("localhost", Some port) in
+  (host, flow)
```
talex5, line 10: Should be able to write just:
```suggestion
let conn env sw port =
  let addr = `Tcp (Net.Ipaddr.V4.loopback, port) in
  let flow = Net.connect ~sw (Stdenv.net env) addr in
  let host = ("localhost", Some port) in
  (host, flow)
```

cohttp-eio/tests/test_client.ml
```diff
@@ -0,0 +1,124 @@
...
+
+(* Read chunk and dump to a "client_chunks2.txt" *)
+let get_chunk env sw port =
+  let write_chunk_to_file flow chunk =
+    let data = Format.asprintf "%a\n\n" Body.pp_chunk chunk in
+    Eio.Flow.copy_string data flow
+  in
+  let res = Client.get (conn env sw port) "/get_chunk" in
+  let path = Eio.Path.(Stdenv.cwd env / "client_chunks2.txt") in
+  Eio.Path.with_open_out ~append:false ~create:(`Or_truncate 777) path
```
talex5, line 97:
```suggestion
  Eio.Path.with_open_out ~append:false ~create:(`Or_truncate 0o666) path
```

cohttp-eio/tests/test_server.ml
```diff
@@ -1,18 +1,93 @@
...
-let app (req, reader, _client_addr) =
+let app (req, reader, _) =
   match Http.Request.resource req with
-  | "/get" -> Server.text_response (Fmt.to_to_string Http.Request.pp req)
+  | "/get" ->
+      let buf = Buffer.create 0 in
+      let fmt = Format.formatter_of_buffer buf in
+      Http.Request.pp fmt req;
+      Format.fprintf fmt "%!";
+      Server.text_response (Buffer.contents buf)
```
talex5, line 10: What is this change for? It's much longer and does the same thing.

cohttp-eio/src/body.ml
bikallem: A missing `Content-Length` (if a client wants to send a fixed body request) represents a malformed request. Should `cohttp-eio` compensate for this scenario? I suppose if we don't care about HTTP pipe-lining  then `Buf_read.take_all` would work.

cohttp-eio/src/rwer.ml
bikallem: Done. Thanks.

cohttp-eio/src/server.ml
bikallem: It needs to raise End_of_file. I have made changes to `take_while1` to throw this error. So now can remove this here. I think this should save some 'ns' in the request processing hot loop.

Thanks. Well spotted.

cohttp-eio/src/client.ml
bikallem: Yes, This is removed now.

cohttp-eio/tests/test_client.ml
bikallem: Done.

cohttp-eio/tests/test_server.ml
bikallem: This is need to pretty print the request.

talex5: The output is identical for me. With `Server.text_response (Fmt.to_to_string Http.Request.pp req)` I get:
```
$ curl http://127.0.0.1:8080/get
meth: GET
resource: /get
version: HTTP/1.1
headers: Header {
 Accept = "*/*"; User-Agent = "curl/7.74.0"; Host = "127.0.0.1:8080" }
```
What do you get with the old version?

bikallem: It seems I misunderstood. Was the code before `Server.text_response (Fmt.to_to_string Http.Request.pp req)`? The latter is shorter, yes. Let me change that.

cohttp-eio/examples/client1.ml
bikallem: It doesn't need the cast now.

cohttp-eio/src/cohttp_eio.mli
bikallem: This is done now.

cohttp-eio/tests/test_server.ml
bikallem: Done.

cohttp-eio/tests/test_client.ml
bikallem: Done.

bikallem:
@talex5 `'a conn` is now `('a, 'b') conn`. Since https://github.com/ocaml-multicore/eio/pull/278 is merged, this should allow us to provide a default connection mechanism which takes uri as an argument, i.e. `val conn_of_uri: (Uri.t, 'b) conn`

bikallem:
@talex5 @mseri I believe I have addressed all outstanding issues. Could we please merge this if there aren't any further issues; It would help the PR to not bitrot😺.

mseri:
Can you update the changelog? @talex5 do you have further comments?

bikallem:
@mseri Added the changelog.

mseri (approved):

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -96,3 +100,80 @@ module Server : sig
...
+  type host = string * int option
+  (** Represents a server host - as ip address or domain name - and an optional
+      port value, e.g. www.example.org:8080, www.reddit.com *)
+
+  type resource_path = string
+  (** Represents HTTP request resource path, e.g. "/shop/purchase",
+      "/shop/items", "/shop/categories/" etc. *)
+
+  type ('a, 'b) conn = 'a -> (resource_path * host * #Eio.Flow.two_way as 'b)
+  (** [('a, 'b conn)] is [(resource_path, host, flow)]. [flow] is the Eio flow
```
talex5, line 116:
```suggestion
  (** [('a, 'b) conn] is [(resource_path, host, flow)]. [flow] is the Eio flow
```
However, this description does not match the type, which is still (strangely) a function.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -96,3 +100,80 @@ module Server : sig
...
+
+  (** {1 Generic HTTP call} *)
+
+  val call :
+    ?meth:Http.Method.t ->
+    ?version:Http.Version.t ->
+    ?headers:Http.Header.t ->
+    ?body:Body.t ->
+    ('a, 'b) conn ->
+    'a ->
```
talex5, line 146:
```suggestion
    'a conn ->
```
There is no point taking a function and its input here, as you just call it once immediately in all cases. It should just take the connection directly.

cohttp-eio/tests/test_client.ml
```diff
@@ -0,0 +1,124 @@
+module Net = Eio.Net
+module Stdenv = Eio.Stdenv
+module Switch = Eio.Switch
+open Cohttp_eio
+
+let conn env sw port resource_path =
```
talex5, line 6: Taking `resource_path` here means that every time you want a different resource, you also need to open a new connection. Why not separate out the connection from individual requests?

cohttp-eio/src/client.ml
talex5: > Two common questions I had in the past are how to timeout requests and how to connect to unix domain sockets. Since with ocaml5+eio it is done differently, I think we should document it either in the interface, in an example to link to the README or in the README, so that we avoid answering the same questions all the time.

Regarding timeouts, `Eio.Time.with_timeout` should do it. However, an example might be nice, indeed.

cohttp-eio/src/body.ml
talex5: Other http clients wait for the end of the connection if there's no length header. That's why the default server responses are all broken at the moment (see #883). Maybe the correct behaviour depends on the HTTP version?

talex5 (commented):
> @talex5 do you have further comments?

I found that some of the previous comments had been marked as resolved accidentally, so I unresolved them (it seems that GitHub lets you mark other people's comments as resolved for some reason).

I think it's mostly OK, but we should separate out connections from requests. Adding an example that sends two requests over one connection would probably make things clearer.

cohttp-eio/src/body.ml
bikallem: Agreed, IIUC #883 and #913 issues are the cohttp-eio servers not behaving correctly by setting the correct content-length header. As such I'd like to handle #883 and #913 in separate PRs after this PR. I would like to keep this PR focused mostly on cohttp Client functionality.

talex5: The point I'm trying to make is that HTTP responses don't have to have length headers, and we still need to handle that. https://www.rfc-editor.org/rfc/rfc7230 says:

> The length of a message body is determined by one of the following
   (in order of precedence):
> [...]
>  7.  Otherwise, this is a response message without a declared message
>       body length, so the message body length is determined by the
>       number of octets received prior to the server closing the
>       connection.

cohttp-eio/src/cohttp_eio.mli
bikallem: @talex5 Aha! Do you mean convert `read` data into `Body.chunk` ? Possibly could work. Let me see.

cohttp-eio/src/body.ml
bikallem: Thanks. This is done now in https://github.com/mirage/ocaml-cohttp/pull/879/commits/073ff8e1a35d7aa9876ee3d521c52c3bd7879dc2.

cohttp-eio/src/cohttp_eio.mli
bikallem: I have removed `'a conn` now. The are now passed directly to `call` as parameters. 7599e46254af78cc176d7a982deb3815d95a191a

cohttp-eio/src/cohttp_eio.mli
bikallem: Done in 7599e46254af78cc176d7a982deb3815d95a191a

cohttp-eio/tests/test_client.ml
bikallem: Done in 09c2d01be65454926d0a3260b199e88e92c7760c

bikallem:
@talex5 
>Regarding timeouts, Eio.Time.with_timeout should do it. However, an example might be nice, indeed.

I can't seem to be able to reply inline so creating a comment instead. This is now done in af9e2b5680040bed64c487f228e6ecd8487c1fcb

bikallem:
@talex5 I think I have addressed all of the latest review comments. Could you please mark them as resolved if so.

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -96,3 +100,81 @@ module Server : sig
...
+      allowed to have a request body. *)
+
+  type 'a body_allowed_call =
+    ?version:Http.Version.t ->
+    ?headers:Http.Header.t ->
+    ?body:Body.t ->
+    conn:(#Eio.Flow.two_way as 'a) ->
+    host ->
+    resource_path ->
+    response
```
talex5, line 133: This is minor, but it might be worth putting the `body` argument first. Then it could be defined more simply as:
```suggestion
  type 'a body_allowed_call = ?body:Body.t -> 'a body_disallowed_call
```
(possibly renaming the types for clarity)

cohttp-eio/src/cohttp_eio.mli
talex5: > Do you mean convert read data into Body.chunk ? Possibly could work. Let me see.

Yes. I don't see why the comment you linked is a problem for that:

1. "read_chunked returns an updated set of headers" : just return the unmodified headers in the fixed case
2. "Each chunk may have chunk extensions" : just return an empty set of extension for fixed bodies

talex5 (approved):
Thanks. There are things can be still be improved here, but I don't see any problem with merging it as it is and then working on the main branch.

cohttp-eio/src/client.ml
```diff
@@ -0,0 +1,115 @@
...
+(* Response Body *)
+
+let read_fixed ((response, reader) : Http.Response.t * Buf_read.t) =
+  match
+    Http.Header.get response.headers "Content-Length"
+    |> Option.get
+    |> int_of_string
+  with
+  | content_length -> Buf_read.take content_length reader
+  | exception _ -> Buf_read.take_all reader
```
talex5, line 112: Don't treat all errors (e.g. not-an-integer) as a missing content-length:
```suggestion
let read_fixed ((response, reader) : Http.Response.t * Buf_read.t) =
  match Http.Header.get response.headers "Content-Length" with
  | Some content_length -> Buf_read.take (int_of_string content_length) reader
  | None -> Buf_read.take_all reader
```

bikallem: Shouldn't an invalid `content-length` value(integer) be understood as an invalid `Content-Length` header and be treated as such? Specifically, 
>  5.  If a valid Content-Length header field is present without
       Transfer-Encoding, its decimal value defines the expected message
       body length in octets.  If the sender closes the connection or
       the recipient times out before the indicated number of octets are
       received, the recipient MUST consider the message to be
       incomplete and close the connection.

I assumed the "invalid" in a broader sense, not just missing 'Content-Length' header. So this would be invalid `Content-length: abc` since `abc` can't be a valid integer value in OCaml.

talex5: That's an interesting point. https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2 says:

> Any Content-Length field value greater than or equal to zero is valid.

I think it's assuming that the value will always be an integer, though it's not clear. It's also unclear why negative values should be ignored - possibly some systems use that to indicate streaming?

The older https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html says:

> When a Content-Length is given in a message where a message-body is allowed, its field value MUST exactly match the number of OCTETs in the message-body. HTTP/1.1 user agents MUST notify the user when an invalid length is received and detected.

bikallem:
@talex5 It seems I can't reply inline to some comments still. 
>Yes. I don't see why the comment you linked is a problem for that:
>
 >  "read_chunked returns an updated set of headers" : just return the unmodified headers in the fixed case
    "Each chunk may have chunk extensions" : just return an empty set of extension for fixed bodies

I did experiment with it a bit and came up with a possible read function as below:

```
val read : 
    ?chunk_handler:(Body.chunk -> unit) ->
    Http.Request.t ->
    Eio.Buf_read.t ->
    [`Header Http.Header.t option | `Content string]
```

Some considerations on the above API:
1. For most the uses cases of the library you are likely to use `read_fixed`, therefore it makes little sense to make the user of the library handle the 'chunked' case in every instance. Likely a user would do the following, 
```
match Cohttp_eio.Client.read req buf_read with
| `Content c -> ... (* do something with [s] *)
| `Header _ -> assert false
```
This makes the lib tedious and feels not optimized for the usage of the most often use case.  

2. In future PRs, I want to add more such specialized `read`s, such as `read_multipart`, `read_urlencoded` to name a few and possibly `read_json`, `read_xml` if there is appetite for it. This means the all knowing such `read` function is going to be less usable and more awkward to use.

The current `read` API design scales to these requirements and as such it would be preferable to maintain such design property.

cohttp-eio/src/cohttp_eio.mli
bikallem: Yes, I did consider such a scheme but found that when reading such types it read as `... so a body_allowed_call is also a body_disallowed_call`.  It didn't quite roll off the tongue for me so I left it as it was even though it is a bit repetitive. Also I didn't want to introduce another level of indirection since the types are mostly for alias and clarity for the values below, i.e get, post, head, etc.

cohttp-eio/src/client.ml
bikallem: Thanks for the pointers. They have been very helpful. I have created issues based on our conversation so far.
https://github.com/mirage/ocaml-cohttp/issues/927
https://github.com/mirage/ocaml-cohttp/issues/926

I want to address them in PRs after this.

mseri:
I am still not fully convinced about the read, but I think we can merge and keep iterating on the interface

mseri:
I would like to release an alpha version of the new cohttp in few weeks, this hopefully will bring more feedback from possible users on the table and help us further shaoing the api and the functionalities

bikallem:
Thanks.

PR #882, opened 2022-06-15
cohttp-eio: use Eio.Buf_read

talex5:
Eio now has its own buffered reader, so cohttp can just use that. This saves a lot of code, avoids embedding another parser combinator library in cohttp, and makes the public API much smaller.

I tested it with `wrk -t 4 -c 1000 -d 5s http://127.0.0.1:8080` while running the example server, and performance seems to be unaffected.

I also changed the `on_error` handler to not say the error happened on accept, since it's actually used for errors after accept, and to output a newline after the end. Perhaps it should instead be user configurable, or use a logging library instead of writing directly to stderr.

`read_fixed` now returns a string instead of bytes, which seems more logical. It also avoided a `Bytes.unsafe_to_string` in the example code.

The new parser doesn't support backtracking. I don't think this changes anything, but it would be worth double-checking we weren't relying on that anywhere.

I configured the reader with `~max_size:max_int`, which matches what the old code did, but we might want to make that configurable.

/cc @bikallem

### DIFF
--- a/cohttp-eio.opam
+++ b/cohttp-eio.opam
@@ -21,7 +21,8 @@ bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "dune" {>= "2.9"}
   "base-domains"
-  "eio"
+  "eio" {>= "0.2"}
+  "eio_main" {with_test}
   "cstruct"
   "bigstringaf"
   "fmt"
--- a/dune-project
+++ b/dune-project
@@ -342,7 +342,8 @@ should also be fine under Windows too.
   "A CoHTTP server and client implementation based on `eio` library. `cohttp-eio`features a multicore capable HTTP 1.1 server. The library promotes and is built with direct style of coding as opposed to a monadic.")
  (depends
   base-domains
-  eio
+  (eio (>= 0.2))
+  (eio_main :with_test)
   cstruct
   bigstringaf
   fmt


cohttp-eio/src/body.ml
```diff
@@ -43,56 +43,78 @@ let pp_chunk fmt = function
...
+  | c -> failwith (Printf.sprintf "Invalid quoted character %C" c)
 
 (*-- quoted-string = DQUOTE *( qdtext / quoted-pair ) DQUOTE --*)
-let quoted_string =
-  let dquote = char '"' in
-  let+ chars = dquote *> many_till (qdtext <|> quoted_pair) dquote <* dquote in
-  String.of_seq @@ List.to_seq chars
+let quoted_string r =
+  char '"' r;
+  let buf = Buffer.create 100 in
```
avsm, line 83: It would be lovely if this buffer could be reused somehow by the parsing fibre...

bikallem:
This was definitely in my to do list to evaluate `Eio.Buf_read` to replace the homegrown `Reader` module which is why I avoided using the back-tracking parser (`<|>`). If the performance is not affected, then this is definitely the way to go. Approved from my end.

cohttp-eio/src/body.ml
talex5: It's not very difficult, but we only call this once per chunk so it has no noticeable performance impact and it makes the code a little harder to read.

talex5:
> I haven't looked into implementation details of Eio.Buf_read yet but is Buf_read inherently not designed to support backtracking?

Yes, although there are ways to achieve it if necessary. But it's very uncommon for a network protocol to require backtracking, because it hurts performance. The example parser you linked (`quoted_string`) is also used in cohttp-eio, and was easy to convert to avoid backtracking.

Note that primitive parsers in Eio *are* atomic, so you could do things like `string "bar" <|> string "baz"`, but it won't work if the first branch needs to consume multiple tokens before realising it has gone wrong.

In theory, it would also be possible to export `Buf_read`'s state directly into another parser (e.g. Angstrom), so you could switch. But I suspect it's not necessary. I think Sprios said he regretted the backtracking support a bit, as people often forget the `commit` and end up with space leaks.

cohttp-eio/src/reader.ml
```diff
@@ -283,40 +20,40 @@ let token =
...
-      let headers = http_headers t in
-      let encoding = Http.Header.get_transfer_encoding headers in
-      commit t;
-      { Http.Request.meth; resource; version; headers; scheme = None; encoding }
+    let meth = p_meth t in
+    let resource = p_resource t in
+    let version = p_version t in
+    let headers = http_headers t in
+    let encoding = Http.Header.get_transfer_encoding headers in
+    { Http.Request.meth; resource; version; headers; scheme = None; encoding }
```
mseri, line 59: Using the current HEAD, you should be able replace this by a call to `Http.Request.make`

talex5: OK, I've rebased and updated to use that now.

cohttp-eio/src/body.ml
```diff
@@ -43,56 +43,78 @@ let pp_chunk fmt = function
...
+  let rec aux () =
+    match any_char r with
+    | '"' -> Buffer.contents buf
```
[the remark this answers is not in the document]
talex5: It works for me. How are you testing it?
```
utop # Eio.Buf_read.parse_string_exn quoted_string {|"hello \" world"|} |> print_endline;;
hello " world
```

cohttp-eio/src/body.ml
```diff
@@ -43,56 +43,78 @@ let pp_chunk fmt = function
...
   | '0' .. '9' -> true
   | 'a' .. 'f' -> true
   | 'A' .. 'F' -> true
   | _ -> false
 
-let quoted_pair =
-  char '\\'
-  *> satisfy (function ' ' | '\t' | '\x21' .. '\x7E' -> true | _ -> false)
+let quoted_char =
+  let+ c = any_char in
```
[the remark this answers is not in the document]
talex5: Before, it did `(qdtext <|> quoted_pair)`. The first parser reads the `\`, sees that it's not acceptable, and raises an exception, which is caught by `<|>`, which resets the state and then calls the second parser, which then reads the character again and checks that it is `\`.

Now, we just read the character once. If it's `\`, we use this parser to get the second character. Otherwise, we add the character we just read.

cohttp-eio/src/body.ml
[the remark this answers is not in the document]
bikallem: Ah, yes, I was still thinking you did the translation to Buf_read parser by parser. You have already read the `\` token.

cohttp-eio/src/body.ml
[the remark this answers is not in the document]
bikallem: Ah, yes, I was still thinking you did the translation to Buf_read parser by parser. You have already read the \ token in `quoted_string`.

cohttp-eio/src/body.ml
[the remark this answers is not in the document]
bikallem: Never mind, I realized `quote_char` in not the equivalent of the old parser.

cohttp-eio/src/body.ml
```diff
@@ -43,56 +43,78 @@ let pp_chunk fmt = function
   | Last_chunk extensions -> pp_chunk_extension fmt extensions
 
 open Reader
+open Eio.Buf_read
```
mseri, line 46: Does this require a lower bound on eio in the dune-project/opam file?

talex5: Good point. I've added a version constraint.
(in fact, the previous code also needed eio.0.2, because it uses the `Fiber` spelling, but the constraint was missing)

mseri:
> I haven't looked into implementation details of `Eio.Buf_read` yet but is `Buf_read` inherently not designed to support backtracking? i.e. can `Buf_read` implement support for backtracking `<|>` operator?

Seems that there is a crucial point still open right now. I think we should not merge this until this is settled if the new interface prevents the introduction of interesting features in the future

mseri:
Ping @bikallem

bikallem:
@mseri I am working with @talex5 to resolve the backtracking issue in `Buf_read`. 

@talex5 Can we do some benchmark performance vis a vis (https://github.com/ocaml-multicore/retro-httpaf-bench). I think it would be conclusive if we could compare the current parser with the `Buf_read` parser on both uring and non-uring(luv) setting. The current `Reader` caters for scenarios where the `read` syscall is expensive and it reduces such `expense` via over fetching in the buffer.

talex5:
> The current Reader caters for scenarios where the read syscall is expensive and it reduces such expense via over fetching in the buffer.

`Eio.Buf_read` does that too - that's the whole point of a buffered reader.

More benchmarks are welcome, but I already used [retro-httpaf-bench](https://github.com/ocaml-multicore/retro-httpaf-bench) for testing this. If you see something interesting when running it on your own machine, please post the results here.

`Reader` and `Eio.Buf_read` are both using the same approach (both are based on Angstrom). They both read into the space at the end of a bigarray and compress it from time to time.

However, getting the API right is more important that the implementation. In the unlikely case that the Eio parser is slower, we can just fix it in a point release. Shipping a custom parser API with cohttp-eio means we have to maintain that.

talex5:
> That's quite impressive. Thanks for the benchmark.

So, I forgot to call `commit` in the benchmark and created a memory leak in the Reader case! I've fixed that now. It still crashes trying to parse a quoted string though.

To avoid that, I added the option to use my non-backtracking quoted string parser (from this PR), which doesn't trigger the Reader bug. With that, the speeds are about the same. The remaining differences seem to be mostly due to how much sanity-checking is done. e.g. Reader uses `unsafe_get` whereas Buf_read uses a checked `get`.

With the default Buf_read, Reader is slightly faster. But if I remove the sanity check on `consume` (to match `Reader.incr_pos`) then Buf_read is slightly faster.

But this is all irrelevant to this PR. It doesn't make sense to ship a slow parser in eio and then override it with a faster one in cohttp-eio. If there are performance problems then we should fix them in eio, and if we think the speed/safety tradeoff should be different then we should change it in eio.

mseri:
I agree with @talex5, my point was more about the change preventing future development. If you say that, in case it is really needed, for example, backtracking can be added, I don't see a reason not to merge and address the optimizations at eio level

bikallem:
> It still crashes trying to parse a quoted string though.

It seems the parser is not finding the `"`. Hmmm ..

PR #886, opened 2022-06-29
Update for Eio 0.3

talex5:
- Use `accept_fork` instead of deprecated `accept_sub`.

- Remove unnecessary calls to `Flow.close`; it will get closed anyway when the connection handler returns.

Doing the close manually triggers a bug in eio 0.3: https://github.com/ocaml-multicore/eio/issues/244 - I didn't notice because I was testing a version of cohttp where I already removed the extra closes.

This PR is just the minimal changes to fix the build. I'll submit another PR on top of this that makes use of the new features soon.

### DIFF
--- a/cohttp-eio/src/server.ml
+++ b/cohttp-eio/src/server.ml
@@ -95,17 +95,13 @@ let rec handle_request reader writer flow handler =
       if not (is_custom body) then Writer.wakeup writer;
       if Http.Request.is_keep_alive request then
         handle_request reader writer flow handler
-      else Eio.Flow.close flow
-  | (exception End_of_file) | (exception Eio.Net.Connection_reset _) ->
-      Eio.Flow.close flow
+  | (exception End_of_file) | (exception Eio.Net.Connection_reset _) -> ()
   | exception Failure _e ->
       write_response writer bad_request_response;
-      Writer.wakeup writer;
-      Eio.Flow.close flow
+      Writer.wakeup writer
   | exception _ ->
       write_response writer internal_server_error_response;
-      Writer.wakeup writer;
-      Eio.Flow.close flow
+      Writer.wakeup writer
 
 let run_domain ssock handler =
   let on_error exn =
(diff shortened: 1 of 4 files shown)


talex5:
BTW, ocaml-ci hasn't been running on this repository for some time due to a malformed .ocamlformat file. If ocamlformat isn't being used, we could delete this file to get CI working again.

cohttp-eio/src/server.ml
```diff
@@ -95,17 +95,13 @@ let rec handle_request reader writer flow handler =
       if not (is_custom body) then Writer.wakeup writer;
       if Http.Request.is_keep_alive request then
         handle_request reader writer flow handler
-      else Eio.Flow.close flow
-  | (exception End_of_file) | (exception Eio.Net.Connection_reset _) ->
-      Eio.Flow.close flow
```
mseri, line 100: Is this always the case? Is there a risk that this changes in the future or that having an explict close would make it close earlier?

A priori, I would feel safer to always call the close, eventually with a catch to drop the relevant double close errors

talex5: Yes, it will always close. Eio ensures that FDs cannot be leaked. And it would be an error to close the connection and then continue trying to handle it. In the git version of Eio trying to close it manually here is a compile-time error (and in Eio 0.3, due to a bug, closing it manually results in an error being reported).

cohttp-eio/src/server.ml
```diff
@@ -114,7 +110,8 @@ let run_domain ssock handler =
   in
   Switch.run (fun sw ->
       while true do
-        Eio.Net.accept_sub ~sw ssock ~on_error (fun ~sw flow _addr ->
+        Eio.Net.accept_fork ~sw ssock ~on_error (fun flow _addr ->
+            Eio.Switch.run @@ fun sw ->
```
mseri, line 114: One last question. What is the reason one needs to nest the two Switch.run? Is it because there is the fork involved and then we want to manage a separate set of fibers?

talex5: It's because the writer fiber should be tied to the lifetime of this connection, not the lifetime of the server (if the handler function crashes, we want to cancel the writer fiber).

Note that this switch goes away in the next commit anyway (#887).

mseri:
The macos failure has nothing to do with this. I opened a separate issue for it

PR #893, opened 2022-07-01
chore: fix dune files in readme

rgrinberg:
formatting and remove commented fields

ps-id: 34572a59-38d9-4149-b1a7-6e27a0c41aec

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ Create this `dune` file
 ```
 cat - > dune <<EOF
 (executable
+ (public_name client_example)
  (name client_example)
  (libraries cohttp-lwt-unix))
 EOF
--- a/README.md
+++ b/README.md
@@ -141,9 +141,8 @@ Create this `dune` file
 ```
 cat - > dune <<EOF
 (executable
-  ; (public_name client_example)
-  (name client_example)
-  (libraries cohttp-lwt-unix))
+ (name client_example)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,9 +141,8 @@ Create this `dune` file
 ```
 cat - > dune <<EOF
 (executable
-  ; (public_name client_example)
-  (name client_example)
-  (libraries cohttp-lwt-unix))
+ (name client_example)
```
mseri, line 144:
```suggestion
 (public_name client_example)
 (name client_example)
```
This is what has been done also in the other two cases.

rgrinberg: Unfortunately, this example is incorrect b/c there's no `client_example` package anywhere. I suggest we drop the public_name so that the users isn't confused when the examples aren't working.

mseri: But that’s a mistake no? Should we add the example among the others or you think it is not worth it?

rgrinberg: What's a mistake? Public names aren't required.

mseri: Sorry, I responded too briefly.
I had added the examples from the README to the examples folder to make sure that they don't get stale (we should probably use mdx instead, and drop the examples folders, but that is a different topic). The error I meant was not to have added also the client among the examples.

We can remova all the public_names in the README then, no?

PR #894, opened 2022-07-01
chore: move examples to package dirs

rgrinberg:
ps-id: 0a9c708c-c207-4090-939f-05c25b82aaa9


mseri:
We need to fix the github action:
```diff
- opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
+ opam exec -- dune build bench/ cohttp-lwt-unix/examples/ cohttp-async/examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
```

mseri:
Macos failure is due to curl c library linking. It should be resolved, bus somehow sometimes it still appears in the ci

.github/workflows/workflow.yml
```diff
@@ -56,11 +56,9 @@ jobs:
 
       - run: opam install --with-test --deps-only http cohttp cohttp-lwt cohttp-lwt-unix cohttp-server-lwt-unix cohttp-mirage cohttp-async cohttp-curl-async cohttp-curl-lwt cohttp-curl cohttp-top
 
-      - run: |
-          opam exec -- dune build --profile=dev --only-packages http,cohttp,cohttp-lwt,cohttp-lwt-unix,cohttp-server-lwt-unix,cohttp-mirage,cohttp-async,cohttp-curl-async,cohttp-curl-lwt,cohttp-curl,cohttp-top @install
-          opam exec -- dune build bench/ examples/ cohttp-curl-async/bin/ cohttp-curl-lwt/bin/
```
mseri, line 61: In this way you don’t try to build bench or examples right? The reason I kept that line is to avoid the risk that they get stale as it happened in the past

PR #898, opened 2022-07-02
fix(cohttp-lwt-server-unix): increase performance

rgrinberg:
by switching to direct access, we work around Lwt_io's poor performance

For now I just did the write end. The performance is still poor but already much better. I can finish 100k requests with ab without timing out now :)

### DIFF
--- a/cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
+++ b/cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
@@ -16,6 +16,24 @@
 
 open Lwt.Syntax
 
+module Direct_access = struct
+  let rec write_sub (da : Lwt_io.direct_access) s ~pos ~len =
+    if len = 0 then Lwt.return_unit
+    else
+      let remaining = da.da_max - da.da_ptr in
+      if remaining = 0 then
+        let* (_ : int) = da.da_perform () in
+        write_sub da s ~pos ~len
+      else
+        let write_len = min remaining len in
+        Lwt_bytes.blit_from_string s pos da.da_buffer da.da_ptr write_len;
+        da.da_ptr <- da.da_ptr + write_len;
+        write_sub da s ~pos:(pos + write_len) ~len:(len - write_len)
+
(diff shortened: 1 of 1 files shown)


rgrinberg:
@anuragsoni can you have a look?

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
```diff
@@ -16,6 +16,24 @@
 
 open Lwt.Syntax
 
+module Direct_access = struct
+  let rec write_sub (da : Lwt_io.direct_access) s ~pos ~len =
+    if len = 0 then Lwt.return_unit
+    else
+      let remaining = da.da_max - da.da_ptr in
+      if remaining = 0 then
+        let* (_ : int) = da.da_perform () in
```
anuragsoni, line 25: `da_perform` can raise an exception I believe? We might want to handle that via a `Lwt.catch`

anuragsoni:
The changes look reasonable to me! This might be good to test, but I'm guessing there won't be significant changes in performance when switching the read path since you'll still have to allocate strings to drive the http parser. If we could get all io backends to work via direct access like APIs using their internal bigstrings, we can switch the cohttp parser to use bigstrings too and avoid that temporary allocation of strings.

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
rgrinberg: It can indeed raise Eof, BADFD, etc. There's not much we can do apart from aborting the connection though. As long as our finalizers are cleaning everything up it should be fine.

rgrinberg:
I had some good results even with this modest change, but I agree that we can do much better.

cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.ml
anuragsoni: That sounds good!

PR #899, opened 2022-07-02
test(cohttp-lwt): demonstrate infinite loop

rgrinberg:
@anuragsoni doesn't seem like the intended behavior

### DIFF
--- a/cohttp-lwt.opam
+++ b/cohttp-lwt.opam
@@ -27,6 +27,7 @@ bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "dune" {>= "2.9"}
   "ocaml" {>= "4.08"}
+  "ppx_expect" {with-test}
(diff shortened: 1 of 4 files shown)


anuragsoni:
This is indeed an embarrassing bug in my read_line implementation. I've proposed a patch against this branch (https://github.com/mirage/ocaml-cohttp/pull/901) that should fix this issue. The bug is also present on the async branch, but once https://github.com/mirage/ocaml-cohttp/pull/900 lands the read_line implementation in the shared bytebuffer module can be fixed by sharing the implementation from #901

rgrinberg:
Yes indeed. The PR sharing the buffer has landed.

anuragsoni:
> Yes indeed. The PR sharing the buffer has landed.

Thanks. I've updated #901 to move the patch to the common bytebuffer module in `cohttp`

PR #901, opened 2022-07-03
Improve Bytebuffer.read_line

anuragsoni:
Bytebuffer.read_line should be able to consume a line even when the line is longer
than the length of the buffer


anuragsoni:
This has been rebased against the new changes, and the read_line patch applied to the bytebuffer module that's part of the cohttp library

rgrinberg:
Cherry picked. Thank you.

PR #904, opened 2022-07-03
fix: stop installing binaries

rgrinberg:
All these various curl and server binaries are hardly useful and
shouldn't be installed.

ps-id: 9579817c-59f0-49d4-8381-c0d905de3a27

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,7 @@
   + fix: deprecate more aliases (rgrinberg #803)
   + refactor: deprecate connection value(rgrinberg #798)
   + refactor: deprecate using attributes (rgrinberg #796)
+  + fix: remove cohttp-{curl,server}-async (rgrinberg #..)
 
 ## v5.0.0 (2021-12-15)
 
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -46,6 +46,7 @@
   + fix: deprecate more aliases (rgrinberg #803)
   + refactor: deprecate connection value(rgrinberg #798)
   + refactor: deprecate using attributes (rgrinberg #796)
+  + fix: remove cohttp-{curl,server}-async (rgrinberg #..)
```
mseri, line 49:
```suggestion
  + cleanup: remove cohttp-{curl,server}-async (rgrinberg #..)
  + cleanup: remove chottp-{curl,server,proxy}-lwt (rgrinberg …)
```

mseri:
I know of many people using those binaries as a replacement for e.g. `python -m http.server` as local helpers. I think this is fine, but may generate some noise

rgrinberg:
if they're so useful, why don't we just provide a dedicated package for them and stick to one version (lwt) of the binary.

PR #905, opened 2022-07-03
refactor: move bytebuffer to http package

rgrinberg:
this way it can be used everywhere without taking on the cohttp dependency (bytebuffer itself doesn't introduce additional deps)

cc @anuragsoni


rgrinberg:
Merging as this is just shuffling stuff around and it's blocking more important PR's.

PR #906, opened 2022-07-04
Bytebuffer: demonstrate fixed reads issue

rgrinberg:
@anuragsoni similar issue with fixed sized reads.

### DIFF
--- a/http/test/bytebuffer/bytebuffer_tests.ml
+++ b/http/test/bytebuffer/bytebuffer_tests.ml
@@ -30,18 +30,18 @@ module Refill =
     end)
     (Src)
 
-let%expect_test "read" =
+let%expect_test "read line" =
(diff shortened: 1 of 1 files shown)


anuragsoni:
> similar issue with fixed sized reads

Not sure this is a problem. `read` is allowed to read shorter than `len` bytes. This should result in the body stream to consist of multiple sequences of strings. This behavior was chosen so the read buffer stays constant size, and we don't allocate temporary buffers/use an accumulator + string concat etc. This works out well considering that on the read path we always create bodies as streams.

rgrinberg:
Yes, the lwt/async backends are handling this well. However, I just realized that I'm not handling it right in the new server.

I think this function should be called either "read_at_most" or its return type changed to reflect partial reads. Otherwise it can be a bit confusing.

PR #909, opened 2022-07-06
Simplify Bytebuffer.read_line and add test for bug

rgrinberg:
@anuragsoni one small issue with read_line remaining


anuragsoni (approved):
looks good to me!

PR #910, opened 2022-07-07
test(eio): build examples with tests

rgrinberg:

ps-id: 4b2d5716-bec4-4eb0-97d2-f47d7bcb630a

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -341,6 +341,7 @@ should also be fine under Windows too.
   base-domains
   (eio (>= 0.3))
   (eio_main :with_test)
+  (uri :with-test)
   cstruct
   bigstringaf
   fmt
--- a/cohttp-eio.opam
+++ b/cohttp-eio.opam
@@ -23,6 +23,7 @@ depends: [
   "base-domains"
   "eio" {>= "0.3"}
   "eio_main" {with_test}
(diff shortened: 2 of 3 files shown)


dune-project
```diff
@@ -341,6 +341,7 @@ should also be fine under Windows too.
   base-domains
   (eio (>= 0.3))
   (eio_main :with_test)
```
mseri, line 343:
```suggestion
  (eio_main :with-test)
```
I had missed this.

mseri:
Th CI is still failing to build nit finding uri, that is surprising

mseri:
We should change https://github.com/mirage/ocaml-cohttp/blob/45996d4b591393dc05717e8831299cdb1323ceea/.github/workflows/workflow.yml#L97

either adding uri to the list or changing the invocation to `opam install --with-test cohttp-eio` (I think the only downside is that this may install all the test dependencies, including http’s, depending on opam’s version)

mseri:
Feel free to include the changes from #903 and then merge one the two fixes needed here are added

mseri:
I think the upcoming 2.2

mseri:
Great, then let’s go with the —with-test opion!

PR #916, opened 2022-07-16
fix(bytebuffer): stripping \r correctly

rgrinberg:
fix the slow path for reading lines to strip \r correctly

cc @anuragsoni

### DIFF
--- a/http/src/bytebuffer/bytebuffer.ml
+++ b/http/src/bytebuffer/bytebuffer.ml
@@ -90,31 +90,41 @@ struct
     drop t (len + 1);
     line
 
-  let rec read_line_slow t reader acc =
+  let get_line_buf t buf idx =
+    let len = idx - t.pos_read in
+    Buffer.add_subbytes buf t.buf t.pos_read len;
+    drop t (len + 1)
+
+  let rec read_line_slow t reader buf =
     if length t = 0 then
       refill t reader >>= function
-      | `Ok -> read_line_slow t reader acc
-      | `Eof -> (
-          IO.return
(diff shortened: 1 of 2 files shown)


anuragsoni:
Is `\r` a valid usecase to support for cohttp? I assumed that for what we use the bytebuffer for if we don't have a full `\r\n` then its an error?

rgrinberg:
> Is \r a valid usecase to support for cohttp? I assumed that for what we use the bytebuffer for if we don't have a full \r\n then its an error?

Do you mean just a `\n`? I don't think we handle `\r`

anuragsoni:
I was thinking about the usecase `foobar\r`. In this scenario the read_line attempts to look for a `\n` and since it doesn't find it and we reach an EOF it just responds with the entire unconsumed payload. I'm not sure if stripping the `\r` is worth it here. I'd understand if we strip `\n` or `\r\n`, but stripping just `\r` seems wrong?

rgrinberg:
Ah yes, that was an accident. It's now fixed.

Btw, should we make the `\r` mandatory? I think it's required by the spec in all cases.

anuragsoni:
> Btw, should we make the `\r` mandatory? I think it's required by the spec in all cases.

It is indeed mandatory according to the RFC https://datatracker.ietf.org/doc/html/rfc7230#section-3 I'd be in favor of enforcing CRLF, and treating an LF alone as an error.

mseri (approved):

mseri:
Should we document this in the CHANGELOG?

rgrinberg:
> Should we document this in the CHANGELOG?

Indeed we should. Although this PR doesn't change this behavior yet. I'll make the change in a subsequent PR.

PR #918, opened 2022-07-17
test: responses must terminate with \r\n

rgrinberg:

ps-id: 91660a79-042c-4831-be67-d467291229ac


rgrinberg:
@mseri @anuragsoni seems like the parser is overly permissive with these responses. Should we keep it this way or be more strict?

anuragsoni:
> Should we keep it this way or be more strict?

Being strict about CRLF seems like a good change to me. The RFC (https://datatracker.ietf.org/doc/html/rfc7230) specifically mentions CRLF, and since there are other breaking changes in the upcoming release, it might be a good time for cohttp to introduce a more strict behavior in its parser.

mseri:
I am also for being more strict given that we have the chance, but we should probably add this to the changelog

PR #921, opened 2022-08-02
fix: require \r\n in all parsers

rgrinberg:
ps-id: 319871fb-6a40-4df3-a48a-2cde4a7c543d


mseri (approved):

mseri:
This seems to be fixing the failing macos tests, the remaining 4 failures are linking errors with libcurl. Why it happens only on macos and with ocaml 4.08-4.11 I have no clue.. I wonder if we should just stop testing those 4 on macos

PR #938, opened 2022-09-08
Workaround for a "comment dropped" bug in ocamlformat

samoht:
See https://github.com/ocaml-ppx/ocamlformat/issues/2154


mseri:
Thanks

Issue #934, opened 2022-09-06
cohttp(client): Host header field should be the first header field


bikallem:
From https://www.rfc-editor.org/rfc/rfc9110#section-7.2

>A user agent that sends Host SHOULD send it as the first field in the header section of a request. For example, a GET >request to the origin server for <http://www.example.org/pub/WWW/> would begin with:
>
>```
>GET /pub/WWW/ HTTP/1.1
>Host: www.example.org
>```

mseri:
It says should, not must, though

bikallem:
> It says should, not must, though

Indeed. According to [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html#section-3) "SHOULD" = "recommended best practice". 
> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

I haven't found a counter explanation to not do it as per the spec.

PR #939, opened 2022-09-15
cohttp-eio(client): Ensure "Host" header is the first header in requests

bikallem:
Fixes https://github.com/mirage/ocaml-cohttp/issues/934

/cc @mseri

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem #939)
 - cohttp-eio: add TE header in client. Check TE header is server (bikallem #941)
 - cohttp-eio: add User-Agent header to request from Client (bikallem #940)
 - cohttp-eio: add Content-Length header to request/response (bikallem #929)
--- a/cohttp-eio/src/client.ml
+++ b/cohttp-eio/src/client.ml
@@ -32,6 +32,8 @@ let write_request request writer body =
       (Http.Request.headers request)
       body
   in
+  let headers = Http.Header.clean_dup headers in
+  let headers = Http.Header.Private.move_to_front headers "Host" in
   let meth = Http.Method.to_string @@ Http.Request.meth request in
   let version = Http.Version.to_string @@ Http.Request.version request in
   Buf_write.string writer meth;
@@ -72,15 +74,16 @@ let response buf_read =
 
 let call ?meth ?version ?(headers = Http.Header.init ()) ?(body = Body.Empty)
     ~conn host resource_path =
+  let host_hdr = "Host" in
   let headers =
-    if not (Http.Header.mem headers "Host") then
+    if not (Http.Header.mem headers host_hdr) then
       let host =
         match host with
         | host, Some port -> host ^ ":" ^ string_of_int port
         | host, None -> host
       in
-      Http.Header.add headers "Host" host
-    else headers
+      Http.Header.add headers host_hdr host
+    else Http.Header.move_to_front headers host_hdr
   in
   let headers =
     Http.Header.add_unless_exists headers "User-Agent" "cohttp-eio"
--- a/cohttp-eio/src/client.ml
+++ b/cohttp-eio/src/client.ml
@@ -74,16 +74,15 @@ let response buf_read =
 
 let call ?meth ?version ?(headers = Http.Header.init ()) ?(body = Body.Empty)
     ~conn host resource_path =
-  let host_hdr = "Host" in
   let headers =
-    if not (Http.Header.mem headers host_hdr) then
+    if not (Http.Header.mem headers "Host") then
       let host =
         match host with
         | host, Some port -> host ^ ":" ^ string_of_int port
         | host, None -> host
       in
-      Http.Header.add headers host_hdr host
-    else Http.Header.move_to_front headers host_hdr
+      Http.Header.add headers "Host" host
+    else headers
   in
   let headers =
     Http.Header.add_unless_exists headers "User-Agent" "cohttp-eio"
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -108,6 +108,8 @@ module Header = struct
     in
     loop h
 
+  let first t = match t with [] -> None | (k, v) :: _ -> Some (k, v)
+
   let get_multi (h : t) (k : string) =
     let rec loop h acc =
       match h with
@@ -163,6 +165,16 @@ module Header = struct
     | xs, _ ->
         let h = remove h k in
         add_multi h k xs
+
+  let move_to_front t hdr_name =
+    match t with
+    | (k, _) :: _ when caseless_equal k hdr_name -> t
+    | _ -> (
+        match get t hdr_name with
+        | Some v ->
+            let headers = remove t hdr_name in
+            add headers hdr_name v
+        | None -> t)
 
   let map (f : string -> string -> string) (h : t) : t =
     List.map
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -14,56 +14,51 @@ module Transfer = struct
 end
 
 module Header = struct
-  module Private = struct
-    external string_unsafe_get64 : string -> int -> int64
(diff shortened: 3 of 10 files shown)


CHANGES.md
```diff
@@ -1,4 +1,5 @@
 ## Unreleased
+# cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem #939)
```
mseri, line 2:
```suggestion
- cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem #939)
```

bikallem: Done

avsm:
Given that a Host header is essential in HTTP/1.1+, and that we have a dedicated Http.Header module, it makes more sense to specialise the internals of Http.Header to put "Host" first rather than expose a function `move_to_first`.  There is no other header value where `move_to_first` could be validly used.

cohttp-eio/src/client.ml
```diff
@@ -72,15 +72,16 @@ let response buf_read =
 
 let call ?meth ?version ?(headers = Http.Header.init ()) ?(body = Body.Empty)
     ~conn host resource_path =
+  let host_hdr = "Host" in
```
mseri, line 75: Why this extra constant? It is not going to change ever, what is the advantage of allocating the extra variable?

http/src/http.ml
```diff
@@ -164,6 +161,16 @@ module Header = struct
         let h = remove h k in
         add_multi h k xs
 
+  let move_to_front t hdr_name =
+    match t with
+    | (k, _) :: _ when String.equal k hdr_name -> t
```
mseri, line 166: This should be caseless_equal

http/src/http.ml
```diff
@@ -14,56 +14,51 @@ module Transfer = struct
 end
 
 module Header = struct
-  module Private = struct
```
mseri, line 17: I liked it better before, with the private module all separate, but I guess in this case that’s not an option

bikallem: Yes.

cohttp-eio/src/client.ml
bikallem: It is used in a few other places below. 

>what is the advantage of allocating the extra variable?

Interesting, is there an extra allocation when doing "let x = "Host" vs "Host"?

http/src/http.ml
bikallem: Done.

cohttp-eio/src/client.ml
mseri: Not sure to be honest. I assumed so unless the compiler unlined it away

bikallem: Done. I have removed the var declaration.

bikallem:
@mseri I believe I have addressed all of your comments.

avsm:
I'm still not a huge fan of this approach, which doesn't seem to take advantage of what we _do_ know about the HTTP request structure.  How about modifying Cohttp.Header to be a GADT (`server t` and `client t`) which enforces that the relevant must be defined when constructing one, and can safely cast from client->server without reallocating the structure?  This way it should be impossible to construct a Header.t without the right fields in place, and still be efficient.

rgrinberg:
> which enforces that the relevant must be defined when constructing one, and can safely cast from client->server without reallocating the structure?

Is that ever possible? A valid client header always requires a `Host` while a valid server header cannot contain one.

rgrinberg:
The GADT seems acceptable as well, but I'd like to hear other people's input if we're changing the fundamental `Http.Header.t` which is meant for interop. On the other hand, changing `Cohttp.Header.t` should be fine.

anuragsoni:
FWIW I'm not sure I'm a huge fan of adding an api that will most likely only ever be used for `Host` header in one very specific scenario. I like @rgrinberg 's proposal for adding a client aware serializer that writes out the Host header in the correct location.

I'm not too familiar with designing easy-to-use GADT based apis so I'm interested in learning about what you have in mind.

avsm:
There are a few more constraints beyond just the Host header, aren't there? E.g. you must have a content-length or be chunked encoding.  Or you can have an Upgrade field, but only with HTTP/1.1.  Full list here https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

It wouldn't be mad to encode Http.Header.t as a standard ADT

```
type t =
 | Host of string
 | Proxy_authorization of [`Basic | `Unknown of string ] * string
 | ...
```

The only reason I didn't do this in the original Cohttp was due to the serialisation/deserialisation cost when you really want to do this lazily (as most apps dont use all the headers).  But if there was a way to make such a representation efficient in modern OCaml, it would be nice to have well-typed headers rather than lots of string munging...

rgrinberg:
An ADT or whatever else we might use internally shouldn't matter much since we're likely to keep the types abstract. We could add additional types that maintain more invariants (but keeping the same internal representation):

```ocaml
module Header : sig
  type t

  module Client : sig
     type t
     val make : host:string -> ... -> t
     val host : t -> string
     ...
  end

  module Server : sig
    type t
    ...
  end
end
```

In this example I've included them in `Http`, but we're likely to incubate them in `Cohttp` first. The downside of course is that there's going to be no polymorphism, therefore the only way to write generic code is to convert from/to `Header.t` but it doesn't seem like a huge deal to me.

bikallem:
@avsm @mseri I am experimenting with the typed http header API in separate PR (https://github.com/mirage/ocaml-cohttp/pull/942). This PR addresses a specific issue and aims for an incremental improvement to the current API. As such I would like to scope this PR to the issue it is trying to address. 

@mseri If there are no other objections to the issue this PR is trying to solve, could you please approve.

mseri (approved):

mseri:
There are conflicts that need to be resolved

bikallem:
@mseri rebased on master and fixed the conflicts.

mseri:
There are new failures appearing:
```
#=== ERROR while compiling eio.0.5 ============================================#
# context     2.1.3 | linux/x86_64 | ocaml-variants.5.0.0+trunk | git+https://github.com/ocaml/opam-repository.git
# path        ~/work/ocaml-cohttp/ocaml-cohttp/_opam/.opam-switch/build/eio.0.5
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p eio -j 2 --promote-install-files=false @install
# exit-code   1
# env-file    ~/.opam/log/eio-5404-739508.env
# output-file ~/.opam/log/eio-5404-739508.out
### output ###
# (cd _build/default && /home/runner/work/ocaml-cohttp/ocaml-cohttp/_opam/bin/ocamlc.opt -w -40 -g -bin-annot -I lib_eio/core/.eio__core.objs/byte -I /home/runner/work/ocaml-cohttp/ocaml-cohttp/_opam/lib/cstruct -I /home/runner/work/ocaml-cohttp/ocaml-cohttp/_opam/lib/fmt -I /home/runner/work/ocaml-cohttp/ocaml-cohttp/_opam/lib/hmap -I /home/runner/work/ocaml-cohttp/ocaml-cohttp/_opam/lib/lwt-dllist -no-alias-deps -open Eio__core__ -o lib_eio/core/.eio__core.objs/byte/eio__core__Debug.cmo -c -impl lib_eio/core/debug.ml)
# File "lib_eio/core/debug.ml", line 32, characters 16-25:
# 32 |     | exception Unhandled -> default_traceln
#                      ^^^^^^^^^
# Error: This variant pattern is expected to have type exn
#        There is no constructor Unhandled within type exn
```
Is it expected?

talex5:
> There are new failures appearing [ on `ocaml-variants.5.0.0+trunk` ]

Yes, the OCaml API changed. Eio will be updated once OCaml 5 beta1 is out.

bikallem:
@mseri I have updated the cohttp-eio ci to use ocaml 5.0 beta1 and eio >= 0.6. Waiting for https://github.com/ocaml/opam-repository/pull/22247 to merge. After which the ci should be green, :crossed_fingers:.

bikallem:
@mseri cohttp-eio is now building ok again.

mseri:
Thanks

Issue #937, opened 2022-09-06
cohttp-eio(client): use User-Agent header


bikallem:
https://www.rfc-editor.org/rfc/rfc9110#section-10.1.5

Add "User-Agent" header in cohttp-eio Client. This is useful for some debugging scenarios.

PR #940, opened 2022-09-15
cohttp-eio(client): add User-Agent header

bikallem:
Adds "User-Agent" header to requests from Client.

Fixes https://github.com/mirage/ocaml-cohttp/issues/937

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- cohttp-eio: add User-Agent header to request from Client (bikallem #940)
 - cohttp-eio: add Content-Length header to request/response (bikallem #929)
 - cohttp-eio: add cohttp-eio client api - Cohttp_eio.Client (bikallem #879)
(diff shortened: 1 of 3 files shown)


mseri:
Is it really worth adding this? It is something trivial to add by hand. If we add all headers as optional parameters we overbloat the api imho. I am not so happy about cohttp logic for this but I find it better: https://github.com/mirage/ocaml-cohttp/blob/f5e35cdc3e3a09cea785271da634805b6a8bb3af/cohttp/src/request.ml#L64

bikallem:
Thanks. That's a valid point. It is only added if it is not added by user now.

mseri (approved):

bikallem:
@mseri Thanks for the approval. Could we merge this is there are no objections? It would make my other existing and incoming PRs a bit easier to work with. Thanks.

bikallem:
Thanks.

Issue #936, opened 2022-09-06
cohttp-eio: use TE headers when using chunked encoding


bikallem:
https://www.rfc-editor.org/rfc/rfc9110#section-10.1.4

Client : generate "TE: trailers" when using "chunked" encoding
Server: check request includes "TE" header before including chunked trailer headers

PR #941, opened 2022-09-16
Add TE header in client. Check TE header in server

bikallem:
Fixes https://github.com/mirage/ocaml-cohttp/issues/936.

### DIFF
--- a/http/src/http.mli
+++ b/http/src/http.mli
@@ -410,6 +410,14 @@ module Request : sig
 
       See https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2 *)
 
+  val supports_chunked_trailers : t -> bool
+  (** [supports_chunked_trailers t] is [true] if [t] contains HTTP header "TE:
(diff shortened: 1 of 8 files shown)


http/src/http.mli
```diff
@@ -410,6 +410,14 @@ module Request : sig
 
       See https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2 *)
 
+  val supports_chunked_trailers : t -> bool
+  (** [supports_chunked_trailers t] is [true] if [t] contains HTTP header "TE:
+      trailers". Otherwise it is [false]. *)
+
+  val add_te_trailers_header : t -> t
```
mseri, line 417: `_header` is redundant here
```suggestion
  val add_te_trailers : t -> t
```
You Are clearly modifying the header from the signature

bikallem: Done.

bikallem:
@mseri If there aren't any objecttions, could you please merge this PR.

PR #946, opened 2022-10-25
Fix dune and opam files

mseri:


http/test/bytebuffer/dune
```diff
@@ -1,5 +1,6 @@
 (library
  (name bytebuffer_tests)
+ (package http)
```
mseri, line 3: @rgrinberg how can we ensure that these tests are compiled only for `http`? If we add the package, it makes the CI fail (See e.g. https://github.com/mirage/ocaml-cohttp/actions/runs/3319027125/jobs/5483665232#step:4:21668) trying to compile the library when compiling http without tests, and if we remove it it makes `opam install --with-test` fail for all packages (See e.g. https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/4f4628e6e4b7bf442b85970ca709dc90e7175bfd/variant/compilers,4.08,cohttp.6.0.0~alpha0,tests) because it tries to run the tests for all packages and not just http

rgrinberg: The easiest way would be having a different test command for different packages: `$ dune runtest $pkg`.

mseri: Yes, but we already have that, no? If you run `dune runtest cohttp-async`, it already just runs `cohttp-async` tests. However, this `ppx_expect` test is running (or attempted) for all of them. And if we add it to the package, it fails because it is run also when the release (not test) build is executed

mseri: The only thing I could think of is to add `enabled_if` and link it to a configuration or the presence of an environment variable, but it seems a hack.

rgrinberg: > Yes, but we already have that, no? If you run dune runtest cohttp-async, it already just runs cohttp-async tests. 

That's not what we do in CI. I think we run `dune build @runtest`. It should be `$ dune build @$pkg/runtest/`

mseri: In the CI we do
```
dune runtest cohttp cohttp-async cohttp-curl cohttp-curl-async cohttp-curl-lwt cohttp-lwt cohttp-lwt-unix cohttp-mirage cohttp-server-lwt-unix cohttp-top http
```
and
```
dune runtest cohttp-eio
```
If `(package http)` is there, `dune build -p http` fails to build and we don't even reach the tests.

The call is opam-repository (automatically done by `opam install --with-test` for all packages) is `build dune build -p cohttp -j 31 --promote-install-files=false @install @runtest` and that, without the `(package http)`, fails with:
```
#=== ERROR while compiling cohttp.6.0.0~alpha0 ================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.13.1 | pinned(https://github.com/mirage/ocaml-cohttp/releases/download/v6.0.0_alpha0/cohttp-eio-6.0.0.alpha0.tbz)
# path                 ~/.opam/4.13/.opam-switch/build/cohttp.6.0.0~alpha0
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p cohttp -j 31 --promote-install-files=false @install @runtest
# exit-code            1
# env-file             ~/.opam/log/cohttp-7-765e9c.env
# output-file          ~/.opam/log/cohttp-7-765e9c.out
### output ###
# File "http/test/bytebuffer/dune", line 11, characters 2-24:
# 11 |   ppx_inline_test.config)
#        ^^^^^^^^^^^^^^^^^^^^^^
# Error: Library "ppx_inline_test.config" not found.
# -> required by library "bytebuffer_tests" in
#    _build/default/http/test/bytebuffer
# -> required by
#    _build/default/http/test/bytebuffer/.bytebuffer_tests.inline-tests/inline_test_runner_bytebuffer_tests.exe
# -> required by alias http/test/bytebuffer/runtest in
#    http/test/bytebuffer/dune:12
# File "http/test/bytebuffer/dune", line 14, characters 7-17:
# 14 |   (pps ppx_expect)))
#             ^^^^^^^^^^
# Error: Library "ppx_expect" not found.
# -> required by
#    _build/default/http/test/bytebuffer/.merlin-conf/lib-bytebuffer_tests
# -> required by
#    _build/default/http/test/bytebuffer/.bytebuffer_tests.inline-tests/inline_test_runner_bytebuffer_tests.exe
# -> required by alias http/test/bytebuffer/runtest in
#    http/test/bytebuffer/dune:12
```

rgrinberg: Note that adding `(package http)` makes the tests installable. That's certainly not what we want.

> I would have expected that dune build -p $pkg @runtest is even more restrictive than dune runtest $pkg

That's why `-p` is not recommended to end users. `dune build -p $pkg @runtest` will also run all tests that are unattached to a package.

rgrinberg: The invocation is incorrect for our repo, but it's the best dune can do in general. That's because doesn't actually know which parts of the repo belong to which package. But in the cohttp repo, we made it obvious using the directory structure.

mseri: I see. Thanks for the clarifications. Is thr solution to create template files for the dune invocation for all the opam files or we have better ways?

rgrinberg: That's the immediate solution. We should rework inline tests somehow to be able to attach them to specific packages without installing them.

PR #948, opened 2022-11-10
cohttp-eio: migrate eio tests to mdx based tests

bikallem:
This PR updates cohttp-eio tests to use mdx based tests primarily using `eio.mock` package. This allows tests to be repeatable especially with regards to networks and system time facilities.

PR #949, opened 2022-11-16
cohttp-eio(client): call Buf_write.flush() after serializing request

bikallem:
Required for https://github.com/mirage/ocaml-cohttp/pull/948

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## Unreleased
 - Upgrade dune to v3.0 (bikallem #947)
+- cohttp-eio: allow client to optionally configure request pipelining (bikallem #949)
 
 ## v6.0.0~alpha0 (2022-10-24)
 - cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem #939)
--- a/cohttp-eio/src/client.ml
+++ b/cohttp-eio/src/client.ml
@@ -6,6 +6,7 @@ type host = string * int option
 type resource_path = string
 
 type 'a body_disallowed_call =
+  ?pipeline_requests:bool ->
   ?version:Http.Version.t ->
   ?headers:Http.Header.t ->
   conn:(#Eio.Flow.two_way as 'a) ->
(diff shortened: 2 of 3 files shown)


talex5:
Why is this needed/useful? What happens without it?

I would guess this will slow pipelined requests down a bit. Is it for back-pressure (we probably need to add some support for that in Buf_write itself)? It might also improve performance to hint to the system that this is a good place to flush, but that might be better done with `Fiber.yield` (which doesn't require everything to be flushed before continuing).

bikallem:
> Why is this needed/useful? What happens without it?

Strictly not necessary or needed. `Client.get/post/head/delete` calls works correctly without flushing. This PR mostly improves individual request/response latency by calling `Buf_write.flush()` after it completes serializing the request. I noticed this while working on https://github.com/mirage/ocaml-cohttp/pull/948. In the client tests there the reading part (server response from mock socket) was done even before request was written to the socket. 

Is request pipelining a more normal usage and something that we want to optimize for? 

Perhaps an optional parameter `?flush` with default value of `false` to `Client.call` can cater to both use-cases well?

talex5:
> Is request pipelining a more normal usage and something that we want to optimize for?

The wrk benchmark doesn't use pipelining, which I guess is why we never optimise for it.

When pipelining, doing the read before flushing should be the fastest option, because we might be able to send two replies in a single packet that way. On the other hand, flushing first will be slightly faster when not pipelining and provides some back-pressure, so this PR seems fine to me (but it's good to document the reason - the current changelog makes it sound like a bug fix).

bikallem:
I have added optional parameter `pipeline_requests` to `Client.get/post/head/delete/put/patch/call`. The default value is [false] which means it is optimized by default for individual request/response, i.e. calls `Buf_write.flush()` after each call. I have also updated API doc and change log.

bikallem:
@mseri @talex5 Is there anything outstanding with this PR? If not could we merge this please.

mseri:
I think it can be merged

bikallem:
Thanks.

PR #950, opened 2022-11-21
ci: cancel in-progress workflow when a new one is started

bikallem:
This PR cancels in-progress workflows for a PR when a new push/commit starts a new workflow for the PR. This should hopefully make the CI a little bit more performant due to not doing unnecessary work.

More details : https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

/cc @mseri


mseri:
Thanks

PR #952, opened 2022-12-07
Update for Eio 0.7

talex5:
(once https://github.com/ocaml/opam-repository/pull/22619 is merged)

/cc @bikallem

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Unreleased
 - Upgrade dune to v3.0 (bikallem #947)
 - cohttp-eio: allow client to optionally configure request pipelining (bikallem #949)
+- cohttp-eio: update to Eio 0.7 (talex5 #952)
 
 ## v6.0.0~alpha0 (2022-10-24)
 - cohttp-eio: ensure "Host" header is the first header in http client requests (bikallem #939)
--- a/cohttp-eio.opam
+++ b/cohttp-eio.opam
@@ -21,7 +21,7 @@ bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "dune" {>= "3.0"}
   "base-domains"
-  "eio" {>= "0.6"}
+  "eio" {>= "0.7"}
   "eio_main" {with-test}
   "mdx" {with-test}
   "uri" {with-test}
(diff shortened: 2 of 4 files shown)


bikallem:
Oops. Sorry I didn't see this before opening mine (https://github.com/mirage/ocaml-cohttp/pull/954). That one has a one more commit (https://github.com/mirage/ocaml-cohttp/pull/954/commits/cfbbe49febeab514a21296083942e067598b3ba4) than this one.

mseri:
Was https://github.com/mirage/ocaml-cohttp/commit/cfbbe49febeab514a21296083942e067598b3ba4 needed as well?

bikallem:
Not strictly necessary but nice to have, I think. Perhaps cherry pick it?

talex5:
> Not strictly necessary but nice to have, I think. Perhaps cherry pick it?

What's the advantage of cfbbe49febeab? The new code is a little longer. Is it faster? I would have thought that allocating the Failure exception wouldn't be any quicker than allocating an empty string, and installing an exception handler on the normal path would be a little slower?

bikallem:
> What's the advantage of https://github.com/mirage/ocaml-cohttp/commit/cfbbe49febeab514a21296083942e067598b3ba4? The new code is a little longer. Is it faster? I would have thought that allocating the Failure exception wouldn't be any quicker than allocating an empty string, and installing an exception handler on the normal path would be a little slower?

I haven't done any bench-marking to verify if using exception handling is faster than allocation. However, my reasoning for the change is that exception handler in OCaml is very fast (since it is a register in a stack) and the overhead of a handler is that of a function call. Secondly, the code before was allocating an extra string in every call to `token`. This by itself is not much of an issue. However `token` is used quite a lot in the hot path (request parsing) - httips://github.com/mirage/ocaml-cohttp/blob/master/cohttp-eio/src/rwer.ml#L38, so therefore it is of some consequence if we can minimize allocation here. 

1. https://stackoverflow.com/questions/8564025/ocaml-internals-exceptions
2. https://stackoverflow.com/questions/12160390/ocaml-performance-of-exceptions

talex5:
I suggest merging this as-is, as it fixes the compiler warning.

The other change can happen in a separate PR if it turns out to be useful.

Issue #932, opened 2022-09-06
cohttp-eio(server) : an origin server must generate Date header


bikallem:
According to https://www.rfc-editor.org/rfc/rfc9110#section-6.6.1, we must generate a `Date` header in some reponses from an origin server.
>An origin server with a clock (as defined in [Section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110#http.date)) MUST generate a Date header field in all [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110#status.2xx), [3xx (Redirection)](https://www.rfc-editor.org/rfc/rfc9110#status.3xx), and [4xx (Client Error)](https://www.rfc-editor.org/rfc/rfc9110#status.4xx) responses, and MAY generate a Date header field in [1xx (Informational)](https://www.rfc-editor.org/rfc/rfc9110#status.1xx) and [5xx (Server Error)](https://www.rfc-editor.org/rfc/rfc9110#status.5xx) responses.

bikallem:
Depends on this https://github.com/ocaml-multicore/eio/pull/308.

PR #955, opened 2022-12-19
cohttp-eio: generate Date header in Server

bikallem:
Generates Date header in responses from `Cohttp_eio.Server`. 

Fixes https://github.com/mirage/ocaml-cohttp/issues/932

/cc @mseri @talex5

### DIFF
--- a/cohttp-eio/src/cohttp_eio.mli
+++ b/cohttp-eio/src/cohttp_eio.mli
@@ -35,6 +35,14 @@ module Server : sig
   type response = Http.Response.t * Body.t
   type handler = request -> response
 
+  type 'a env =
+    < domain_mgr : Eio.Domain_manager.t
+    ; net : Eio.Net.t
+    ; clock : Eio.Time.clock
+    ; .. >
+    as
+    'a
+
   (** {1 Request Body} *)
 
   val read_fixed : Http.Request.t -> Eio.Buf_read.t -> string option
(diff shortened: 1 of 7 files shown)


cohttp-eio/src/cohttp_eio.mli
```diff
@@ -84,15 +92,14 @@ module Server : sig
   (** {1 Run Server} *)
 
   val run :
-    ?socket_backlog:int ->
-    ?domains:int ->
-    port:int ->
-    < domain_mgr : Eio.Domain_manager.t ; net : Eio.Net.t ; .. > ->
-    handler ->
-    'a
+    ?socket_backlog:int -> ?domains:int -> port:int -> 'b env -> handler -> 'a
```
mseri, line 95:
```suggestion
    ?socket_backlog:int -> ?domains:int -> port:int -> 'a env -> handler -> 'b
```
Pointless comment, but why not just using the same as below

cohttp-eio/src/server.ml
```diff
@@ -65,14 +73,48 @@ let internal_server_error_response =
...
+    | 6 -> "Jun"
+    | 7 -> "Jul"
+    | 8 -> "Aug"
+    | 9 -> "Sep"
+    | 10 -> "Oct"
+    | 11 -> "Nov"
+    | 12 -> "Dec"
+    | _ -> failwith "Invalid HTTP datetime value"
+  in
+  Format.sprintf "%s, %02d %s %04d %02d:%02d:%02d GMT" weekday dd month year hh
```
mseri, line 106: The specification is https://www.rfc-editor.org/rfc/rfc5322#section-3.3

If I am not mistaken `to_date_time` will give the time as UTC/GMT, is that what the env#clock generates or we need to be careful with the timezone?

cohttp-eio/src/server.ml
```diff
@@ -65,14 +73,48 @@ let internal_server_error_response =
...
+
+let write_response ?request env writer (response, body) =
   let headers =
     let request_meth = Option.map Http.Request.meth request in
     Body.add_content_length
       (Http.Response.requires_content_length ?request_meth response)
       (Http.Response.headers response)
       body
   in
+  let headers = Http.Header.add headers "Date" (http_date env) in
```
mseri, line 117: Do we want to make it conditional only on the status being `2xx`, `3xx`, `4xx`?
We could save on some unnecessary clock calls

cohttp-eio/src/server.ml
bikallem: Yes. `Eio.Time.now` uses `Unix.gettimeofday()` so the time is in UTC timezone.

cohttp-eio/src/cohttp_eio.mli
bikallem: Done.

cohttp-eio/src/server.ml
bikallem: Good suggestion. Date header generation is now excluded on Informational and Server error responses.

bikallem:
@mseri I have addressed all of your feedback.

PR #958, opened 2022-12-22
Further improve Cohttp_eio.Client ergonomics

bikallem:
While working on documentation for cohttp-eio and https://github.com/mirage/ocaml-cohttp/pull/957, it dawned on me that the `Cohttp_eio.Client` ergonomics could be further improved. Such that users of `Cohttp_eio.Client` don't even need to use `Eio.Net.with_tcp_connect` for most use cases. This PR aims to do just that. Now we do,

```
open Cohttp_eio

let () =
  Eio_main.run @@ fun env ->
  let res = Client.get env ~host:"www.example.org" "/" in
  print_string @@ Client.read_fixed res
```
rather than a bit verbose below: 

```
open Cohttp_eio

let () =
  let host, port = ("www.example.org", 80) in
  Eio_main.run @@ fun env ->
  Eio.Net.with_tcp_connect ~host ~service:(string_of_int port) env#net
    (fun conn ->
      let host = (host, Some port) in
      let res = Client.get ~conn host "/" in
      print_string @@ Client.read_fixed res)
```
Note, however, we still retain the ability to pass in the `flow/connection` parameter like before. It is however now an optional parameter rather than a required one.

### DIFF
--- a/cohttp-eio/src/cohttp_eio.mli
+++ b/cohttp-eio/src/cohttp_eio.mli
@@ -112,20 +112,27 @@ end
 module Client : sig
   type response = Http.Response.t * Eio.Buf_read.t
 
-  type host = string * int option
-  (** Represents a server host - as ip address or domain name - and an optional
-      port value, e.g. www.example.org:8080, www.reddit.com *)
+  type host = string
+  (** Represents a server host - as ip address or domain name, e.g.
+      www.example.org:8080, www.reddit.com*)
+
+  type port = int
+  (** Represents a tcp/ip port value *)
 
   type resource_path = string
   (** Represents HTTP request resource path, e.g. "/shop/purchase",
       "/shop/items", "/shop/categories/" etc. *)
 
-  type 'a body_disallowed_call =
+  type 'a env = < net : Eio.Net.t ; .. > as 'a
+
+  type ('a, 'b) body_disallowed_call =
     ?pipeline_requests:bool ->
     ?version:Http.Version.t ->
     ?headers:Http.Header.t ->
-    conn:(#Eio.Flow.two_way as 'a) ->
-    host ->
+    ?conn:(#Eio.Flow.two_way as 'a) ->
+    ?port:port ->
+    'b env ->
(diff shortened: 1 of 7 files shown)


cohttp-eio/src/cohttp_eio.mli
```diff
@@ -112,20 +112,27 @@ end
...
+  type ('a, 'b) body_disallowed_call =
     ?pipeline_requests:bool ->
     ?version:Http.Version.t ->
     ?headers:Http.Header.t ->
-    conn:(#Eio.Flow.two_way as 'a) ->
-    host ->
+    ?conn:(#Eio.Flow.two_way as 'a) ->
+    ?port:port ->
+    'b env ->
+    host:host ->
```
mseri, line 135: why did we decide to not use `uri` here?
A `uri` will contain all the information at once

bikallem: Not really sure now but if I remember right being able to use the `flow/connection` was one I think. It wasn't obvious to me then but now I think it was the correct design. Additionally, I believe we can't use `unix` domain sockets in url without inventing our own syntax since http uri spec.  doesn't specify how we can use `unix` domain socket scheme in the uri.

mseri: That makes sense, the main reason I am asking is because right now it feels weird to separate the host from the path (mostly because any other library will take that all at once) but this is surely an improvement over the previous version

mseri: Should we try to be smart and add `/` as the first character if not already present?

bikallem: > Should we try to be smart and add / as the first character if not already present?

I think that is a good idea but not sure atm. However, I would like to address that in future PR please if we do go down that route.

bikallem: I have created a new issue so that we can discuss it further. https://github.com/mirage/ocaml-cohttp/issues/959

Issue #961, opened 2023-01-10
cohttp-eio: Document `Server.run`, possibly use single domain by default?


patricoferris:
Hello! 

`cohttp-eio` is yet unreleased officially but there is an `alpha` so hopefully that means you are happy to accept feedback :)) 

I'm using the server in a real-world application and I think it would be good to document `Server.run` and all of the parameters, in particular what the default parameters are for each of the optional arguments.

I also think it would be good to make the server use only one domain to accept connections by default, leaving it up to the user to add more if they please. I had assumed this and then happened to be using `Irmin` in the handlers (it was a REST server on top of an irmin store) and then was surprised by some weird bugs but the tl;dr is that Irmin is not domain-safe and I was reading/writing to it from multiple domains in parallel without realising it. I feel like potentially others might do the same.

bikallem:
Thanks for the feedback. Indeed, the API needs to be documented, possibly sprinkled with examples if the text is not clear. 

> I also think it would be good to make the server use only one domain to accept connections by default, leaving it up to the user to add more if they please.

Not too sure about this since in the new OCaml 5.0 world, multicore is the new default. But yes, the  default value need to be documented.

patricoferris:
Ah actually I misread the code and the error was a domain problem I created myself apologies, the cohttp server is single threaded by default my bad (unless `COHTTP_DOMAINS` is set, which should maybe also be documented). I've removed that from the title of the issue. 

Regarding the use of `env`:

https://github.com/mirage/ocaml-cohttp/blob/ec82f50f261b53d73839c22b6fd600273489503d/cohttp-eio/src/cohttp_eio.mli#L99

After a cursory glance, I don't think the connection handler needs everything in `env` just the clock?

talex5:
> I'm using the server in a real-world application and I think it would be good to document Server.run and all of the parameters

I think we should move `Server.run`'s functionality to Eio itself.

Though I'm not a big fan of creating the socket and running the accept loop in a single function. You almost always want to split those (e.g. to allow receiving a listening socket from systemd, or to send a notification that the socket is ready after you start listening but before entering the loop). Also, both parts may have lots of options (e.g. to set various socket options on the listening socket and to control how many concurrent users to allow in the loop), so splitting it in two might make the code easier to read.

> After a cursory glance, I don't think the connection handler needs everything in env just the clock?

Yes, looks like this should just be taking a clock.

bikallem:
[...]

PR #962, opened 2023-01-11
cohttp-eio : server api improvements

bikallem:
Addresses some points raised in #961, Specifically,

1. Make `Server.connection_handler` accept `env#clock` only.
2. Remove COHTTP_DOMAINS env variable usage.
3. Document `Server.run` and `Server.not_found_handler`.

/cc @patricoferris

### DIFF
--- a/cohttp-eio/src/cohttp_eio.mli
+++ b/cohttp-eio/src/cohttp_eio.mli
@@ -93,6 +93,16 @@ module Server : sig
 
   val run :
     ?socket_backlog:int -> ?domains:int -> port:int -> 'a env -> handler -> 'b
+  (** [run ~socket_backlog ~domains ~port env handler] runs a HTTP/1.1 server
+      executing [handler] and listening on [port]. [env] corresponds to
+      {!val:Eio.Stdenv.t}.
+
+      [socket_backlog] is the number of pending connections for tcp server
+      socket. The default is [128].
+
+      [domains] is the number of OCaml 5.0 domains the server will use. The
+      default is [1]. You may use {!val:Domain.recommended_domain_count} to
+      configure a multicore capable server. *)
 
   val connection_handler :
     handler ->
@@ -106,6 +116,7 @@ module Server : sig
   (** {1 Basic Handlers} *)
 
   val not_found_handler : handler
+  (** [not_found_handler] return HTTP 404 response. *)
 end
 
 (** [Client] is a HTTP/1.1 client. *)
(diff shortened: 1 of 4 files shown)


mseri:
Looks fine, thanks

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -93,10 +93,20 @@ module Server : sig
...
   val run :
     ?socket_backlog:int -> ?domains:int -> port:int -> 'a env -> handler -> 'b
+  (** [run ~socket_backlog ~domains ~port env handler] runs a HTTP/1.1 server
+      executing [handler] and listening on [port]. [env] corresponds to
+      {!val:Eio.Stdenv.t}.
+
+      [socket_backlog] is the number of pending connections for tcp server
+      socket. The default is [128].
+
+      [domains] is the number of OCaml 5.0 domains the server will use. The
```
patricoferris, line 103: I still think this is a little ambiguous (I did like the idea of having two functions to be very explicit about this)?

 - Does having `domains:1` mean I'm spawning one extra domain or is it single-threaded? I know the answer is the latter but I still think people might be confused?
 - If we go with this API I think it has to be stated that `handler` should be a domain-safe function

cohttp-eio/src/cohttp_eio.mli
```diff
@@ -93,10 +93,20 @@ module Server : sig
 
   val run :
     ?socket_backlog:int -> ?domains:int -> port:int -> 'a env -> handler -> 'b
+  (** [run ~socket_backlog ~domains ~port env handler] runs a HTTP/1.1 server
+      executing [handler] and listening on [port]. [env] corresponds to
+      {!val:Eio.Stdenv.t}.
```
patricoferris, line 98: Is this correct? `env` corresponds to the `env` above which only needs `domain_manager`, `net` and `clock` not all of `Eio.Stdenv.t`?

patricoferris (commented):
Thanks for the quick changes @bikallem :))

mseri:
Sorry, I have been too fast in merging. I suggest continuing the discussion here and making eventual fixes in a separate PR

cohttp-eio/src/cohttp_eio.mli
bikallem: Yes, not all of it. With that sentence, I was mostly trying to answer the question, "where do I get the instance of env?". Perhaps an example of `Server.run` will be more illustrative for that.

cohttp-eio/src/cohttp_eio.mli
bikallem: > Does having domains:1 mean I'm spawning one extra domain or is it single-threaded? I know the answer is the latter but I still think people might be confused?

This should be addressed (multiple specific functions and docs on `handler` being thread-safe) in future PR along with a few other suggestions from @talex5.

PR #967, opened 2023-02-01
cohttp-eio: remove unused code from tests to work with Eio 0.8

talex5:
The domain_mgr API changed in Eio 0.8, breaking the tests. But the code that broke isn't used for anything anyway, so remove it.

/cc @bikallem


bikallem:
Since we are not using domain manager in the tests, the changes looks correct.

PR #971, opened 2023-02-17
cohttp-eio : split Rwer module into Buf_read and Buf_write. 

bikallem:
Rwer module mostly served as miscellaneous dump bag of various readers and writers. The name Rwer always felt a bit incoherent even when it was initially added. However, I couldn't quite pin the correct module name for it. Recently, while working on another PR, it dawned on me that the readers and writers can be logically split into Buf_read and Buf_write modules by extending Eio Buf_read and Buf_write modules respectively.

Split off from https://github.com/mirage/ocaml-cohttp/pull/969 for easy review. 

/cc @patricoferris @mseri


bikallem:
The PR may look big (due to Files changed). However, it is mostly "lift and shift" operation.

bikallem:
Closing this in support of discussion happening at https://github.com/mirage/ocaml-cohttp/pull/969

PR #972, opened 2023-02-18
extend Body with reader/writer semantics, split Chunked_body

bikallem:
This PR is split off from https://github.com/mirage/ocaml-cohttp/pull/969 and builds on top of https://github.com/mirage/ocaml-cohttp/pull/971. Therefore only the last 2 commits belong to this PR. The first commit should go away with rebase to master after #971 is merged.

The PR improves the abstraction mechanism to define/use HTTP request/response body. A HTTP request/response body needs to also define Headers that go along with it. For e.g. a fixed body needs at least a Content-Type and Content-Length header, a chunked encoding needs `Transfer-Encoding` headers and possibly a Trailer header and trailers, A JSON body needs a different Content-Type value and so on. The current abstraction is leaky since it leaves an important requirement for the user of the library to devise/figure out and makes reuse of writers a bit less useful and convenient.

The current Body.t type also lacks a precise/formal notion of reading a body, for e.g. we have `Custom buf_write` for writing a body but we lack similar mechanism for reading. From the types (in the Body) it we get a sense that they are mostly for writing the body and not for reading the body.

The `reader/writer` abstraction aims to address the shortcomings identified above. We no longer need the brittle `Body.t`. However, for the purposes of keeping the trunk/master buildable, tests passing and the CI green, we still retain the "legacy" code in this PR. `Body.t` will be cleaned up in a future PR after Server/Client PRs (future PRs) are merged. 

`Body.writer` abstracts HTTP request/response body that can be written. `Body.reader` abstracts HTTP request/response body that can be read. The PR includes implementations of two common writers/readers, `content_writer/reader` and `form_values_writer/reader` in Body module. 

`content_writer/reader` encapsulates previous `Fixed` body semantics and `form_values_writer/reader` is a new body that reuses `content_writer/reader` to implement HTTP form body writing/reading. Notice the convenience with which we reuse writers/readers. Future PRs should add json and xml reader/writer by again reusing the content_writer/reader. 

Chunked encoding is split into separate module `Chunked_body` and now follows the reader/writer semantics. However, as mentioned above the `Body` module still includes chunked body functionality for now so as to keep the ci/trunk green. This duplication will be cleaned up once the relevant future PRs are merged.

- [ ] #971 needs to be merged.

/cc @patricoferris @mseri


bikallem:
Closing this in support of discussion happening at https://github.com/mirage/ocaml-cohttp/pull/969

PR #979, opened 2023-04-04
Cohttp-lwt: Send 100-continue on POST with Expect: 100-continue

edwintorok:
See RFC7231 section 5.1.1 on Expect 'Requirements for servers'.

I've tested this with Ezcurl-0.2.2 and a POST request which sends a 100-continue header.
Without this Curl will pause for 1s waiting for the server to send 100-continue. I also tested that inserting an artifical 10s sleep doesn't confuse curl (it gives up after 1s and sends the body anyway).

See [Curl documentation on 100-continue ](https://everything.curl.dev/http/post/expect100) and [HTTP/1.1 RFC7231 5.1.1 Expect](https://www.rfc-editor.org/rfc/rfc7231.html#page-34).

I've only implemented this for the Lwt backend for now.

### DIFF
--- a/cohttp-lwt/src/server.ml
+++ b/cohttp-lwt/src/server.ml
@@ -87,13 +87,30 @@ module Make (IO : S.IO) = struct
     in
     respond_string ~status:`Not_found ~body ()
 
-  let read_body ic req =
+  let read_body ?send_100_continue_if_needed ic req =
     match Http.Request.has_body req with
     | `Yes ->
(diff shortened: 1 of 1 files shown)


mseri:
Thanks a lot!

mseri:
Would it be possible to add it also to async?
And a test (if feasible). I think it should be easy to add in cohttp-lwt-unix `test_sanity.ml`.

Issue #978, opened 2023-04-04
Fd leak in handle_files


mseri:
There are scenarios if a client leaves before consuming the full response,
we might have a reference to a stream that doesn't get drained.

Not draining or closing the stream can lead to issues if its holding a resource like
a file handle as those resources will never be closed, leading to a leak.

Reported by Stéphane Glondu
Fix by samhot and anuragsoni in https://github.com/mirage/ocaml-cohttp/pull/977

PR #982, opened 2023-06-01
Cohttp_lwt_unix.Server: Do not leak fd serving potentially never ending bodies

pirbo:
Please tell me if I should be doing what I do in a complete different way (and c70ad81544eca744b9e8d7d3c7d08d90622c9f10 makes me think I do because Cohttp.Connection **are** useful in my case)

I want to use Cohttp to serve Server Sent Events (aka stream of things living as long as there is a client to receive them) and the way I implement them follows the sketch of 3226fdd33d165260400525cb16204e6d96da4ac8 but I realize that without this patch "Cohttp" never close the connections:
- I `dune exec cohttp-lwt-unix/examples/server_sent_event_lwt.exe` in one terminal and in an other I can see
```
$ lsof -p PID | wc -l
15
$ curl http://localhost:8800
Not found: //localhost:8800/
$ lsof -p PID | wc -l
15
$ curl http://localhost:8800/lol
Not found: //localhost:8800/lol
$ lsof -p PID | wc -l
15
```
but
```
$ curl http://localhost:8800/tic
Tic
Tic
^C
$ lsof -p PID | wc -l
16
$ curl http://localhost:8800/tic
Tic
^C
$ lsof -p PID | wc -l
17
```
And obviously, if you wait to drain the infinite body of the reply before calling `conn_closed` that would terminates it, it would take a while...

So, here I'm (maybe naively) swapping the 2.

### DIFF
--- a/cohttp-lwt/src/server.ml
+++ b/cohttp-lwt/src/server.ml
@@ -111,51 +111,60 @@ module Make (IO : S.IO) = struct
                 `Response rsp))
       (fun () -> Body.drain_body body)
 
-  let handle_response ~keep_alive oc res body handle_client =
+  let handle_response ~keep_alive oc res body conn_closed handle_client =
     IO.catch (fun () ->
         let flush = Response.flush res in
         Response.write ~flush
           (fun writer -> Body.write_body (Response.write_body writer) body)
           res oc)
     >>= function
-    | Ok () -> if keep_alive then handle_client oc else Lwt.return_unit
+    | Ok () ->
+        if keep_alive then handle_client oc
+        else
+          let () = conn_closed () in
+          Lwt.return_unit
     | Error e ->
         Log.info (fun m -> m "IO error while writing body: %a" IO.pp_error e);
+        conn_closed ();
         Body.drain_body body
 
-  let rec handle_client ic oc conn callback =
+  let rec handle_client ic oc conn spec =
     Request.read ic >>= function
-    | `Eof -> Lwt.return_unit
+    | `Eof ->
+        spec.conn_closed conn;
+        Lwt.return_unit
     | `Invalid data ->
         Log.err (fun m -> m "invalid input %s while handling client" data);
+        spec.conn_closed conn;
         Lwt.return_unit
     | `Ok req -> (
         let body = read_body ic req in
-        handle_request callback conn req body >>= function
+        handle_request spec.callback conn req body >>= function
         | `Response (res, body) ->
             let keep_alive =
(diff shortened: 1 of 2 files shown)


cohttp-lwt/src/server.ml
```diff
@@ -111,52 +111,60 @@ module Make (IO : S.IO) = struct
...
         let flush = Response.flush res in
         Response.write ~flush
           (fun writer -> Body.write_body (Response.write_body writer) body)
-          res oc
-        >>= fun () -> if keep_alive then handle_client oc else Lwt.return_unit)
+          res oc)
     >>= function
-    | Ok () -> Lwt.return_unit
+    | Ok () ->
+        if keep_alive then handle_client oc
```
mseri, line 122: If `handle_client` fails to read here we will never drain the body. This was related to an issue that led to a DOS: https://github.com/mirage/ocaml-cohttp/pull/977

mseri: I am not against this change, but we need to make it in such a way as to not re-open https://github.com/mirage/ocaml-cohttp/issues/978

The rule of thumb was that there shoudl be no infinite body, and if tje server has it, there should be a server-side way to prevent it. Maybe in this case it could be more suitable to use https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-server-lwt-unix/src/cohttp_server_lwt_unix.mli which has a finer control on the body?

mseri: (I know, it is rather undocumented unfortunately, it is still a work in progress)

pirbo: cohttp-server-lwt-unix looks promising and clearly it may be the way, going forward. 

That being said, I think this patch does not reopen #978: If `IO.catch [...]`returns `Ok ()`, it means that `Response.write [...]` succeed which means that the `Body.write_body [...]` have all been successfully completed so this body of that response fully consumed and there is no more things to drain here, isn't it ?

mseri: You are right, I went too fast there. I think it is fine. I wonder if we can add a test case, some kind of update of https://github.com/mirage/ocaml-cohttp/pull/977/files#diff-efd414a623c8f96d3b50c273441dd9a953d0d721a3369f4588ba01a3d1a0710d perhaps using the example you linked.

mseri:
Please add a changelog entry

mseri:
If I am not mistaking anything, this seems fine to me. I also spent some time testing it locally and it seems all right

PR #983, opened 2023-06-12
Cohttp_eio: Relax requirement on `Server.connection_handler`.

mefyl:
`Server.connection_handler` does not rely on its argument being a socket, a two way flow is sufficient. Relaxing the type constraint enables eg. unit testing with in-memory buffers.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - cohttp-lwt server: call conn_closed before drainig the body of response on error (pirbo)
+- cohttp-eio: Relax socket interface requirement on `Server.connection_handler`.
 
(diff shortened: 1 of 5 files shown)


mseri:
Ping @bikallem

bikallem:
LGTM.

mseri:
@mefyl bfore I merge, can you update the code and the lower bound on eio to fix the warning and the changelog?

mefyl:
I filled the Changelog and bumped eio. I think the remaining issues are because of API breakage in base v0.16, I'm not sure whether we want to upgrade base or reject 0.16.

mseri:
There is a separate PR to update to 0.16, I am going to merge it after this

mseri:
Thanks a lot

Issue #970, opened 2023-02-15
cohttp-eio: expose the underlying flow additionally to Buf_write when sending the response body


TheLortex:
Currently the body type is:
```ocaml
  type t =
    | Fixed of string
    | Chunked of chunk_writer
    | Custom of (Eio.Buf_write.t -> unit)
    | Empty
```

I believe it would be nice to have an additional body type that enables to write directly into the connection flow.
```ocaml
| Flow of (Eio.Flow.sink -> unit)
```

This way, one can use `Eio.Flow.copy` to send a file in a streaming fashion. I believe that using Buf_write would require writing the full content in the buffer, right ?

talex5:
@bikallem:
> i.e. IIUC flushing should ensure that you don't wait

The link says:

> [flush t] waits until all prior writes have been successfully completed.

@TheLortex:

> I believe that using Buf_write would require writing the full content in the buffer, right?

You can do is use `schedule_cstruct` to avoid a copy.

I'm not sure providing the raw sink will work, because we'll still typically need to do chunking. What are you planning to use it for?

TheLortex:
> What are you planning to use it for?

I wanted to fix this TODO in `eeww`: https://github.com/avsm/eeww/blob/main/src/main.ml#L78
>```ocaml    
>    |`Regular_file ->
>       (* TODO stream body instead of loading *)
>       let body = Eio.Path.load fname in
>       Cohttpx.respond_file fname body
>```
When serving files over HTTP, it's useful to be able to manipulate the flow so that a single `copy` is sufficient to stream the content of the file:
```ocaml
  let respond_file fname =
    ...
    let headers = Http.Header.of_list
      [ "content-type", mime_type;
        "content-length", string_of_int length;
      ] in
    let response =
      Http.Response.make ~version:`HTTP_1_1 ~status:`OK ~headers ()
    in
    response, Body.Flow (fun flow ->
      Eio.Path.with_open_in fname @@ fun file ->
      Eio.Flow.copy file flow)
```
without having to resort to the chunked encoding.

Issue #951, opened 2022-11-28
cohttp-eio(client): can cohttp-eio handle https connections now?


RCmerci:
I haven't found an example of HTTPS connections using cohttp-eio.

Does cohttp-eio support HTTPS conn now?

patricoferris:
Yes! But you have to do it yourself at the moment, see for example:

https://github.com/geocaml/carbon-intensity/blob/main/src/http_client.ml

(@bikallem or @talex5 might be able to comment more on the proper way to do this :))

RCmerci:
@patricoferris, thanks for the reply!

mseri:
@bikallem can we add an example in the docs or the mdx tests?

bikallem:
Yep, sure can do. I think mdx based README.md seems to have much more edifying effect so I will add it to cohttp-eio folder.

PR #984, opened 2023-06-13
Eio backend using stock parsing, serialization and signatures from Cohttp.

mefyl:
# Context

While trying out `cohttp-eio`, I found that all requests handler where blocking forever when trying to exhaust request or response bodies. After some digging, it appeared that - if I'm not mistaken - the current implementation simply forwards the raw socket as the body, not limiting the content length and potentially exposing chunked encoding, which is not the behavior of the other cohttp APIs.

While digging around, I found [this PR discussion](https://github.com/mirage/ocaml-cohttp/pull/969) suggesting that cohttp-eio should provide an interface as close as other cohttp APIs instead, and I gave it a shot.

# Implementation

This implementation relies mostly around providing a `Cohttp.S.IO` implementation suitable for eio, and relies mostly on the generic parser and serialization from the root `Http` and `Cohttp` libraries to do the heavy lifting - like cohttp-async does and cohttp-lwt partly does. This provides a working client and server with relatively little code and could IMO provide maximum consistency between the different cohttp APIs.

I have also experimented with generating a client compliant with a generic signature from a minimal request interface. I'm confident a generic client signature could be extracted into `Cohttp`, parameterized by its `Cohttp.S.IO`, and all APIs could generate a client compliant with that interface quite easily using this process - similarly to what `cohttp-lwt` does with its `Cohttp_lwt.S.Client` and `Cohttp_lwt.S.Server`.

Apart from this, the main difference is that request and response bodies are `Eio.Flow.{source,sink}` that entirely abstract the content-length and encoding. Iow, one can eagerly consume a whole body, and the flow will dry up at some point when everything has been consumed. The user never has to worry about content encoding, like in the async and lwt versions.

# Status

This PR is more of an RFC regarding this approach before going any further. It rewrites the content of `cohttp-eio/` but could perfectly live next to it, I just had no better idea baring `cohttp-eio-2/`.

Clients and servers are fully functional, but not heavily tested or very featureful - connection keepalive, multiple domains, ... I suppose it would be good to extract a common interface before making opinionated decisions over the interface. @talex5 [suggests it should mimic](https://github.com/mirage/ocaml-cohttp/pull/969#issuecomment-1435649809) the cohttp-lwt API, and I'd be happy to take a stab at it.

All in all, eager for any feedback and to possibly push this further if the approach is deemed beneficial.


mefyl:
I actually went ahead and :
* Extracted generic signatures for clients and servers from `cohttp-lwt` to `cohttp`
* Constrained `Cohttp_lwt.S.Server` and `Cohttp_lwt.S.Client` to these interfaces.
* Migrated my previous work in cohttp-eio to honor these interfaces.
* Added `Cohttp.Client.Make ()` to factor some clients boilerplate generation from a minimal set of functions (ie. generate `get`, `put` etc. from `call`) and used that in both lwt and eio.

IMO these interfaces could really use some facelift, but for now I strictly adhered to them to separate concerns. While `cohttp-lwt` was touched, it should just be code being moved around and it should behave exactly the same.

cohttp-eio/examples/server1.ml
```diff
@@ -28,19 +28,29 @@ let text =
...
   let port = ref 8080 in
   Arg.parse
     [ ("-p", Arg.Set_int port, " Listening port number(8080 by default)") ]
     ignore "An HTTP/1.1 server";
-
-  Eio_main.run @@ fun env -> Server.run ~port:!port env app
+  Eio_main.run @@ fun env ->
+  Eio.Switch.run @@ fun sw ->
+  let socket =
+    Eio.Net.listen env#net ~sw ~backlog:128 ~reuse_addr:true ~reuse_port:true
```
talex5, line 53: `reuse_port` is a bit specialised and can be confusing (you can have two servers on the same port at once). Suggest not using it in examples:
```suggestion
    Eio.Net.listen env#net ~sw ~backlog:128 ~reuse_addr:true
```

cohttp-eio/src/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name cohttp_eio)
  (public_name cohttp-eio)
- (libraries eio http fmt ptime))
+ (libraries cohttp eio eio.mock fmt http logs ptime uri uri.services))
```
talex5, line 4: The main library shouldn't depend on `eio.mock` (which is for testing). It should go in `tests/dune` instead.

cohttp-eio/src/io.ml
```diff
@@ -0,0 +1,52 @@
...
+
+  let with_input_buffer ic ~f =
+    let contents = Eio.Buf_read.peek ic in
+    let res, consumed =
+      f (Cstruct.to_string contents) ~pos:0 ~len:(Cstruct.length contents)
+    in
+    let () = Eio.Buf_read.consume ic consumed in
+    res
+
+  let read_line ic =
```
talex5, line 25: Could this just be `Eio.Buf_read.line`? The difference seems to be that this one discards the final line if it doesn't have a line terminator, whereas Eio's one returns it, but I'm not sure if that's intentional.

cohttp-eio/src/io.ml
```diff
@@ -0,0 +1,52 @@
...
+        Some line
+
+  let rec read ic len =
+    let contents = Eio.Buf_read.peek ic in
+    if Cstruct.length contents = 0 then
+      match refill ic with `Eof -> "" | `Ok -> read ic len
+    else
+      let consumed = Int.min len (Cstruct.length contents) in
+      let () = Eio.Buf_read.consume ic consumed in
+      Cstruct.to_string ~len:consumed contents
```
talex5, line 44: Could perhaps be simplified to:
```suggestion
    match Eio.Buf_read.ensure ic 1 with
    | exception End_of_file -> ""
    | () ->
      let len = Int.min len (Eio.Buf_read.buffered_bytes ic) in
      Eio.Buf_read.take len ic
```

cohttp-eio/src/server.ml
```diff
@@ -1,205 +1,119 @@
...
-    Eio.Std.Fiber.fork ~sw (fun () ->
-        Eio.Domain_manager.run domain_mgr (fun () ->
-            run_domain env ssock handler))
-  done;
-  run_domain env ssock handler
-
-(* Basic handlers *)
-
-let not_found_handler _ = not_found_response
+  accept ()
```
talex5, line 119: Could be just this, I think:
```ocaml
let run socket server =
  Eio.Net.run_server socket ~on_error:raise (fun socket peer_address ->
      Logs.info (fun m ->
          m "%a: accept connection" Eio.Net.Sockaddr.pp peer_address);
      let input = Eio.Buf_read.of_flow ~max_size:max_int socket in
      Eio.Buf_write.with_flow socket @@ fun output ->
      callback server socket input output
    )
```
But would probably be better to expose the connection handler function and let users call `Eio.Net.run_server` themselves, as it has extra arguments, and they can decide what to do about errors.

talex5 (commented):
I only skimmed this, but seems reasonable. Some minor suggestions in-line.

cohttp-eio/examples/client1.ml
```diff
@@ -1,6 +1,15 @@
 open Cohttp_eio
 
+let () = Logs.set_reporter (Logs_fmt.reporter ())
```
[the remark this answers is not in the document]
mefyl: The eio backend does not leverage multi-domain yet, I'm feeling like using this in the examples would falsely imply the opposite ?

cohttp-eio/src/io.ml
mefyl: ~~My understanding is that `Eio.Buf_read.line` also stops on `\n`, which we do not want here.~~ We actually do want to stop on `\n` because RFC7230 states parser MAY accept `\n` as terminator. The more you know 🌈

cohttp-eio/examples/client1.ml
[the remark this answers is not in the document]
mefyl: I've made the change with an additional comment.

adatario:
I very much like these changes!

This should make [`ocaml-websocket`](https://github.com/vbmithr/ocaml-websocket) work with `cohttp-eio`. Some WIP experiments: https://github.com/adatario/eio-ws/blob/main/src/main.ml#L37-L99. 

The only change required would be to expose the `IO` module (https://github.com/adatario/ocaml-cohttp/commit/41f5f10e86da0b41b28965a1cb562d3f43686be4). Maybe that could be added to this PR?

cohttp-eio/src/io.ml
```diff
@@ -0,0 +1,40 @@
+module IO = struct
+  type 'a t = 'a
+
+  let ( >>= ) v f = f v
+  let return v = v
+
+  type ic = Eio.Buf_read.t
+  type oc = Eio.Buf_write.t
+  type conn = Eio.Net.Sockaddr.stream
```
adatario, line 9: I suggest chaning this to:

`type conn = Eio.Switch.t * Eio.Net.Sockaddr.stream`

The switch can be used by the connection handler to allocate any resources needed (e.g. a fiber).

cohttp-eio/src/server.ml
```diff
@@ -1,205 +1,113 @@
...
-  run_domain env ssock handler
-
-(* Basic handlers *)
-
-let not_found_handler _ = not_found_response
+  handle ()
+
+let run socket server =
+  Eio.Net.run_server socket ~on_error:raise (fun socket peer_address ->
+      let () =
```
adatario, line 108: I suggest running a new switch in the `connection_handler` passed to `Eio.Net.run_server`:

```
(fun socket peer_address ->
  Eio.Switch.run (fun sw ->
  ...
))
```

The new switch is passed on to the user provided callback (see also suggested changes to `IO.conn`) and is used to group any fibers and resources allocated by the user. 

The `connection_handler` will only return when all resources are released. This also works nicely with the `max_connections:int` argument that can be passed to `Eio.Net.run_server` as the connection will be considered open as long as there are fibers running that were started by the user provided callback.

cohttp-eio/src/server.mli
```diff
@@ -0,0 +1,3 @@
+include Cohttp.Server.S with module IO = Io.IO and type body = Body.t
+
+val run : #Eio.Net.listening_socket -> t -> unit
```
adatario, line 3: I suggest adding all optional arguments that can be passed to `Eio.Net.run_server`:

```
  ?max_connections:int ->
  ?additional_domains:(#Domain_manager.t * int) ->
  ?stop:'a Promise.t ->
  on_error:(exn -> unit) ->
```
These can be passed to directly to `Eio.Net.run_server`.

adatario (commented):
Two suggestions that allows better usage of `Eio` functionality:

- Pass a switch to the user provided callback that can be used to start fibers and allocate resources.
- Accept all optional arguments that can be passed to `Eio.Net.run_server` in `Server.run`.

cohttp-eio/src/server.mli
talex5: Another option would be to expose the connection handler function, so that callers can call `run_server` themselves, e.g.

```ocaml
Eio.Net.run_server socket (Cohttp_eio.handle_connection http) ...
```

cohttp-eio/src/io.ml
```diff
@@ -0,0 +1,40 @@
+module IO = struct
+  type 'a t = 'a
+
+  let ( >>= ) v f = f v
+  let return v = v
+
+  type ic = Eio.Buf_read.t
+  type oc = Eio.Buf_write.t
```
adatario, line 8: Maybe it would also make sense to have:

```
type ic = Eio.Flow.sink
type oc = Eio.Flow.source
```

This would maybe allow directly streaming a file (see https://github.com/mirage/ocaml-cohttp/issues/970) and ping @TheLortex.

mefyl: It's been some time since I wrote this, but I remember I started with these and it didn't work out. AFAIR multiple functions from `IO` are called successively, and one creating then discarding eg. a `Buf_read.t` over a `Flow.source` each time would lose the buffered data. It seems to me the correct data type to persist the read state is the buffer, which includes the stateful buffered data and cursor position.

cohttp-eio/src/server.mli
mefyl: This is, I think, already the case: the "actual" API from `cohttp` we implement is `  val callback : t -> IO.conn -> IO.ic -> IO.oc -> unit IO.t`, which does not concern itself with connection handling. `Server.run` is just a convenience specific to cohttp-eio for the most common case.

- I'll still expose more options as this can hardly hurt.

talex5:
> Pass a switch to the user provided callback that can be used to start fibers and allocate resources.

It would certainly be useful to scope resources to a particular request. At the moment, any kind of streaming is painful. It's a bit painful with Lwt too (e.g. https://github.com/ocurrent/ocurrent/blob/3c297620ceaa4f7ce676dd1f3121fe010e0cabcc/lib_web/job.ml#L91-L111). The problem is that cohttp requires the user's function to finish before it starts writing the result. So you can't just perform a sequence of writes; you have to provide a stream to do them later.

Giving the user a switch so that they can start a fiber in it is one solution, but it might be simpler to have the callback do the writing itself. If cohttp made the response type abstract then we could have cohttp-eio redefine it so that the callback takes the response stream as an additional argument. `respond_string`, etc would still work as before, due to partial application.

e.g.

```diff
diff --git a/vendor/cohttp/cohttp/src/server.ml b/vendor/cohttp/cohttp/src/server.ml
index cbe08006..5ecc401a 100644
--- a/vendor/cohttp/cohttp/src/server.ml
+++ b/vendor/cohttp/cohttp/src/server.ml
@@ -3,10 +3,11 @@ module type S = sig
 
   type body
   type conn = IO.conn * Connection.t [@@warning "-3"]
+  type response
 
   type response_action =
     [ `Expert of Http.Response.t * (IO.ic -> IO.oc -> unit IO.t)
-    | `Response of Http.Response.t * body ]
+    | `Response of response ]
   (** A request handler can respond in two ways:
 
       - Using [`Response], with a {!Response.t} and a {!body}.
@@ -38,7 +39,7 @@ module type S = sig
 
   val make :
     ?conn_closed:(conn -> unit) ->
-    callback:(conn -> Http.Request.t -> body -> (Http.Response.t * body) IO.t) ->
+    callback:(conn -> Http.Request.t -> body -> response IO.t) ->
     unit ->
     t
 
@@ -48,7 +49,7 @@ module type S = sig
     status:Http.Status.t ->
     body:body ->
     unit ->
-    (Http.Response.t * body) IO.t
+    response IO.t
   (** [respond ?headers ?flush ~status ~body] will respond to an HTTP request
       with the given [status] code and response [body]. If [flush] is true, then
       every response chunk will be flushed to the network rather than being
@@ -64,7 +65,7 @@ module type S = sig
     status:Http.Status.t ->
     body:string ->
     unit ->
-    (Http.Response.t * body) IO.t
+    response IO.t
 
   val callback : t -> IO.conn -> IO.ic -> IO.oc -> unit IO.t
 end
```

cohttp-lwt can define `type response = Http.Response.t * Body.t` and avoid any other changes, while cohttp-eio could use `type response = Eio.Buf_write.t -> unit` internally.

mseri:
We also need to update this for `eio > 0.12`

talex5:
> We also need to update this for eio > 0.12

I have a branch with 0.12 support at https://github.com/mefyl/ocaml-cohttp/compare/eio...talex5:ocaml-cohttp:eio

(I used this for the Eio tutorial at ICFP)

mseri:
@adatario can you include those two commits on the PR?

mefyl:
I had limited time to work on this during the past months, sorry; let me know if additional help is needed beside @talex5's commits.

mseri:
My suggestion is to include the two commits from that eio branch here. Do another pass of review, merge this, and postpone changing the reponse to a separate PR (I will open an issue for it)

talex5:
It looks like this PR is also missing TLS support. e.g.

```ocaml
open Eio.Std

let net =
  let net = Eio_mock.Net.make "mocknet" in
  Eio_mock.Net.on_getaddrinfo net [`Return [`Tcp (Eio.Net.Ipaddr.V4.loopback, 443)]];
  let conn = Eio_mock.Flow.make "connection" in
  Eio_mock.Net.on_connect net [`Return conn];
  net

let () =
  Eio_main.run @@ fun env ->
  Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () ->
  let client = Cohttp_eio.Client.make net in
  Switch.run @@ fun sw ->
  let resp, _body = Cohttp_eio.Client.get client ~sw (Uri.of_string "https://www.example.org") in
  traceln "status: %S" (Cohttp.Code.string_of_status resp.status);
  ()
```
outputs the request in plain text:
```
+mocknet: getaddrinfo ~service:https www.example.org
+mocknet: connect to tcp:127.0.0.1:443
+connection: wrote "GET / HTTP/1.1\r\n"
+                  "host: www.example.org\r\n"
+                  "user-agent: ocaml-cohttp/\r\n"
+                  "\r\n"
+connection: closed
Fatal error: exception Failure("connection closed by peer")
```
Probably `Cohttp_eio.Client.make` should let you pass in a (`Uri.t -> Eio.Net.stream_socket_ty r`) function.

mseri:
I was coming to leave a similar comment, after somebody pointed at https://github.com/mirage/ocaml-cohttp/issues/951 in discord (https://discord.com/channels/436568060288172042/439062744105484288/1152047945731018782).

We should also update the client example to show how to make calls with TLS.

talex5:
I'll have a go at TLS support.

talex5:
I've pushed a branch with TLS here: https://github.com/mefyl/ocaml-cohttp/compare/eio...talex5:ocaml-cohttp:eio-tls?expand=1

mefyl:
WDYT about making the `https` argument of `Client.make` optional, to avoid all the `~https:None` ?

mseri:
I agree with you, I am also not convinced by the https name of the parameter. Perhaps haatp_authenticator, authenticator, tls_authenticator, something pointing that this is what will check the certificates?

Can we also have an example showing how to create a meaningful authenticator?

talex5:
> WDYT about making the https argument of Client.make optional, to avoid all the `~https:None` ?

I don't have a strong opinion, but I thought it was better to remind users that they need to decide whether they want HTTPS support or not, and where to put it if so.

> I agree with you, I am also not convinced by the https name of the parameter. Perhaps https_authenticator, authenticator, tls_authenticator, something pointing that this is what will check the certificates?

It's not that though, it's the thing that wraps a connection for https. Authentication is only part of what it does. We can't just take an authenticator here and do the TLS bit in the library because then cohttp-eio would have to depend on a TLS library.

I expect that eventually there will be a separate library to combine cohttp-eio with tls-eio (cohttp-eio-tls?). In the Lwt world, conduit handles this.

mseri:
I don’t have too strong of an opinion, I think the optional is cleaner but perhaps https is so pervasive nowadays that it is good to keep it explicit instead.


Thanks for the context, then I am ok in keeping the name provided that we explain a bit more its role. 

>> We can't just take an authenticator here and do the TLS bit in the library because then cohttp-eio would have to depend on a TLS library.

>I expect that eventually there will be a separate library to combine cohttp-eio with tls-eio (cohttp-eio-tls?). In the Lwt world, conduit handles this.

I understand that and I think it is a good place here to experiment with explicit tls management, it is something that we discussed to do in cohttp as well recently and I think we are moving in the same direction there as well. It will also solve the issue of hidden failures when 5e right ssl/tls libraries are not present.

mefyl:
Shall we make the parameter optional and merge this as-is then ? Or if we don't want to commit to an interface yet, we could merge it as-is and open a separate PR to add TLS support.

mseri:
Yes, let's go on with 2 PRs. Can you fix the conflict? I am going to merge this and then we can discuss TLS and the API separately

mseri:
I aim at merging this and https://github.com/mefyl/ocaml-cohttp/compare/eio...talex5:ocaml-cohttp:eio-tls as they are now though, release a new beta release and see if there are comments

talex5:
Note: tls-eio with Eio 0.12 support has now been released, so I uncommented the https example in my branch.

mseri:
Almost there, some test dependencies are missing from the opam file:
```
 File "cohttp-eio/tests/dune", line 3, characters 12-20:
3 |  (libraries alcotest cohttp-eio eio eio.mock eio_main logs.fmt)
                ^^^^^^^^
Error: Library "alcotest" not found.
-> required by _build/default/cohttp-eio/tests/test.exe
-> required by alias cohttp-eio/tests/all
-> required by alias cohttp-eio/default
File "cohttp-eio/tests/dune", line 6, characters 7-15:
6 |   (pps ppx_here)))
           ^^^^^^^^
Error: Library "ppx_here" not found.
-> required by _build/default/cohttp-eio/tests/test.pp.ml
-> required by alias cohttp-eio/tests/all
-> required by alias cohttp-eio/default
Error: Process completed with exit code 1.
```

dune-project
```diff
@@ -356,10 +363,12 @@ should also be fine under Windows too.
...
  (depends
   base-domains
+  (cohttp
+   (= :version))
   (eio
    (and
-    (>= 0.10)
-    (< 0.12)))
+    (>= 0.12)
+    (< 0.13)))
```
talex5, line 371: Shouldn't need an upper-bound here. We can add one later if it turns out to be needed, but we want to know if cohttp-eio isn't compatible with a new version of Eio.

mefyl: My rationale here is that the most important is that users build do not break out of the blue one morning (including their CI etc.) just because of a new eio release. One can use subminor versions for non-API breaking changes. I do however realize that there's no consensus on that.

mseri:
Sorry for the delay. This looks good to me. Thanks @mefyl and @talex5 
I think we can merge this and iterate over any necessary change from that

PR #985, opened 2023-06-14
Do not omit mandatory null content-length headers.

mefyl:
The code currently reuses the code that determines whether a body should be read to determine whether a `Content-Length` header should be sent. However, for unchunked bodies of size 0, there is no body to read, but `Content-Length: 0` is still mandatory for methods which may allow a body.

### DIFF
--- a/http/src/http.mli
+++ b/http/src/http.mli
@@ -21,6 +21,11 @@ module Method : sig
     | `Other of string ]
 
   val compare : t -> t -> int
+
+  val body_allowed : t -> bool
+  (** [body_allowed meth] returns whether [meth] allows a payload body to be
+      present per RFC7231. *)
+
   val of_string : string -> t
   val to_string : t -> string
   val pp : Format.formatter -> t -> unit
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,6 @@
 ## Unreleased
+
+- do not omit mandatory null Content-Length headers (mefyl #985)
 - cohttp-async, cohttp-curl-async: compatibility with core/async v0.16.0 (mseri, dkalinichenko-js #976)
 - cohttp-lwt server: call conn_closed before drainig the body of response on error (pirbo #982)
 - cohttp-eio: Relax socket interface requirement on `Server.connection_handler`. (mefyl #983)
--- a/cohttp/src/request.ml
+++ b/cohttp/src/request.ml
@@ -199,9 +199,9 @@ module Make (IO : S.IO) = struct
     in
     let headers = req.headers in
     let headers =
-      match Http.Request.has_body req with
-      | `Yes | `Unknown -> Header.add_transfer_encoding headers req.encoding
-      | `No -> headers
+      if Http.Method.body_allowed req.meth then
+        Header.add_transfer_encoding headers req.encoding
+      else headers
     in
     IO.write oc fst_line >>= fun _ -> Header_IO.write headers oc
 
--- a/cohttp/test/dune
+++ b/cohttp/test/dune
@@ -26,7 +26,7 @@
  (name test_request)
  (modules test_request)
  (forbidden_libraries base)
- (libraries cohttp alcotest fmt))
+ (libraries alcotest cohttp fmt http_bytebuffer))
 
 (rule
  (alias runtest)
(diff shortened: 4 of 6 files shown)


rgrinberg:
@anuragsoni can you have a look?

anuragsoni (approved):
This looks good!

One note for future releases is to re-think how we track this in a future release as the `encoding` field this api depends on has been deprecated and will be removed in a future release (https://github.com/mirage/ocaml-cohttp/blob/master/http/src/http.mli#L388-L389)

anuragsoni:
Hmm, that said, going through the async implementation again as a refresher, it looks like the current scenarios that are handled are:

* Servers fill in transfer encoding based on response bodies returned by users: https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-async/src/server.ml#L109-L113 but lwt's implementation is similar.
* Clients fill in transfer encoding for outgoing requests: https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-lwt/src/client.ml#L19-L33

The header module does perform a check if there's a transfer encoding header present before updating the header value so this change should be safe, but it'd be nice to either run this header setup in the individual backends based on the body types written for each backend (I think I lean towards linking the encoding type to the body as opposed to request as it is in the current cohttp version), or rethink whether the encoding type needs to be deprecated/removed.

mefyl:
I also think it would be better to automatically deduce this from the body type and entirely abstract this from the users, because in the vast majority of cases they do not care. HTTP 1 is a bit messy because it mixes transport-level and application-level headers, but IMO `content-length`, `transfer-encoding` etc. should not be manually configurable by the user. If they really want some manual control on these, `Request` and `Response` provide primitives to manually send headers and chunks of body.

It is also probably possible and desirable to factor all these behavior once and for all in the common `cohttp` library to ensure consitency across all backends. I'm currently toying with such factoring in [this MR](https://github.com/mirage/ocaml-cohttp/pull/984).

anuragsoni:
> I also think it would be better to automatically deduce this from the body type

+1

> It is also probably possible and desirable to factor all these behavior once and for all in the common `cohttp` library to ensure consitency across all backends. I'm currently toying with such factoring in [this MR](https://github.com/mirage/ocaml-cohttp/pull/984).

I haven't looked at the other pull-request, (and I'm not a maintainer for cohttp so take this with a grain of salt), but I found the shared signature/functor approach in cohttp a little awkward to work with when working on the new parser. This was exacerbated in the lwt backend where there was another layer added on top with the split between `lwt`, `lwt.unix`, and `mirage`. 

Personally my preference after working on cohttp and opium (which used to be built on top of cohttp), leans towards using the `http` library for writing the common types and operations on them (header manipulation, stateless parsers for requests/responses/chunked bodies etc), and letting each IO backend orchestrate the rest according to the best options available in each backend. As I see it, having the shared signatures in cohttp is a little helpful for maintainers of cohttp in that there can be a shared set of features that can be managed centrally, in-practice the shared signatures have been a little awkward to work with personally since any interaction with them necessitates instantiating a functor, which leads to either other libraries building on top of cohttp also being functorized, and/or picking a single backend and sticking to it.

mefyl:
I'm not talking about functors here, just :

```ocaml
module type CLIENT = sig
  type 'a io
  val call : request -> body -> (response * body) io
end

module Client_lwt : CLIENT with type 'a io = 'a Lw.t
module Client_eio : CLIENT with type 'a io = 'a
```

The end users don't need to instantiate a functor, they can use the values directly. Actually they needn't even know there is a common signature. The only aim would be indeed to help maintaining consistency, so that when switching backend the same feature set / API is guaranteed. One could also imagine writing a library that works with any cohttp backend, but that would indeed require a functor.

Now here I'm not even arguing for / against this, I just want to emulate the current API word-for-word to ease possible adoption of that MR, without bundling API design choice in it; that can come in a second time.

These discussion probably belong over there, for that matter :)

rgrinberg:
LGTM. can you rebase?

http/src/http.mli
```diff
@@ -21,6 +21,7 @@ module Method : sig
     | `Other of string ]
 
   val compare : t -> t -> int
+  val body_allowed : t -> bool
```
mseri, line 24:
```suggestion
  val body_allowed : t -> bool
  (** [body_allowed meth] is true if the http method allows a payload
      body to be present following RFC7231.  *)
  
```
I think we should document this method. This is just a suggestion, feel free to edit as appropriate

mseri:
Thanks!

PR #986, opened 2023-06-30
Cohttp: ensure "host" is the first header in requests and other fixes

mseri:
In addition fixes the test for null-header, which on releases would fail with
```
ASSERT null content-length header are sent
FAIL null content-length header are sent

   Expected: `"PUT / HTTP/1.1\r\nhost: someuri.com\r\nuser-agent: ocaml-cohttp/\r\ncontent-length: 0\r\n\r\n"'

   Received: `"PUT / HTTP/1.1\r\nhost: someuri.com\r\nuser-agent: ocaml-cohttp/v6.0.0_alpha2\r\ncontent-length: 0\r\n\r\n"'

Raised at Alcotest_engine__Test.check in file "src/alcotest-engine/test.ml", line 200, characters 4-261
Called from Alcotest_engine__Core.Make.protect_test.(fun) in file "src/alcotest-engine/core.ml", line 181, characters 17-23
Called from Alcotest_engine__Monad.Identity.catch in file "src/alcotest-engine/monad.ml", line 24, characters 31-35
```


mseri:
I would like to draft a new alpha release with these fixes, before we merge #984

avsm:
I'm ok with fixing the issue like this, but a simpler fix would be to just track the `host` header separately in Request outside of the Header module, and not introduce ordering constraints into Header itself.  The host header is special and no other cases like it exist, do they?

mseri:
The `header` fix is currently only affecting the `Request` module of `cohttp`. 

The change in `Header` is to introduce an API that has predictable iteration behaviour over an header list and document the behaviour, but nothing there knows nothing about `host`.

PR #988, opened 2023-07-06
Remove special treatment of "host" header

mseri:
As correctly pointed out by @anuragsoni, the RFC says that order of header fields isn't significant within a well-behaved server.https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2

> The order in which header fields with differing field names are
> received is not significant.  However, it is good practice to send
> header fields that contain control data first, such as Host on
> requests and Date on responses, so that implementations can decide
> when not to handle a message as early as possible.  A server MUST NOT
> apply a request to the target resource until the entire request

Since this introduces unnecessary complexity and unnecessary extra functions in the http package header module, I am removing the special treatment in this PR.

Users that want to enforce the ordering of the headers can do it with the same type of overhead directly in their code and can take 0d252ec as inspiration. Using `Header.to_list` and `of_list` in their code it is also safer than what we do here, in case the implementation details of the header module change.

### DIFF
--- a/cohttp-eio/src/client.ml
+++ b/cohttp-eio/src/client.ml
@@ -41,7 +41,6 @@ let write_request pipeline_requests request writer body =
       body
   in
   let headers = Http.Header.clean_dup headers in
-  let headers = Http.Header.Private.move_to_front headers "Host" in
   let meth = Http.Method.to_string @@ Http.Request.meth request in
   let version = Http.Version.to_string @@ Http.Request.version request in
   Buf_write.string writer meth;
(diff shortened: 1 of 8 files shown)


mseri:
Ping @anuragsoni @avsm

mseri:
In this case, we are favouring simplicity of the implementation over best practices (see also the relevant issue: https://github.com/mirage/ocaml-cohttp/issues/934)

mseri:
Moreover users that want to enforce Host as the first header can always add it as first header in the Headers parameter.

PR #992, opened 2023-07-19
cohttp-lwt: Don't leak asynchronous `Retry` exceptions.

mefyl:
Cohttp-lwt currently let `Retry` exceptions escape from `Lwt.async`, which wrongfully triggers `Lwt.async_exception_hook`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- cohttp-lwt: Do not leak `Retry` exceptions to `Lwt.async_exception_hook`. (mefyl #992)
+
 ## v6.0.0~alpha2 (2023-07-1)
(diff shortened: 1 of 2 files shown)


mseri:
Ah fantastic, thanks!

mseri (approved):

mefyl:
Actually wait a second, it looks like there are other occurences, patch incoming.

mefyl:
There is another `Lwt.async` is `shutdown`, but it looks like this function is never ever called.
The other occurrences are in the caching client, which I have not yet used, so I supposed merging this is already a win.

mseri:
I think the leaked `Retry` was just the one here. If I am wrong, the issue can be reopened

PR #1002, opened 2023-10-25
cohttp-eio: add Client.make_generic and HTTPS support

talex5:
https://github.com/mirage/ocaml-cohttp/pull/984#issuecomment-1737065489 said:

> I aim at merging this and https://github.com/mefyl/ocaml-cohttp/compare/eio...talex5:ocaml-cohttp:eio-tls as they are now though (perhaps with some more explanations on how to do tls properly), release a new beta release and see if there are comments

However, the TLS support wasn't included in the end, so I'm opening a separate PR for it here.

/cc @SGrondin

### DIFF
--- a/cohttp-eio/examples/client_tls.ml
+++ b/cohttp-eio/examples/client_tls.ml
@@ -5,18 +5,24 @@ let () =
   Logs_threaded.enable ();
   Logs.Src.set_level Cohttp_eio.src (Some Debug)
 
-let null_auth ?ip:_ ~host:_ _ = Ok None     (* Warning: use a real authenticator in your code! *)
+let null_auth ?ip:_ ~host:_ _ =
+  Ok None (* Warning: use a real authenticator in your code! *)
 
 let https ~authenticator =
   let tls_config = Tls.Config.client ~authenticator () in
   fun uri raw ->
-    let host = Uri.host uri |> Option.map (fun x -> Domain_name.(host_exn (of_string_exn x))) in
+    let host =
+      Uri.host uri
+      |> Option.map (fun x -> Domain_name.(host_exn (of_string_exn x)))
+    in
     Tls_eio.client_of_flow ?host tls_config raw
 
 let () =
   Eio_main.run @@ fun env ->
   Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () ->
-  let client = Client.make ~https:(Some (https ~authenticator:null_auth)) env#net in
+  let client =
+    Client.make ~https:(Some (https ~authenticator:null_auth)) env#net
+  in
   Eio.Switch.run @@ fun sw ->
   let resp, body =
     Client.get ~sw client (Uri.of_string "https://example.com")
--- a/cohttp-eio/examples/client_tls.ml
+++ b/cohttp-eio/examples/client_tls.ml
@@ -1,0 +1,26 @@
+open Cohttp_eio
+
+let () =
+  Logs.set_reporter (Logs_fmt.reporter ());
+  Logs_threaded.enable ();
+  Logs.Src.set_level Cohttp_eio.src (Some Debug)
+
+let null_auth ?ip:_ ~host:_ _ = Ok None     (* Warning: use a real authenticator in your code! *)
+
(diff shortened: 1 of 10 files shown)


mseri:
Thanks! Just in time, I was getting ready to release the new beta :)

cohttp-eio/src/client.mli
```diff
@@ -9,16 +9,20 @@ include
      and type body = Body.t
 
 val make :
-  https:(Uri.t -> [`Generic] Eio.Net.stream_socket_ty r -> _ Eio.Flow.two_way) option ->
-  _ Eio.Net.t -> t
-(** [make ~https net] is a convenience wrapper around {!make_generic} that
-    uses [net] to make connections.
+  https:
+    (Uri.t -> [ `Generic ] Eio.Net.stream_socket_ty r -> _ Eio.Flow.two_way)
+    option ->
```
mseri, line 14:
```suggestion
  ->
```
Since it is no longer optional, why not dropping the option from the signature?

cohttp-eio/examples/client_tls.ml
```diff
@@ -0,0 +1,32 @@
...
+
+let () =
+  Eio_main.run @@ fun env ->
+  Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () ->
+  let client =
+    Client.make ~https:(Some (https ~authenticator:null_auth)) env#net
+  in
+  Eio.Switch.run @@ fun sw ->
+  let resp, body =
+    Client.get ~sw client (Uri.of_string "https://example.com")
```
samoht, line 28: I am a bit sad with the amount of boilerplate we have to come up to do a simple `https` call. Could we somehow ship a default authenticator as well? The current API doesn't look very simple to use.

cohttp-eio/src/client.mli
```diff
@@ -1,9 +1,29 @@
+open Eio.Std
```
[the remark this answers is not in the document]
talex5: It would be a bit ugly - several other places would get a fair bit longer too.

cohttp-eio/src/client.mli
talex5: https support is optional (you don't have to use it); see https://github.com/mirage/ocaml-cohttp/pull/984#issuecomment-1726062338:

> I don’t have too strong of an opinion, I think the optional is cleaner but perhaps https is so pervasive nowadays that it is good to keep it explicit instead.

cohttp-eio/examples/client_tls.ml
talex5: We can't ship a default authenticator without adding extra dependencies to cohttp-eio. The plan is to have another package for that later (either cohttp-eio-tls, or maybe something conduit-like); this PR is just making that possible.

mseri: I think we could merge this and release it as beta, so that people can test it out in the meantime and inform possible api refinements. But it would be nice to have `cohttp-eio-tls` before we release the new stable 6.0.0

samoht: That sounds like a good plan. I'm already using that PR in a project of mine and it's working great 👍 Thanks for this!

cohttp-eio/src/client.mli
samoht: if the plan is to have a separate `cohttp-eio-tls` package I suggest keeping the parameter optional - otherwise it's a bit clumsy for HTTP calls to have to thread that `~https:None` parameter everwhere.

mseri: Yes, I agree. I was recalling what we had discussed incorrectly

mseri: Is this signature ok as is, or do we need to change `https` into `?https`?

samoht: I think I would prefer `?https` but I also understand the rationale to keep it explicit as the error mode (the call will just hang) is not great right now.

mseri: Let's leave it as is. We can make it optional in due course once there is the tls package if we can think of a better way to deal with it

PR #1008, opened 2023-11-21
Cookies: case-insensitive check of set-cookies

### LINKED ISSUE #1007: cohttp 5.3.0: `Cookie.Set_cookie_hdr.extract` only work with lower case `set-cookie` and not uppercase `Set-Cookie`
My work around is:
```ocaml
    let header =
      Header.add_multi header "set-cookie" @@ Header.get_multi header "Set-Cookie"
    in
```

I noticed that this has already been fixed in master with `Http.Header.Private.caseless_equal` but it'll be nice if the fix can be backported.

Thank you :)

chomosuke:
I ended up coping `caseless_equal` from header.ml to cookie.ml since I didn't want to expose `caseless_equal` in `header.mli` and I didn't know where to put shared private functions. Hope that's okay.

Closes #1007

### DIFF
--- a/cohttp/src/cookie.ml
+++ b/cohttp/src/cookie.ml
@@ -133,13 +133,29 @@ module Set_cookie_hdr = struct
       :: alist
     with Failure _ -> alist
 
+  let caseless_equal a b =
+    if a == b then true
+    else
+      let len = String.length a in
+      len = String.length b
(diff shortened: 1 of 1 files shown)


mseri:
Thanks and sorry for the delay

PR #1009, opened 2023-11-24
Add Eio server benchmark

Sudha247:
Adds a eio_server benchmark that I wrote for measuring the performance of cohttp-eio, similar to the `async_server` and `lwt_unix_server_new` benchmarks. Results obtained on a server machine (Intel Xeon Gold 5120) are below:

**eio_server**

```
141829 requests in 5.00s, 283.23MB read
Requests/sec:  28361.57
Transfer/sec:     56.64MB
```

**async_server**

```
105063 requests in 5.00s, 212.22MB read
Requests/sec:  21011.74
Transfer/sec:     42.44MB
```

**lwt_unix_server_new**

```
166961 requests in 5.00s, 333.42MB read
Requests/sec:  33393.19
Transfer/sec:     66.69MB
```

### DIFF
--- a/cohttp-bench.opam
+++ b/cohttp-bench.opam
@@ -26,8 +26,10 @@ depends: [
   "dune" {>= "3.0"}
   "core" {>= "v0.13.0"}
   "core_bench"
+  "eio" {>= "0.12"}
   "http" {= version}
   "cohttp" {= version}
+  "cohttp-eio" {= version}
   "cohttp-lwt-unix" {= version}
   "cohttp-server-lwt-unix" {= version}
   "cohttp-async" {= version}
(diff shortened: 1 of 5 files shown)


talex5:
Thanks! This will be useful for anyone wanting to optimise it.

Note: ocaml-ci is failing with `No solution found for any supported platform`. This is because async requires OCaml < 5.1 and Eio requires OCaml >= 5.1. It is possible to work around this by putting an explicit constraint of the version of OCaml in the opam files (e.g.
if ocaml-ci sees that cohttp-eio requires OCaml 5.1 then it won't try to include cohttp-eio in the set for older versions).

Sudha247:
Thank you, I'll take a stab at setting the constraints right.

mseri:
I think it should be enough to add `cohttp-eio` and `eio` to the dependencies of `cohttp-bench` in the `dune-project` file and regenerate https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-bench.opam

mseri:
The remaining failures are CI errors related to homebrew and apt-get. I am merging

PR #1011, opened 2023-12-29
cohttp-eio.Server: Don't blow up on client disconnections.

mefyl:
Currently any peer closing the connection unexpectedly will cause the server to shutdown.


mseri (approved):

mseri:
Can you add a changelog entry?

PR #1015, opened 2024-01-23
cohttp-eio: Don't blow up in `Server.callback` on client disconnections.

mefyl:
My [previous fix on this ](https://github.com/mirage/ocaml-cohttp/pull/1011) fixed the issue for `Server.run`, but not `Server.callback`.

### DIFF
--- a/cohttp-eio/src/server.ml
+++ b/cohttp-eio/src/server.ml
@@ -92,22 +92,37 @@ let write output (response : Cohttp.Response.t) body =
   in
   Eio.Buf_write.flush output
 
-let callback { conn_closed; handler } conn input output =
+let callback { conn_closed; handler } ((_, peer_address) as conn) input output =
   let id = (Cohttp.Connection.create () [@ocaml.warning "-3"]) in
   let rec handle () =
     match read input with
-    | `Eof -> conn_closed (conn, id)
+    | `Eof ->
+        let () =
+          Logs.info (fun m ->
+              m "%a: disconnected" Eio.Net.Sockaddr.pp peer_address)
+        in
+        conn_closed (conn, id)
+    | exception Eio.Io (Eio.Net.E (Connection_reset _), _) ->
+        let () =
+          Logs.info (fun m ->
+              m "%a: connection reset" Eio.Net.Sockaddr.pp peer_address)
+        in
+        ()
     | `Invalid e ->
         write output
           (Http.Response.make ~status:`Bad_request ())
           (Body.of_string e)
     | `Ok (request, body) ->
         let () =
-          match handler (conn, id) request body with
-          | `Response (response, body) -> write output response body
-          | `Expert (response, handler) ->
-              let () = Io.Response.write_header response output in
-              handler input output
+          try
+            match handler (conn, id) request body with
+            | `Response (response, body) -> write output response body
+            | `Expert (response, handler) ->
+                let () = Io.Response.write_header response output in
+                handler input output
+          with Eio.Io (Eio.Net.E (Connection_reset _), _) ->
+            Logs.info (fun m ->
+                m "%a: connection reset" Eio.Net.Sockaddr.pp peer_address)
         in
         if Cohttp.Request.is_keep_alive request then handle ()
   in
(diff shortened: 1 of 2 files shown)


cohttp-eio/src/server.ml
```diff
@@ -116,14 +134,10 @@ let callback { conn_closed; handler } conn input output =
...
-        Eio.Buf_write.with_flow socket @@ fun output ->
-        callback server (sw, peer_address) input output
-      with Eio.Io (Eio.Net.E (Connection_reset _), _) ->
+      Eio.Switch.run @@ fun sw ->
+      let () =
         Logs.info (fun m ->
-            m "%a: disconnected" Eio.Net.Sockaddr.pp peer_address))
+            m "%a: accept connection" Eio.Net.Sockaddr.pp peer_address)
+      and input = Eio.Buf_read.of_flow ~max_size:max_int socket in
+      Eio.Buf_write.with_flow socket @@ fun output ->
```
talex5, line 142: I would expect `with_flow` to raise if the connection is reset, so I think the error handling here is still needed.

cohttp-eio/src/server.ml
```diff
@@ -92,22 +92,40 @@ let write output (response : Cohttp.Response.t) body =
...
+          Logs.info (fun m ->
+              m "%a: disconnected" Eio.Net.Sockaddr.pp peer_address)
+        in
+        conn_closed (conn, id)
+    | exception Eio.Io (Eio.Net.E (Connection_reset _), _) ->
+        let () =
+          Logs.info (fun m ->
+              m "%a: connection reset" Eio.Net.Sockaddr.pp peer_address)
+        in
+        ()
```
talex5, line 110: When should `conn_closed` be called? I'd assume whenever the connection is closed for any reason, including this and other errors. Possible it should be attached to the switch in `run` instead - then it will always be called when a connection is closed, because the switch is what closes it.

talex5 (commented):
GitHub won't let me click Submit review unless I write something here.

cohttp-eio/src/server.ml
mefyl: Thanks

cohttp-eio/src/server.ml
mefyl: I'm unsure, since the argument is not documented in `cohttp/src/server.ml`. I'm not certain it should be called in case the expert response handler finished properly.

mefyl: If we want it called in all cases, I can't rely on the switch AFAICT since `callback` doesn't have it. This is the symptom of a subtler issue I think: when callback finishes - even successfully, it is done buffering the answer but not actually sending it, so it is too early to call `conn_closed`. So unless we `flush` in `callback`, it's tricky to find a relevant point at which to call it.

Looking at the Lwt implementation, it seems to me that we can exit the handling loop in the `Expert` case without calling `conn_closed`.

All in all right now it is my impression that `conn_closed` is not fit for any purpose other than maybe logging the disconnection, but I could be wrong. At the very least it needs a formal specification.

Aside from this, I don't think this PR changes the circumstances in which `conn_closed` is called or not, so I think this can be treated as a separate issue.

mefyl: I'm taking the liberty of resolving this since this PR does not change this behavior: `conn_closed` was only called on proper EOF before, and that didn't change.

I do agree that this is spurious, and `conn_closed` should be properly documented (if not dropped altogether).

mseri (approved):

cohttp-eio/src/server.ml
mseri: We can revise the connection closed since it is very limited right now. If my memory is not failing me, the expert mode must be dealt with by hand, while conn_closed was used to do "automatic" cleanup of resources in the least worse way we could think of for the "classical" connections, but we could not really find any good way to properly manage the connections in the lwt case and fully prevent leakages

talex5 (approved):
Looks fine to me.

It might be nice to add a test for connection reset errors. The old cohttp-eio had some tests where you could inject errors (though it only tried with `End_of_file`), but it looks like they got removed in the rewrite: https://github.com/mirage/ocaml-cohttp/blob/b7182a17e9769c7d08c5b05e2512e0eca605f74e/cohttp-eio/tests/server.md#tests

mseri:
Thanks

PR #1016, opened 2024-01-23
cohttp-eio: Add wired data debug logs like cohttp-lwt-unix.

mefyl:
This adds the same `>>> outbound wire data` and `<<< inbound wire data` as the lwt-unix implementation, enabling the same level of traceability when switching backend.

I don't suppose this requires a changelog entry, but lmk.


mseri:
No that's ok, thanks

PR #1017, opened 2024-01-27
Fix assertion in `Parser.Source.to_string_trim`.

mefyl:
I'm occasionally getting the assertion below. I didn't try reproducing, but an indexing error in `to_string_trim` is probably the culprit.

```
routine-push-notifications: internal error, uncaught exception:
                            Invalid_argument("String.sub / Bytes.sub")
                            Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
                            Called from Stdlib__String.sub in file "string.ml" (inlined), line 43, characters 2-23
                            Called from Http.Parser.Source.to_string_trim in file "http/src/http.ml" (inlined), line 956, characters 6-34
                            Called from Http.Parser.header in file "http/src/http.ml", line 1053, characters 16-60
                            Called from Http.Parser.headers.loop in file "http/src/http.ml", line 1065, characters 16-29
                            Called from Http.Parser.headers.(fun) in file "http/src/http.ml" (inlined), line 1069, characters 18-32
                            Called from Http.Parser.request in file "http/src/http.ml", line 1149, characters 18-32
                            Called from Http.Parser.run_parser in file "http/src/http.ml", line 1164, characters 10-18
                            Called from Http.Parser.parse_request in file "http/src/http.ml" (inlined), line 1171, characters 36-68
                            Called from Cohttp__Request.Make.read.(fun) in file "cohttp/src/request.ml", line 176, characters 16-63
                            Called from Cohttp_eio__Io.IO.with_input_buffer in file "cohttp-eio/src/io.ml", line 20, characters 6-74
                            Called from Cohttp__Request.Make.read in file "cohttp/src/request.ml", line 175, characters 6-270
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- http: Fix assertion in `Source.to_string_trim` when `pos <> 0` (mefyl #1017)
+
 ## v6.0.0~beta2 (2024-01-05)
 
 - cohttp-eio: Don't blow up Server on client disconnections. (mefyl #1011)
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -944,7 +944,7 @@ module Parser = struct
              "Http_parser.Source.substring: Index out of bounds., Requested \
               off: %d, len: %d"
              pos len);
-      let last = ref (t.pos + len - 1) in
+      let last = ref (t.pos + pos + len - 1) in
       let pos = ref (t.pos + pos) in
       while is_space (String.unsafe_get t.buffer !pos) do
         incr pos
--- a/http/src/http.ml
+++ b/http/src/http.ml
@@ -953,7 +953,7 @@ module Parser = struct
         decr last
       done;
       let len = !last - !pos + 1 in
-      String.sub t.buffer !pos len
+      if len < 0 then "" else String.sub t.buffer !pos len
 
     let rec index_rec t ch idx len =
       if idx = len then -1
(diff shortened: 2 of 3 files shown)


mseri:
Looks correct, can you add some tests for the function?

mefyl:
Laziness got the better of me, but you pushing me to add the test got me to the actual root of the problem so thanks for that:

* This bug was never actually triggered since `to_string_trim` is always called with `~pos:0`. For that reason, I can't test it with the public API, we would need to either expose the `Source` internals or add inline unit tests. We could also just fix it and call it a day for now since it's never actually used :innocent: 
* The actual bug is triggered with an empty header, which I fixed and added a test for.

rgrinberg:
@anuragsoni could you take a look?

anuragsoni:
This looks reasonable. The problem with the initial implementation was in the while loop that scans for whitespace leading to `pos` and `last` crossing over. Another potential change that addresses this scenario is listed below.

```ocaml
    let[@inline always] to_string_trim t ~pos ~len =
      if
        pos < 0
        || t.pos + pos >= t.upper_bound
        || len < 0
        || t.pos + pos + len > t.upper_bound
      then
        invalid_arg
          (Format.asprintf
             "Http_parser.Source.substring: Index out of bounds., Requested \
              off: %d, len: %d"
             pos len);
      let last = ref (t.pos + len - 1) in
      let pos = ref (t.pos + pos) in
      while is_space (String.unsafe_get t.buffer !pos) && !pos < !last do
        incr pos
      done;
      while is_space (String.unsafe_get t.buffer !last) && !last > !pos do
        decr last
      done;
      let len = !last - !pos + 1 in
      String.sub t.buffer !pos len
```

mseri:
Maybe your alternative suggestion is a bit more explicit. I don't have a strong preference between the two to be honest, do you have any preference?

anuragsoni:
> do you have any preference?

I don't have a strong preference on this!

mseri:
Let's merge then, we can always amend this later if we deem it necessary

mseri:
Thank you all a lot!

PR #1023, opened 2024-02-27
cohttp-eio: Improve error handling in example server

talex5:
The example was using `~on_error:raise`, which causes the server to crash if any error occurs handling a connection. This is unlikely to be what you want. Instead, log the error and continue.

This also now matches the behaviour of the Lwt example, which just logs errors and continues.

(from https://github.com/mirage/ocaml-cohttp/pull/1021#issuecomment-1934203471)


mseri:
Thanks

PR #1024, opened 2024-02-28
Make server response type abstract and allow streaming in cohttp-eio

### LINKED ISSUE #998: Make response type abstract
From https://github.com/mirage/ocaml-cohttp/pull/984#issuecomment-1684943089:

> Pass a switch to the user provided callback that can be used to start fibers and allocate resources.

It would certainly be useful to scope resources to a particular request. At the moment, any kind of streaming is painful. It's a bit painful with Lwt too (e.g. https://github.com/ocurrent/ocurrent/blob/3c297620ceaa4f7ce676dd1f3121fe010e0cabcc/lib_web/job.ml#L91-L111). The problem is that cohttp requires the user's function to finish before it starts writing the result. So you can't just perform a sequence of writes; you have to provide a stream to do them later.

Giving the user a switch so that they can start a fiber in it is one solution, but it might be simpler to have the callback do the writing itself. If cohttp made the response type abstract then we could have cohttp-eio redefine it so that the callback takes the response stream as an additional argument. `respond_string`, etc would still work as before, due to partial application.

e.g.

```diff
diff --git a/vendor/cohttp/cohttp/src/server.ml b/vendor/cohttp/cohttp/src/server.ml
index cbe08006..5ecc401a 100644
--- a/vendor/cohttp/cohttp/src/server.ml
+++ b/vendor/cohttp/cohttp/src/server.ml
@@ -3,10 +3,11 @@ module type S = sig
 
   type body
   type conn = IO.conn * Connection.t [@@warning "-3"]
+  type response
 
   type response_action =
     [ `Expert of Http.Response.t * (IO.ic -> IO.oc -> unit IO.t)
-    | `Response of Http.Response.t * body ]
+    | `Response of response ]
   (** A request handler can respond in two ways:
 
       - Using [`Response], with a {!Response.t} and a {!body}.
@@ -38,7 +39,7 @@ module type S = sig
 
   val make :
     ?conn_closed:(conn -> unit) ->
-    callback:(conn -> Http.Request.t -> body -> (Http.Response.t * body) IO.t) ->
+    callback:(conn -> Http.Request.t -> body -> response IO.t) ->
     unit ->
     t
 
@@ -48,7 +49,7 @@ module type S = sig
     status:Http.Status.t ->
     body:body ->
     unit ->
-    (Http.Response.t * body) IO.t
+    response IO.t
   (** [respond ?headers ?flush ~status ~body] will respond to an HTTP request
       with the given [status] code and response [body]. If [flush] is true, then
       every response chunk will be flushed to the network rather than being
@@ -64,7 +65,7 @@ module type S = sig
     status:Http.Status.t ->
     body:string ->
     unit ->
-    (Http.Response.t * body) IO.t
+    response IO.t
 
   val callback : t -> IO.conn -> IO.ic -> IO.oc -> unit IO.t
 end
```

`cohttp-lwt can define `type response = Http.Response.t * Body.t` and avoid any other changes, while `cohttp-eio` could use `type response = Eio.Buf_write.t -> unit` internally.

- - - 

Perhaps this could be a chance to revise how we do the writing also in the other backends.

talex5:
This fixes #998.

The first commit adds an abstract `response` type and then uses `type response = Http.Response.t * Body.t` in cohttp-lwt and cohttp-eio to keep the final types the same (cohttp-async already defined this type anyway and so didn't need any updates).

The second is just a few trivial changes to cohttp-eio to make the third commit diff smaller.

The third commit changes the `response` type in cohttp-eio to `type response = writer -> unit`. This allows
handlers to write the response inside the function, rather than returning a request to cohttp to write it later. That's useful because it allows e.g. streaming from an open file and then closing it afterwards.

Partial application means that code using `respond_string` etc will continue to work as before.

This also exposes a more polymorphic version of the `respond` function that accepts sub-types of `Flow.source`, so that callers don't need to cast the body.

/cc @mefyl


mefyl:
Looks good to me!

I would even suggest making `Cohttp_eio.Server.respond`'s `status` argument default to OK, just like `Cohttp.Response.make`, so that one need only replace `(Cohttp.Response.make (), body)` with `Cohttp_eio.Server.respond ~body` to adapt to this change.

talex5:
That probably is a better API, but it's trying to implement the signatures in the shared cohttp package, so they would have to be extra functions with different names if we did that. Maybe something for a separate PR?

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- cohttp-eio: Make server response type abstract and allow streaming in cohttp-eio (talex5 #1024)
```
mseri, line 3:
```suggestion
- cohttp-eio: allow streaming in cohttp-eio (talex5 #1024)
- cohttp: make server response type abstract (talex5 #1024)
```
This is also a change in cohttp signature itself, I think we should make it clear

cohttp-eio/src/server.ml
```diff
@@ -3,44 +3,42 @@ module IO = Io.IO
...
- *   Http.Request.t ->
- *   Eio.Flow.source ->
- *   Http.Response.t * Eio.Flow.source *)
+  [ `Expert of Http.Response.t * (IO.ic -> IO.oc -> unit)
+  | `Response of response ]
 
 type t = {
   conn_closed : conn -> unit;
-  handler : conn -> Http.Request.t -> body -> response_action IO.t;
+  handler : conn -> Http.Request.t -> body -> IO.ic -> IO.oc -> unit;
```
mseri, line 15: My only concern is that the change to old-style response was done to align the APIs, with this they APIs are diverging again (at least in the expectations and signatures)

CHANGES.md
talex5: Good point - I've updated it as suggested.

cohttp-eio/src/server.ml
talex5: Yes, although not when using the normal response functions. So it's certainly possible to write code that works with both.

mseri:
What is the status of this? I'd like to make a RC release of cohttp in the next couple of weeks. @mefyl and @talex5 do you think we should merge it as is or it needs further discussion?

mefyl:
AFAIR this looks fine to me and we can merge it.

mseri:
@talex5 can you have a look at the conflict? I think after that we could merge

talex5:
I rebased this. The conflict was due to #1025. If I understand the logic there correctly, this is what cohttp-lwt now does:

1. The user's handler uses e.g. `respond*` to create an HTTP 1.1 response, possibly missing the connection header, and returns it.
2. cohttp-lwt then notices that this isn't appropriate for the request and creates a fixed-up response with the correct header for the request (though still always HTTP 1.1, by the look of it).

The seems a bit odd. It would perhaps make more sense to create the response correctly in the first place, but that would require passing the request to `respond`, etc, which would be an API change.

For cohttp-eio, I modified the (abstract) `writer` type to include the request, so `respond` can build a correct response by itself. The test is simplified a bit because (as with Lwt's responders) the response is always HTTP 1.1, so `Http.Response.is_keep_alive` is always `true` when the header is missing.

mseri:
Thanks. I think it may be worth revising the api for cohttp-lwt but it could be done in a future 7.0 release

PR #1025, opened 2024-03-10
cohttp-{lwt,eio}: server: return connection header

ushitora-anqou:
ocaml-async returns a `connection` HTTP header if one doesn't exist in the response ([here](https://github.com/mirage/ocaml-cohttp/blob/cf2ae3344ed9211230a5680f251613326eacb296/cohttp-async/src/server.ml#L90-L95)), but cohttp-lwt and cohttp-eio do not.

This PR solves the above issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- cohttp-{lwt,eio}: server: return connection header (ushitora-anqou #1025)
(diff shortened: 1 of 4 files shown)


mseri:
Thanks. Can you add a changelog entry?

mseri (approved):

ushitora-anqou:
I updated my patch.

PR #1040, opened 2024-06-19
Fix deprecation warnings

MisterDA:
This PR fixes deprecation warnings, and the error that `ppx_expect.common` isn't found. As we're not using dune's explicit transitive deps feature (anymore?), they can be deleted from the dune file too.

### DIFF
--- a/cohttp-async/src/server.ml
+++ b/cohttp-async/src/server.ml
@@ -85,7 +85,6 @@ let handle_client handle_request sock rd wr =
                     Http.Request.is_keep_alive req
                     && Http.Response.is_keep_alive res
                   in
-                  let flush = Http.Response.flush res in
                   let res =
                     let headers =
                       Http.Header.add_unless_exists
@@ -95,7 +94,7 @@ let handle_client handle_request sock rd wr =
                     in
                     { res with Http.Response.headers }
                   in
-                  Io.Response.write ~flush
+                  Io.Response.write
                     (Body.Private.write_body Io.Response.write_body res_body)
                     res wr
                   >>= fun () ->
--- a/cohttp-curl-async/src/cohttp_curl_async.ml
+++ b/cohttp-curl-async/src/cohttp_curl_async.ml
@@ -40,7 +40,7 @@ module Context = struct
           | None -> ()
           | Some w ->
               Hashtbl.remove t.wakeners h;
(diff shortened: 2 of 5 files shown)


cohttp-async/src/server.ml
```diff
@@ -85,7 +85,6 @@ let handle_client handle_request sock rd wr =
                     Http.Request.is_keep_alive req
                     && Http.Response.is_keep_alive res
                   in
-                  let flush = Http.Response.flush res in
```
samoht, line 88: It's not clear to me we want to remove this completely - @rgrinberg what's the intended use for those nowadays?

rgrinberg: I think they're useless. Flushing should just be done for every request/response.

http/test/bytebuffer/dune
```diff
@@ -1,14 +1,6 @@
...
  (name bytebuffer_tests)
- (libraries
-  http_bytebuffer
-  ;; This is because of the (implicit_transitive_deps false)
-  ;; in dune-project
-  ppx_expect.config
-  ppx_expect.config_types
-  ppx_expect.common
-  base
-  ppx_inline_test.config)
```
rgrinberg, line 11: I separated this change from updating to ppx_expect 0.17.0.

cohttp-curl-async/src/cohttp_curl_async.ml
```diff
@@ -168,7 +168,7 @@ module Request = struct
           timeout
       in
       Cohttp_curl.Request.create ?timeout_ms ?headers method_ ~uri ~input
-        ~output ~on_response:(Ivar.fill response_ready)
+        ~output ~on_response:(Ivar.fill_exn response_ready)
```
rgrinberg, line 171: Does this need a minimum version bump for async?

cohttp-async/src/server.ml
MisterDA: `flush` defaults to true, so removing these lines is ok.

cohttp-curl-async/src/cohttp_curl_async.ml
MisterDA: Yes, `async_kernel >= v0.17.0`. I've added it to this PR.

PR #1052, opened 2024-06-27
fix: get rid of flush

rgrinberg:
We haven't yet released [Http] so it's ok to change things. It doesn't seem right to release the library with a deprecated feature.

It's easy enough to just drop this field in [Http]

### DIFF
--- a/cohttp/src/response.ml
+++ b/cohttp/src/response.ml
@@ -14,20 +14,17 @@
  *
   }}}*)
 
-open Sexplib0.Sexp_conv
-
 type t = Http.Response.t = {
   headers : Header.t;
   version : Code.version;
   status : Code.status_code;
-  flush : bool;
 }
 [@@deriving sexp]
 
-let compare { headers; flush; version; status } y =
+let compare { headers; version; status } y =
   match Header.compare headers y.headers with
   | 0 -> (
-      match Bool.compare flush y.flush with
+      match Stdlib.compare status y.status with
       | 0 -> (
           match Stdlib.compare status y.status with
           | 0 -> Code.compare_version version y.version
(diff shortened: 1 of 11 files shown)


cohttp/src/response.ml
```diff
@@ -43,16 +39,16 @@ let headers t = t.headers
 let encoding t = t.encoding
 let version t = t.version
 let status t = t.status
-let flush t = t.flush
 
 let make ?(version = `HTTP_1_1) ?(status = `OK) ?(flush = false)
     ?(encoding = Transfer.Chunked) ?(headers = Header.init ()) () =
+  ignore flush;
```
mseri, line 45:
```suggestion
let make ?(version = `HTTP_1_1) ?(status = `OK)
    ?(encoding = Transfer.Chunked) ?(headers = Header.init ()) () =
```

Why not dropping `flush` also from here?

rgrinberg: I think this would be a bigger breaking change. Since this `make` function has been around for a long time in many released versions. I'm honestly happy to get rid of it as well though if you prefer.

mseri: I think at this point why not. Removing the flush field is already breaking, we may as well do it once and for all

rgrinberg:
@mefyl do you want to have a look?

cohttp/src/response.ml
mefyl: If we're talking about the 6.0 major version upgrade, I think the point is precisely to allow API breakage? One has to do some cleanup at some point, and major versions are the only opportunity we have of doing so.

mefyl:
LGTM !

PR #1059, opened 2024-06-30
fix(async): do not read pipe to check if its empty

rgrinberg:
this isn't safe as it drops chunks from the user

cc @anuragsoni


anuragsoni (approved):
This looks good to me!

PR #1063, opened 2024-06-30
refactor: make [flush] non optional in private API

rgrinberg:
This change isn't visible to users. It just makes it easier to keep track internally where we explicitly pass [flush]


mseri (approved):

psafont:
> This change isn't visible to users.

I'm updating the dependencies of xapi-projects and this change breaks the usages of Request.write:

https://github.com/xapi-project/xen-api/blob/master/ocaml/xen-api-client/lib/xen_api.ml#L99


I'm just writing so it gets noticed for similar changes in the future

rgrinberg:
Sorry for the breakage. These functions were only exported to be shared between the async and lwt backends. They were never intended to be part of the public API and we should have communicated that better.

PR #1064, opened 2024-06-30
test(http): line folding

rgrinberg:
@anuragsoni did we omit this on purpose?


anuragsoni:
> did we omit this on purpose?

I vaguely remember reading that this wasn't allowed in the spec? I could be mistaken though.

rgrinberg:
It's allowed but marked as "obsolete". Seems to be used in the wild though

PR #1065, opened 2024-06-30
refactor: make [Cohttp.Body.drain] private

rgrinberg:


smuenzel:
There is a comment saying that connections are leaked if the response body isn't drained here:
https://github.com/mirage/ocaml-cohttp/blob/7b77cc7f18778967cbe3f64742290f268782fb7d/README.md?plain=1#L324-L325

I assume this is no longer the case, since `drain` is now private?

PR #1074, opened 2024-07-10
Fix `Client` overwriting its connection cache when using a custom `Net.ctx`

MisterDA:
When given a custom context, the client would always overwrite its connection cache. This prevents using a custom connection cache and a custom context. Access the connection cache directly.

### DIFF
--- a/cohttp-lwt/src/client.ml
+++ b/cohttp-lwt/src/client.ml
@@ -6,15 +6,15 @@ module Make (Connection : S.Connection) = struct
   module No_cache = Connection_cache.Make_no_cache (Connection)
   module Request = Make.Request (Net.IO)
 
-  let cache = ref No_cache.(call (create ()))
-  let set_cache c = cache := c
-
   type ctx = Net.ctx
 
+  let cache = ref None
+  let set_cache c = cache := Some c
+
   let cache ?ctx =
-    match ctx with
-    | None -> !cache
-    | Some ctx -> No_cache.(call (create ~ctx ()))
+    match !cache with
+    | None -> No_cache.(call (create ?ctx ()))
+    | Some cache -> cache
 
   include
     Cohttp.Generic.Client.Make


MisterDA:
I'm not convinced that this patch is complete, if the `ctx` changes, should the cache be recreated? It's probably not expected though, and it's tied to the problem of hiding a global ref behind the scenes…

rgrinberg:
Can't be of much help here as I have no idea how any of this is supposed to work. A test would be useful to understand what this is changing though

mseri:
Ping @madroach

madroach:
Hi @MisterDA,
I believe I understand your problem. You set a connection cache using `set_cache`, which won't be used if you use a custem `Net.ctx`. This behaviour is necessary, although maybe unintuitive, because a _`Net.ctx` is hiding in the connection cache_, which would be used instead of the provided custom `Net.ctx`.
If you want to use a custom `Net.ctx` with a connection cache, the way to achieve this is to create a connection cache providing your custom `Net.ctx` and then use this connection cache; possibly setting it as new default via `set_cache`.

This could maybe be improved upon by removing the `ctx` optional argument from the `cache` and various `call` functions altogether and thereby forcing users to go through a custom connection cache. This would make the semantics clearer, although remove some convenience.

mseri:
> Yes, I fell right into that. Should it be documented somehow? I find this behavior a bit surprising.

Sounds like a very good idea. Could you suggest an edit for the docstrings?

mseri:
> This would make sense, I think. I find it quite confusing to have a global cache hidden when calling the `Client.xxx` functions. But then, if both the connection cache and the `Net.ctx` have references to each other, and it would sound a bit like a design flaw…

I agree, sounds like a good idea to just have it in the `Net.ctx`. Would that impose a breaking change only for the lwt api or for all the apis?

madroach:
I did something similar. Since `ctx` is abstracted in `Cohttp.Generic.Client`, it can easily be changed just in `Cohttp_lwt.Client`.
This will obviously break any code using `ctx` in `Cohttp_lwt.Client`. I did this change in #1075 as a proof of concept.

madroach:
> Yes, I fell right into that. Should it be documented somehow? I find this behavior a bit surprising.

It is documented. See the `Cohttp_lwt.S.Client.ctx`.

madroach:
A downside of the approach to change `ctx` in `Cohttp_lwt.Client` is that it is confusing that `ctx` has a different type just in `Cohttp_lwt.Client`.

PR #1080, opened 2024-07-23
Support proxying client requests through direct and tunneling HTTP/HTTPS proxies

MisterDA:
This PR implements a new `Cohttp_lwt_unix.Connection_proxy` module satisfying `Cohttp_lwt.S.Connection_cache`, allowing client HTTP/HTTPS requests to be routed through proxies.

I recommend a quick glance at [Everything curl - 9.8 Proxies](https://everything.curl.dev/usingcurl/proxies/index.html) for a nice introduction to proxies. As curl is well established, I try following the same terminology and support similar features.

- HTTP proxy

  > An HTTP proxy is a proxy that the client speaks HTTP with to get the transfer done.

  If the remote server is reachable through HTTP, then the client should send HTTP requests where targets are in _absolute form_ (see [RFC 9112](https://www.rfc-editor.org/rfc/rfc9112#name-absolute-form)).
  
  > When making a request to a proxy, other than a CONNECT or server-wide OPTIONS request (as detailed below), a client MUST send the target URI in "absolute-form" as the request-target.

  > `GET http://www.example.org/pub/WWW/TheProject.html HTTP/1.1`
  
  The first step is to thread an `~absolute_form` parameter when building requests. This way, the `Cohttp_lwt_unix.Connection_cache` module can simply be extended with an optional `?proxy` parameter, taking the URI of the proxy. If a proxy is specified, all requests will be send in the absolute form, and instead of connecting to the remote server, the client will transparently connect to the proxy. Conduit handles the connection to an HTTPS proxy automatically.
  
- [HTTPS with HTTP proxy](https://everything.curl.dev/usingcurl/proxies/http.html#https-with-http-proxy)

  If the remote server uses HTTPS, then a tunneling proxy needs to be used. The CONNECT HTTP requests asks the proxy to open a connection to the remote server. If it succeeds, the proxy then blindly forwards data between the client and the remote server. Conduit needs to be modified to handle an HTTPS connection (more specifically, a TLS handshake and encryption layer) on top of an already opened connection. If the proxy itself is reachable through HTTPS, conduit needs to be able to tunnel a TLS connection on top of another TLS connection.
  
  To handle tunneled client connections, we provide the (private) `Cohttp_lwt.Connection_tunnel`, which takes the proxy URI when created. It then maintains a cache of opened connections via the proxy to each remote server, and is able to distribute new requests amongst currently opened connections (the code is mostly copied from `Cohttp_lwt.Connection_cache` balancing). 
  
  We expose the `Cohttp_lwt.Connection_proxy` via `Cohttp_lwt_unix.Connection_proxy` which, based on the remote URI, selects either a direct (absolute-form queries) or tunneled connection.
  
- [MITM proxy](https://everything.curl.dev/usingcurl/proxies/mitm.html)
  
  > [MITM proxies] require users to install a custom "trust root" (Certificate Authority (CA) certificate) in the client, and then the proxy terminates all TLS traffic from the client, impersonates the remote server and acts like a proxy.
  
  This can be achieved by installing a new certificate system-wide, or passing it to the ca-certs library via external means (environment variable, parameter at startup, ...). See https://github.com/mirage/ca-certs/pull/30.
  
- [Proxy environment variables](https://everything.curl.dev/usingcurl/proxies/env.html)

  We show how to support `ALL_PROXY`, `NO_PROXY`, and `<scheme>_proxy` (such as `http_proxy`, `https_proxy`), as documented by the curl project.
  
This PR relies on tunneling the TLS-on-TLS feature being added to ocaml-tls and Conduit. We've added some opam pin-depends to allow it to build with the latest dependencies.

- [x] https://github.com/mirage/ocaml-conduit/pull/428
- [x] https://github.com/mirleft/ocaml-tls/pull/499
- [ ] https://github.com/mirage/ca-certs/pull/30 (optional)


rgrinberg:
I don't think you should modify the `Http.Request.t` type. The point of this type is to stay minimal and not reflect any application specific concerns.

MisterDA:
> I don't think you should modify the `Http.Request.t` type. The point of this type is to stay minimal and not reflect any application specific concerns.

We're only adding a boolean, `absolute_form`, which I'd argue is a property of the request. If we don't add it, `write_header` in `Cohttp.Request.Make` has no way of knowing whether the request needs to be formatted in absolute form.

MisterDA:
> EDIT: after glancing at the spec, seems like 1. is the way to go. the `resource` field should already handle requests in "absolute form"

I did not follow this option, because the `resource` field would then need to duplicate the scheme and the host, which are already stored in the request record. I can try this and see what happens.

rgrinberg:
Good point, doesn't that mean we can get rid of scheme from the request as well? @anuragsoni why did we add scheme?

anuragsoni:
> why did we add scheme?

From what I remember we kept it around mostly to maintain compatibility with the existing release of cohttp.

rgrinberg:
@MisterDA do you want to take the honors of getting rid of it?

cohttp-lwt-unix/examples/client_lwt_proxy.ml
```diff
@@ -0,0 +1,146 @@
...
+  ]
+
+(* Boilerplate code to handle redirects *)
+
+let rec http_get_and_follow ~max_redirects ?headers uri =
+  let open Lwt.Syntax in
+  let* ans = Cohttp_lwt_unix.Client.get ?headers uri in
+  follow_redirect ~max_redirects ?headers uri ans
+
+and follow_redirect ~max_redirects ?headers request_uri (response, body) =
```
samoht, line 42: doesn't need to be done for this PR but that would be nice to have a similar function in the client API (so that every user do not have to re-implement an half-broken version of it)

MisterDA:
> @MisterDA do you want to take the honors of getting rid of it?

It doesn't seem possible to remove the `scheme` field of the request, as `Cohttp.Request.uri` needs to be able to reconstruct the full URI from a request, as exhibited by the test:

https://github.com/mirage/ocaml-cohttp/blob/53fbf397c284bee029a1c6f0bf00e4e6983c0cd2/cohttp/test/test_request.ml#L251-L255

we might want to remove this feature altogether. This mostly impacts logging and testing, but also the `Cohttp_lwt.Client.callv` function (that is deprecated).

rgrinberg:
Indeed it's not worth it. I got rid of it here https://github.com/mirage/ocaml-cohttp/pull/1086

http/test/test_parser.ml
```diff
@@ -162,6 +162,26 @@ let parse_result_notifies_start_of_body () =
...
+    make_req
+      ~headers:
+        (Http.Header.of_list
+           [
+             ("Host", "example.com");
+             ("Proxy-Authorization", "Basic dXNlcjpwYXNz");
+           ])
+      `GET "http://example.com/foo.html"
+  in
+  assert_req_success ~here:[ [%here] ] ~expected_req ~expected_consumed:104 buf
```
rgrinberg, line 183: Could this test come in a separate PR? AFAIK, nothing in this PR was added to make this pass.

MisterDA: Extracted to #1089.

ajbt200128:
Hey wondering if there was any work left here besides the merge conflicts?

samoht:
LGTM - happy to merge this once rebased

rgrinberg:
I was waiting to do another round of review after this PR went past the "draft" stage

samoht:
@rgrinberg - ha ok! feel free to dismiss my comment and make a proper round of review then :-)

I'm still interested in moving the helper functions to manage redirects into the lib, but I am not sure where exactly.

ajbt200128:
@samoht I think @rgrinberg is waiting for this PR to no longer a draft. Is there anything I can help do to get this out of the draft stage?

rgrinberg (approved):

cmellazchy:
Appreciate all the effort put into this. These changes are really valuable to us as well. So, I just wanted to check in on the status. Thanks again!

samoht:
@MisterDA @art-w - do you think you could fix the CI errors? I'm happy to merge and release it now that @rgrinberg has approved it.

MisterDA:
The CI errors will be fixed when core_unix 0.18 is released.

samoht:
What about the lint errors:

```
/src: (run (shell "opam exec -- opam-dune-lint"))
cohttp-async.opam: OK
cohttp-bench.opam: OK
cohttp-curl-async.opam: OK
cohttp-curl-lwt.opam: OK
cohttp-curl.opam: OK
cohttp-eio.opam: OK
cohttp-lwt-jsoo.opam: OK
cohttp-lwt-unix.opam: OK
cohttp-lwt.opam: changes needed:
"ipaddr" {>= "5.6.0"}                    [from cohttp-lwt/src]
cohttp-mirage.opam: OK
cohttp-server-lwt-unix.opam: OK
cohttp-top.opam: OK
cohttp.opam: OK
http.opam: OK
Note: version numbers are just suggestions based on the currently installed version.
```

Would it be possible to add a bound somewhere (maybe to the version of core in opam-repo) to fix the CI issues?

cohttp-async.opam
```diff
@@ -25,7 +25,7 @@ doc: "https://mirage.github.io/ocaml-cohttp/"
 bug-reports: "https://github.com/mirage/ocaml-cohttp/issues"
 depends: [
   "dune" {>= "3.8"}
-  "ocaml" {>= "4.14"}
+  "ocaml" {>= "4.14" & < "5.3.0"}
```
samoht, line 28: what's the issue with OCaml 5.3 here?

art-w: This is unfortunate but [`core_unix.v0.17` fails to build with OCaml 5.3](https://github.com/janestreet/core_unix/issues/14) but we also can't force an earlier version of `core_unix` as they all require OCaml <5.3, so [the CI then complains that there are no solutions for 5.3](https://github.com/mirage/ocaml-cohttp/actions/runs/13214152712/job/36891404090) (as it is specifically trying to test that version of OCaml, unless we tell it that <5.3 is expected). We would still hit this "no solution" issue if `core_unix.v0.17` was flagged as `ocaml<5.3` in the opam-repo, so I don't see how to lift that ocaml constraint until a new release of `core_unix` is published (but I'm open to suggestions!)

Issue #1032, opened 2024-06-05
Parsing of malformed headers fails silently


askvortsov1:
The [header parsing implementation](https://github.com/mirage/ocaml-cohttp/blob/master/cohttp/src/header_io.ml#L36) returns early if it encounters a malformed header. It doesn't communicate any error info to the caller, and presumably the unparsed lines are interpreted as part of the body.

Athishpranav2003:
I guess i can check this out in my free time

Athishpranav2003:
@askvortsov1 can you list some kinda malformed headers for reference?

Athishpranav2003:
```
The request/status line and headers must all end with <CR><LF> (that is, a [carriage return](https://en.wikipedia.org/wiki/Carriage_return) followed by a [line feed](https://en.wikipedia.org/wiki/Line_feed)). The empty line must consist of only <CR><LF> and no other [whitespace](https://en.wikipedia.org/wiki/Whitespace_(computer_science)).
```
I got this in wiki and few more places(which referenced RFCs. I guess i will go with this as the standard

PR #1082, opened 2024-08-09
Added check to parser header till trailing \r\n

Athishpranav2003:
This PR addresses #1032 . 
Here we basically raise Exception whenever we find a incorrect section while parsing headers.

### DIFF
--- a/cohttp/src/header_io.ml
+++ b/cohttp/src/header_io.ml
@@ -16,6 +16,8 @@
  *
   }}}*)
 
+exception Invalid_headers
+
 let split_header str =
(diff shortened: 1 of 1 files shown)


Athishpranav2003:
@rgrinberg can you please check this

rgrinberg:
Why do we need a separate exception for invalid headers? Why can't we just raise whatever we raise when we fail to parse the request for any other reason? E.g. `Content-Length` isn't a valid number for example.

Athishpranav2003:
@rgrinberg i cannot find the exception we raise for such issues. Could you please show which exception are you referring to

PR #1086, opened 2024-09-04
feature: remove [scheme] from requests

rgrinberg:
This makes it so that [Request.make ~uri |> Request.uri] will no longer return the same URI as [uri]. Also, this property was never preserved with respect to other URI fields.

cc @MisterDA 

### DIFF
--- a/cohttp/src/request.ml
+++ b/cohttp/src/request.ml
@@ -19,29 +19,24 @@ open Sexplib0.Sexp_conv
 type t = Http.Request.t = {
   headers : Header.t;
   meth : Code.meth;
-  scheme : string option;
   resource : string;
   version : Code.version;
 }
 [@@deriving sexp]
 
-let compare { headers; meth; scheme; resource; version } y =
+let compare { headers; meth; resource; version } y =
   match Header.compare headers y.headers with
   | 0 -> (
       match Code.compare_method meth y.meth with
       | 0 -> (
-          match Option.compare String.compare scheme y.scheme with
-          | 0 -> (
-              match String.compare resource y.resource with
-              | 0 -> Code.compare_version version y.version
-              | i -> i)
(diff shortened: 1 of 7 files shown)


cohttp/src/request.ml
```diff
@@ -110,49 +104,42 @@ let is_valid_uri path meth =
...
+  | path -> (
+      let uri = Uri.of_string path in
+      match Uri.scheme uri with
+      | Some _ -> (
+          Uri.(
+            (* we have an absoluteURI *)
+            match path uri with "" -> with_path uri "/" | _ -> uri))
+      | None -> (
+          let empty = Uri.of_string "" in
+          let empty_base = Uri.of_string "///" in
```
avsm, line 125: A Uri.t is a fairly heavy record - I just noticed both of these can be lifted into constants at the toplevel and not reallocated every request

rgrinberg: Indeed. I wonder if we should just get rid of this function altogether and just introduce separate functions to extract the Uri components that can be safely extracted out of the request.

avsm: Probably just better to 'deallocationise' this library separately. I actually wanted to use it to test out the JS with-extensions tree to see how well local annotations would do in some places here...

rgrinberg: Sure, though I think this function cannot be be used in a low allocation context anyway. Many of the components this function is returning can be extracted without any allocation at all. While the `Uri.t` this function returns will always be allocated - even if it is only done once.

avsm: Yes, it would need changing the Uri library as well... the accessors there could be more lazy. That library's design hasn't changed for about a decade -- about high time for a refresh!

MisterDA (approved):

PR #1084, opened 2024-08-29
chore: fix CI

art-w:
There was a number of breaking changes in the recent releases of `tls` and `mirage-crypto`. Furthermore `ppx_expect` bump to v0.17.0 in https://github.com/mirage/ocaml-cohttp/pull/1043 requires ocaml >=5.1.0, so github actions couldn't run the tests anymore.

PR #1087, opened 2024-09-07
remove use of Sexplib (Sexplib0 has the same function)

copy:
I think with this change the bounds on `conduit-lwt<"7.0.0"` can be removed  (they were added in https://github.com/ocaml/opam-repository/pull/26463/commits/f77632c3736d2025f17a60857140dc17bb3af879


copy:
Just noticed that this change is already included in https://github.com/mirage/ocaml-cohttp/pull/1084, closing.

PR #1091, opened 2024-10-02
cohttp-eio: Use system authenticator in example

talex5:
It turns out that using a proper authenticator is quite easy, so I've updated the example to use that. Otherwise, people are likely to copy the null authenticator code (despite the warning in the comment).


MisterDA (approved):

talex5:
I can merge it myself, but I like to wait for a review if possible (thanks @MisterDA!).

PR #1096, opened 2024-11-21
6.0.0 release

avsm:
A few housekeeping changes:

- bump dune lang to 3.8 to get a reasonable mdx bound
- this brings in some new default warnings, so remove unused names in signatures
- bytebuffer has an unnecessary mutable annotation in the implementation, as Bytes is already mutable


samoht:
LGTM

PR #1107, opened 2025-03-06
Add maintenance intent

### LINKED ISSUE #1102: Add x-maintenance-intent to your opam files
Dear Madam or Sir,

due to the opam repository archival process, we are keen to have the maintainers of packages specify their intention. We noticed that cohttp has a lot of releases, and are curious what your maintenance intention is.

You can read up on the policy at https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md and the `x-maintenance-intent` field at https://github.com/ocaml/opam-repository/blob/master/governance/policies/archiving.md#specification-of-the-x--fields-used-in-the-archiving-process

Thanks a lot for taking this into consideration. Please note that adding `x-maintenance-intent` in the latest release is sufficient (please feel free to PR such a field directly to the opam-repository).

Also note that if there exists a package that is to be retained in opam-repository, and this depends on a specific cohttp version (not part of the maintenance-intent), this cohttp version is kept -- so we don't end up in uninstallable packages.

samoht:
Fix #1102

### DIFF
--- a/cohttp-async.opam
+++ b/cohttp-async.opam
@@ -63,3 +63,4 @@ build: [
   ]
 ]
 available: arch != "s390x"
+x-maintenance-intent: [ "(latest)" ]
--- a/cohttp-bench.opam
+++ b/cohttp-bench.opam
@@ -51,3 +51,4 @@ build: [
     "@doc" {with-doc}
   ]
 ]
+x-maintenance-intent: [ "(latest)" ]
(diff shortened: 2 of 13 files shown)


samoht:
@anmonteiro any idea how to fix the nix issue?

hannesm:
There's an issue with this: you use dune-project to generate opam files. For the time being, you'll need to carry `xxx.opam.template` files around with the `x-maintenance-intent: [ "(latest)" ]` -- see e.g. https://github.com/mirage/ocaml-xenstore/pull/57

hannesm:
Also, you missed http.opam.

samoht:
I didn't know JS needed 2.5 support -- @rgrinberg is this the case?

PR #1113, opened 2025-06-08
Fix `Connection_cache.call` hanging on unreachable server

### LINKED ISSUE #1112: Cohttp_lwt_unix.Connection_cache.call hangs if the server is unreachable
Assuming no service is bound to `127.0.0.1:55000`, the following minimal snippet will hang indefinitely.

```ocaml
  let cache = Cohttp_lwt_unix.Connection_cache.create ~parallel:1 () in
  Cohttp_lwt_unix.Connection_cache.call cache `GET
    (Uri.of_string "http://127.0.0.1:55000")
```

I quick look at the functor in `Cohttp_lwt.Connection_cache` and I guess the `get_connection` recursive function is to blame? I’ll try to dig deeper, but no matter the result of my search, I wanted to share my discovery.

lthms:
* What

The `Connection_cache.call` function could hang indefinitely when attempting to establish a connection to an unreachable server. This change resolves that issue by ensuring that the connection creation process is asynchronous and non-blocking.

Fixes #1112 

* Why

Previously, the `Connection.create` function returns before the connection is established. When the returned connection is binded, something in the code of `get_connection` prevents the correct handling of the `ECONNREFUSED` Unix error.

* How

The `create` helper function now uses `Connection.connect` (which returns an Lwt promise) and binds its result using `>>=`. This transforms the `create` function itself into an Lwt promise. Consequently, the `Connection_cache.call` function no longer blocks on connection establishment but rather awaits the connection promise.

### DIFF
--- a/cohttp-lwt/src/connection_cache.ml
+++ b/cohttp-lwt/src/connection_cache.ml
@@ -74,9 +74,9 @@ module Make (Connection : S.Connection) (Sleep : S.Sleep) = struct
       Lwt.return_unit
     in
     let create () =
-      let connection =
-        Connection.create ~persistent:true ~finalise ~ctx:self.ctx endp
-      and timeout = ref Lwt.return_unit in
+      Connection.connect ~persistent:true ~finalise ~ctx:self.ctx endp
+      >>= fun connection ->
+      let timeout = ref Lwt.return_unit in
       let rec busy () =
         Lwt.cancel !timeout;
         if Connection.length connection = 0 then (
@@ -88,11 +88,11 @@ module Make (Connection : S.Connection) (Sleep : S.Sleep) = struct
         Lwt.on_termination (Connection.notify connection) busy
       in
       busy ();
-      connection
+      Lwt.return connection
     in
     match Hashtbl.find_all self.cache endp with
     | [] ->
-        let connection = create () in
+        create () >>= fun connection ->
         Hashtbl.add self.cache endp connection;
         Lwt.return connection
     | conns -> (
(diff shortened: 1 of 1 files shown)


samoht:
The change looks fine to me but /cc @art-w and @MisterDA who worked on this piece the most recently.

samoht (approved):

cohttp-lwt/src/connection_cache.ml
```diff
@@ -74,9 +74,9 @@ module Make (Connection : S.Connection) (Sleep : S.Sleep) = struct
       Lwt.return_unit
     in
     let create () =
-      let connection =
-        Connection.create ~persistent:true ~finalise ~ctx:self.ctx endp
-      and timeout = ref Lwt.return_unit in
+      Connection.connect ~persistent:true ~finalise ~ctx:self.ctx endp
+      >>= fun connection ->
```
art-w: This looks a bit weird, we are blocking on establishing the connection and only then setting up a timeout... I believe this works if we get the `ECONNREFUSED` immediate error but otherwise we would still hang with no control on the timeout?

Looking at the code around https://github.com/mirage/ocaml-cohttp/blob/main/cohttp-lwt/src/connection.ml#L291 , I wonder if the issue isn't larger than the cache: `Connection.t` has two queues of `Lwt.wait()` promises ("waiting" and "in flight"), but when an error happens with the channel, it looks like we don't signal the error to the queues? (leaving their `Lwt.u` unresolved)

Even if `ECONNREFUSED` wasn't propagated correctly, I was confused why the timeout below didn't trigger, but `Connection.close` seems to have the same issue of leaving the queues unresolved such that `Connection.call` is stuck hanging even though the timeout did close the connection (?)

lthms: Okay, I don’t know when but i’ll try to have a look.

I’m not sure to understand _who_ is maintaining cohttp :sweat_smile:

PR #1114, opened 2025-07-19
Undo deprecation of connection value

ansiwen:
In #798 the connection value has been deprecated with the claim it is useless.  However, in fact it is used at least with the Mirage backend in order to track the connections for logging etc.  (See last two comments of the PR.) In over 3 years no replacement has been added, so this change removes the deprecation until some alternative exists.  (Which could be possibly Mirage only.)


avsm:
Thanks, I'm happy with removing the deprecation until the actual replacement is present.

PR #1118, opened 2025-09-11
Client: Expose `IO`.

mefyl:
The lwt `IO` implementation defines a private `IO_error` exception, which can only be handled via the `catch` method. Yet since the `IO` module is not exposed, when working with an agnostic `Cohttp.Client.S`, there is AFAICT no way to actually stop these exceptions.

This PR exposes the `IO` module so that it's possible to properly handle all errors with zero assumptions on the IO implementation. I suppose it also doesn't hurt to be able to write cohttp code that works with any IO monad.


samoht:
Could you rebase this on main? Seems that ocaml-ci is complaining about unrelated fixes

PR #1119, opened 2025-10-17
Update code for Cmdliner

Leonidas-from-XIV:
I tried building with a fresh switch but it failed with an issue due to Cmdliner.

Cmdliner 2.0 made `Cmdliner.Arg.conv` abstract, so constructing it via a tuple of `parser * printer` is no longer possible. Instead it provides `Cmdliner.Arg.Conv.make`, so this changes the code to use exactly that.

The code also adds the cmdliner dependency in a package where it was missing before and bumps the requirement to use Cmdliner 2.0.0 since that's where `Cmdliner.Arg.Conv.make` was introduced.

### DIFF
--- a/cohttp-curl-lwt.opam
+++ b/cohttp-curl-lwt.opam
@@ -27,7 +27,7 @@ depends: [
   "cohttp-curl" {= version}
   "stringext"
   "lwt" {>= "5.3.0"}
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV:
Fair enough. Is there an opam variable that can express this? Should I mark it as `with-dev-setup`? `with-test`?

Issue #1120, opened 2025-10-23
Support proxying client requests in cohttp-eio


shonfeder:
@Leonidas-from-XIV and I have begun work towards extending cohttp-eio to support proxying client requests, reflecting the earlier work by MisterDA in https://github.com/mirage/ocaml-cohttp/pull/1080 to bring this functionality to cohttp-lwt.

Our current plan is to follow @MisterDA's design and implementation as closely as possibly, since this has already set a precedent for the expected API, which is in use in the wild.

We are opening this issue to track our work, share design considerations as they develop, and invite any feedback or preliminary guidance from maintainers.

PR #1126, opened 2025-11-12
Add support for proxying client requests in the cohttp-eio client

shonfeder:
Closes https://github.com/mirage/ocaml-cohttp/issues/1120

## Forward proxy support for the cohttp-eio client

This builds on #1125 to add proxy support the cohttp-eio client for one-shot requests. Once the prerequisite changes, the examples, the readme, and the refactoring is subtracted, this amounts to a small addition on top of #1125.

Configuration of the proxy has been added as an atomic ref of global state. This is generally discouraged be Eio's design, but this state is expected to be set only once on program initialization, and it allows users a simple interface for configuring proxies, even when using libraries that don't expose the client creating logic. Of course, if maintainers prefer I explore providing this configuration by some other means, I'm happy to make changes.

## API change for the `Client.t` constructors type

In order to support use of proxies over https, we need TLS over TLS, which means applying a provided https wrapper to a flow twice (once before, and once after creating the tunnel). Fortunately, this is easily achieved using Eio flows. 

However, this requires narrowing the types of the flow produced by `Client.make_generic`'s socket creation function, and the `https` wrapper taken by `Client.make`. While this is technically a breaking API change, we do not expect users of the library to actually hit the breakage or require code changes.

Thanks to @talex5 and especially @art-w for key pointers in working out the needed typing. 

## An example client using the proxy support

This change set includes an executable that exercises the proxying functionality of the client library. Most of the code in the executable was written by @Leonidas-from-XIV.

A readme has also been added to the examples directory, along with an explanation of how how to run the `client_proxy.exe` to exercise some of the functionality.

## Unit tests

I have added a few unit tests for the core functionality here in a followup pr: https://github.com/mirage/ocaml-cohttp/pull/1129

cc @samoht

### DIFF
--- a/cohttp-eio/examples/client_proxy.ml
+++ b/cohttp-eio/examples/client_proxy.ml
@@ -1,0 +1,196 @@
+open Cohttp_eio
+
+let authenticator =
+  match Ca_certs.authenticator () with
+  | Ok x -> x
+  | Error (`Msg m) ->
+      Fmt.failwith "Failed to create system store X509 authenticator: %s" m
+
+let () =
+  Logs.set_reporter (Logs_fmt.reporter ());
+  Logs_threaded.enable ();
+  Logs.Src.set_level Cohttp_eio.src (Some Debug)
+
+let https ~authenticator =
+  let tls_config =
+    match Tls.Config.client ~authenticator () with
+    | Error (`Msg msg) -> failwith ("tls configuration problem: " ^ msg)
+    | Ok tls_config -> tls_config
+  in
+  fun uri socket ->
+    let host =
+      Option.bind (Uri.host uri) (fun x ->
+          Domain_name.(host (of_string_exn x)) |> Result.to_option)
+    in
+    Tls_eio.client_of_flow ?host tls_config socket
+
+let get_request_exn ~sw client url =
+  let resp, body = Client.get ~sw client url in
+  match resp.status with
+  | `OK ->
+      Eio.traceln "%s"
+      @@ Eio.Buf_read.(parse_exn take_all) body ~max_size:max_int
+  | otherwise -> Fmt.epr "Unexpected HTTP status: %a\n" Http.Status.pp otherwise
+
+let run_client url cacert all_proxy no_proxy http_proxy https_proxy proxy_auth =
+  let scheme_proxy =
+    List.filter_map Fun.id
+      [
+        Option.map (fun p -> ("http", p)) http_proxy;
+        Option.map (fun p -> ("https", p)) https_proxy;
+      ]
+  in
+  let proxy_headers =
+    Option.map
+      (fun credential ->
+        Http.Header.init_with "Proxy-Authorization"
+          (Cohttp.Auth.string_of_credential credential))
+      proxy_auth
+  in
+
+  Eio_main.run @@ fun env ->
+  Mirage_crypto_rng_unix.use_default ();
+
+  let net = env#net in
+
+  let authenticator =
+    match cacert with
+    | None -> authenticator
+    | Some pem ->
+        (* Load a custom cacert from a file *)
+        let fs = Eio.Stdenv.fs env in
+        X509_eio.authenticator (`Ca_file Eio.Path.(fs / pem))
+  in
+
+  Client.set_proxies ?proxy_headers ?default_proxy:all_proxy
+    ?no_proxy_patterns:no_proxy ~scheme_proxies:scheme_proxy ();
+
+  let client = Client.make ~https:(Some (https ~authenticator)) net in
+
+  Eio.traceln ">>> Make calls in sequence";
+  Eio.Switch.run (fun sw ->
+      get_request_exn ~sw client url;
+      get_request_exn ~sw client url;
+      get_request_exn ~sw client url);
+
+  Eio.traceln ">>> Make calls concurrently";
+  Eio.Switch.run (fun sw ->
+      for _ = 0 to 5 do
+        Eio.Fiber.fork ~sw (fun () -> get_request_exn ~sw client url)
+      done);
+
+  Eio.traceln ">>> Make calls in parallel";
+  let dm = Eio.Stdenv.domain_mgr env in
+  Eio.Fiber.all
+    [
+      (fun () ->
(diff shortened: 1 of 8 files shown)


cohttp-eio/src/client.mli
```diff
@@ -2,18 +2,19 @@ open Eio.Std
...
      and type 'a io = 'a
      and type body = Body.t
 
-val make :
-  https:
-    (Uri.t -> [ `Generic ] Eio.Net.stream_socket_ty r -> _ Eio.Flow.two_way)
-    option ->
-  _ Eio.Net.t ->
-  t
+val make : https:(Uri.t -> connection -> connection) option -> _ Eio.Net.t -> t
```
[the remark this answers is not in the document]
shonfeder: Your patch worked perfectly. Thanks, @art-w !

shonfeder:
> So changing the type seems reasonable, but I'm surprised it needs to be wrapped. It probably just needs an extra cast somewhere (e.g. casting the result of Eio.Net.connect to two-way+close).

You were right, @talex5 ! Think I just hit a wall last night. Thanks for the guidance :pray:

cohttp-eio/src/client.ml
```diff
@@ -1,8 +1,73 @@
...
+let get_proxy uri =
+  match Atomic.get proxies with
+  | None -> None
+  | Some (headers, proxies) -> (
+      match Proxy.get proxies uri with
+      | None -> None
+      | Some (Proxy.Direct _) as proxy -> proxy
+      | Some (Proxy.Tunnel p) -> Some (Proxy.Tunnel (headers, p)))
+
+let call_on_socket ~sw ?headers ?body ?(chunked = false) meth uri socket =
```
shonfeder: This is factored out of the `call` function on line 82 with no changes.

cohttp-eio/src/client.ml
```diff
@@ -80,25 +104,56 @@ let tcp_address ~net uri =
...
+(* Create a socket for the uri, and signal whether it requires https *)
+let scheme_conn_of_uri ~sw net uri =
+  match Uri.scheme uri with
+  | Some "httpunix" ->
+      (* FIXME: while there is no standard, http+unix seems more widespread *)
+      `Plain (Eio.Net.connect ~sw net (unix_address uri) :> connection)
+  | Some "http" ->
+      `Plain (Eio.Net.connect ~sw net (tcp_address ~net uri) :> connection)
+  | Some "https" ->
+      `Https (Eio.Net.connect ~sw net (tcp_address ~net uri) :> connection)
```
shonfeder: We move to this static signal via the `Https` variant, instead of applying the https wrap directly because, when using a proxy over http, we first need access to the socket flow unencrypted to set up the tunnel. But then we still need to wrap the socket in our tls encryption once the connection is established.

cohttp-eio/src/client.ml
```diff
@@ -80,25 +104,55 @@ let tcp_address ~net uri =
   | ip :: _ -> ip
   | [] -> failwith "failed to resolve hostname"
 
+(* Create a socket for the uri, and signal whether it requires https *)
+let scheme_conn_of_uri ~sw net uri =
+  match Uri.scheme uri with
+  | Some "httpunix" ->
+      (* FIXME: while there is no standard, http+unix seems more widespread *)
```
shonfeder: This FIXME is from the existing code.

cohttp/src/proxy.mli
```diff
@@ -32,15 +32,15 @@ module Forward : sig
         calls with that scheme.
 
       @param direct
-        A function to create ['direct] connections for the given proxy uri.
+        A function to create ['direct] connections for the given proxy URI.
 
       @param tunnel
-        A function to create ['tunnel] connections for the given proxy uri. *)
+        A function to create ['tunnel] connections for the given proxy URI. *)
```
shonfeder: These were followup from review requests in #1125

art-w (approved):
This looks good! I would instinctively prefer if proxies didn't use a global and were configured when creating a `Client.t`, but this wouldn't allow transitive dependencies to inherit the proxies (... but then I think those deps shouldn't be creating `Client.t` themselves, but instead ask their caller to provide it).

cohttp-eio/src/client.ml
samoht: Where is the HTTP -> HTTPS promotion happening? Is this in the pure proxy code?

cohttp-eio/examples/client_proxy.ml
```diff
@@ -0,0 +1,196 @@
...
+  in
+  let proxy_headers =
+    Option.map
+      (fun credential ->
+        Http.Header.init_with "Proxy-Authorization"
+          (Cohttp.Auth.string_of_credential credential))
+      proxy_auth
+  in
+
+  Mirage_crypto_rng_unix.use_default ();
```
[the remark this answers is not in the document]
shonfeder: I think this initially followed the example `examples/client_tls.ml`:

https://github.com/mirage/ocaml-cohttp/blob/bb0f2c4de3f9aba1e1f6abd2c67e14a36bc7dc02/cohttp-eio/examples/client_tls.ml#L27-L31

But I see now it is neither in the `Eio_main.run` nor at the top level. Do you have a preference which way it goes?

cohttp-eio/src/client.ml
samoht: No it's ok.

cohttp-eio/examples/client_proxy.ml
```diff
@@ -48,8 +48,6 @@ let run_client url cacert all_proxy no_proxy http_proxy https_proxy proxy_auth =
       proxy_auth
   in
 
-  Mirage_crypto_rng_unix.use_default ();
```
samoht: actually it was probably better defined close the Eio_main.run call. Sorry to nitpick on this...

shonfeder: No worries. I but it right under the main now in a fixup, matching the position from 

https://github.com/mirage/ocaml-cohttp/blob/bb0f2c4de3f9aba1e1f6abd2c67e14a36bc7dc02/cohttp-eio/examples/client_tls.ml#L27-L31

PR #1127, opened 2025-11-16
Fix nix build

shonfeder:
WIP to try to fix the failing nix builds on main reported in https://github.com/mirage/ocaml-cohttp/issues/1128

I have fixed the initial problem, and a subsequent error from an out of date ocamlformat version. Now troubleshooting the cmdliner dep, which is why this is still in draft.

The current problems is that nix is selecting the wrong version of cmdliner:

```
File "cohttp-lwt-unix/bin/cohttp_curl_lwt.ml", line 68, characters 4-26:
68 |     Cmdliner.Arg.Conv.make ~parser ~pp ~docv:"URI" ()
         ^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module "Cmdliner.Arg.Conv"
(cd _build/default/cohttp-async/test && ./test_async_integration.exe)
```

```
cat /nix/store/nkfll7c4fv8zamvb92jqv9rrli5wmdfw-cmdliner-1.3.0/lib/ocaml/5.2.1/site-lib/cmdliner/META
description = "Declarative definition of command line interfaces for OCaml"
version = "1.3.0"
```

But I have not found any indication of what is required to change the cmdliner version on the nix side. 

It looks like @rgrinberg and @anmonteiro have set up the nix stuff. Can either of you advise on what more is needed to fix the CI here?

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -23,11 +23,11 @@
         "nixpkgs": "nixpkgs_2"
       },
       "locked": {
-        "lastModified": 1734153276,
-        "narHash": "sha256-/cvtpMFp0HArEpFi0PrPMsheauc3IJ7qWpSHnw8so2M=",
+        "lastModified": 1762899092,
+        "narHash": "sha256-Nl6547Q+Hw+RDV7cQnPmP6OTS7kncj9aN2Zx+HTlIeg=",
         "owner": "nix-ocaml",
         "repo": "nix-overlays",
-        "rev": "4247b28ce426ccdea09a1ec014fa52785bc7ba1d",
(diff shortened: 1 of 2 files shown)


shonfeder:
On [#general > Nix help @ 💬](https://ocaml.zulipchat.com/#narrow/channel/405353-general/topic/Nix.20help/near/556942607) I got a helpful pointer that the problem is cmdliner 2 not being available in nixpkgs (and I guess also not being the overlay used in the flake here).

That should point the way to getting a fix. But maybe we should just remove the flake? It has been breaking for a long time and I'm not sure who is using it....

samoht:
@anmonteiro any opinion?

PR #1137, opened 2025-12-23
Fix uncatchable IO exceptions.

mefyl:
There is currently a bug where IO write exception are "uncatchable": even if the underlying `IO_error` is handled, it would still find its way out and kill the main loop as demonstrated in [this test](141253d724d95d74ddca3b9e1439b5004a1f95de).

Bamboozled by the backtrace, I initially [reported this in lwt](https://github.com/ocsigen/lwt/issues/1067) and lived with it for some time. I finally had enough and jumped into the rabbit hole, copiously hitting my head every meter during the fall, but finally hunted it down to [this Lwt.dont_wait](https://github.com/mirage/ocaml-cohttp/commit/7150778c57f56e17315fe589d6a9d8da840c1dc8). Cruel irony, I already [addressed this very issue](https://github.com/mirage/ocaml-cohttp/pull/995) but was too shy. This is the harsher version, which I think is the correct one:

* No exception escaping `res` should be reported twice.
* AFAICT the only other thing that could raise is `Connection.close`, but I don't think any exception there should kill the Lwt loop either.

### DIFF
--- a/cohttp-lwt-unix/test/test_client.ml
+++ b/cohttp-lwt-unix/test/test_client.ml
@@ -203,6 +203,24 @@ let test_client_cached uri =
   Client.set_cache (Cache.call cache);
   test_client uri
 
+(* Check an IO exception can be properly handled. *)
+let test_exception uri =
+  let open struct
+    exception Dummy_io_error
+  end in
+  let call () =
+    Client.put
+      ~body:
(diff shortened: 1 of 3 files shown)


mefyl:
Superseded  by #1140

PR #1141, opened 2026-06-01
Don't add Transfer-Encoding to responses that cannot have a body.

mefyl:
RFC 7230 §3.3.1 forbids Transfer-Encoding in 1xx, 204, and 304 responses. The previous cohttp 6.x change that defaulted Unknown encoding to Chunked did not account for these response types, causing e.g. WebSocket 101 Switching Protocols responses to include a spurious Transfer-Encoding: chunked header and a trailing chunked body terminator (`0\r\n\r\n`), breaking the WebSocket handshake for conforming clients.


rgrinberg:
I think the tests need to be updated?

avsm:
Rebased on main, and I've added a bunch of tests to cover some of the body allowed variations

PR #1142, opened 2026-07-23
cohttp-eio: update tests for new Eio 1.4 getaddrinfo exception

talex5:
Trying to connect to "http://foo.invalid" now raises `Eio.Io Net Address_lookup_failed`, not `Failure`.

Before (`getaddrinfo` returns `[]` and cohttp-eio raises an exception):
```
test_forward_proxy.exe: [INFO] GET http://foo.invalid
[failure] failed to resolve hostname
```
With Eio 1.4 `getaddrinfo` raises:
```
test_forward_proxy.exe: [INFO] GET http://foo.invalid
[exception] Eio.Io Net Address_lookup_failed NONAME (name or service is not known),
  looking up "foo.invalid" (service "http")
```


avsm:
lgtm, and backwards compatible with older eio

PR #1145, opened 2026-08-14
cohttp: urldecode before resolving path components for files

avsm:
This patch prevents escaping the docroot in resolve_local_file via urlencoded components. In order to prevent request path gating, it also exposes a new `normalise` function that can be called when splitting paths to apply the same logic; e.g.

```
let callback _conn req _body =
    let path = Cohttp.Path.normalise (Cohttp.Request.uri req) in
    match String.split_on_char '/' path with
    | "admin" :: _ when not (authorised req) -> Server.respond_not_found ()
    | _ -> Server.respond_file ~fname:path ()
```

This also applies the change to cohttp-mirage. The change affects cohttp-lwt, cohttp-mirage and cohttp-async, but not cohttp-eio which uses its own path resolution mechanism.

### DIFF
--- a/cohttp-mirage/src/static.ml
+++ b/cohttp-mirage/src/static.ml
@@ -37,35 +37,28 @@ module HTTP (FS : Mirage_kv.RO) (S : Cohttp_lwt.S.Server) = struct
     | Error e -> Fmt.failwith "exists %a" FS.pp_error e
 
   let dispatcher request_fn =
-    let rec fn fs uri =
-      match Cohttp.Path.normalise uri with
-      | "" ->
-          Logs.info (fun f -> f "request for '%s'" (Uri.path uri));
-          fn fs (Uri.with_path uri "index.html")
-      | path ->
-          Logs.info (fun f -> f "request for '%s'" path);
-          Lwt.catch
-            (fun () ->
-              read_fs fs path >>= fun body ->
-              let mime_type = Magic_mime.lookup path in
-              let headers = Cohttp.Header.init_with "content-type" mime_type in
-              let headers =
-                match request_fn with
-                | None -> headers
-                | Some fn -> fn uri headers
-              in
-              S.respond_string ~status:`OK ~body ~headers ())
-            (fun _exn ->
-              let with_index = Fmt.str "%s/index.html" path in
-              exists fs with_index >>= function
-              | true ->
-                  (* [with_path] expects an encoded path, and [path] has
-                     already been decoded by [normalise]. *)
-                  let encoded = Uri.pct_encode ~component:`Path with_index in
-                  fn fs (Uri.with_path uri encoded)
-              | false -> S.respond_not_found ())
+    let respond uri path body =
+      let mime_type = Magic_mime.lookup path in
+      let headers = Cohttp.Header.init_with "content-type" mime_type in
+      let headers =
+        match request_fn with None -> headers | Some fn -> fn uri headers
+      in
+      S.respond_string ~status:`OK ~body ~headers ()
     in
-    fn
+    fun fs uri ->
+      let path =
+        match Cohttp.Path.normalise uri with
+        | "" -> "index.html" | path -> path
+      in
+      Logs.info (fun f -> f "request for '%s'" path);
+      Lwt.try_bind
+        (fun () -> read_fs fs path)
+        (respond uri path)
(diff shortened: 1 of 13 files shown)


cohttp/test/test_path.ml
```diff
@@ -70,6 +70,50 @@ let test_resolve_local_file () =
...
+        "/foo/bar/baz/%2e%2e%2f%2e%2e%2fbuzz" );
+      ( "triply percent-encoded grandparent blocked",
+        "/foo/bar/baz",
+        "..%25252f..%25252fbuzz",
+        "/foo/bar/baz/..%252f..%252fbuzz" );
+      ( "encoded space preserved",
+        "/foo/bar/baz",
+        "/my%20file.txt",
+        "/foo/bar/baz/my file.txt" );
+      ( "encoded utf8 preserved",
```
avsm, line 97: I kept these tests as we experimented with stripping out control characters, but decided not to for compatibility. However, these tests form a useful base to change behaviour in a future major bump.

cohttp-mirage/src/static.ml
```diff
@@ -46,7 +46,8 @@ module HTTP (FS : Mirage_kv.RO) (S : Cohttp_lwt.S.Server) = struct
       | path when String.is_suffix ~affix:"/" path ->
```
mdales: Can't we still escape here, admittedly to a specific file, but I'd still expect the same normalisation to happen.

mdales (commented):
Looks okay, other than perhaps inconsistent application.

cohttp/test/test_path.ml
```diff
@@ -70,6 +70,50 @@ let test_resolve_local_file () =
...
+        "/foo/bar/baz",
+        "/caf%C3%A9.txt",
+        "/foo/bar/baz/caf\xc3\xa9.txt" );
+      ( "encoded NUL preserved",
+        "/foo/bar/baz",
+        "/secret%00.txt",
+        "/foo/bar/baz/secret\x00.txt" );
```
edwintorok, line 104: This is safe on Unix since OCaml 4.03, but not Lwt. I've opened https://github.com/ocsigen/lwt/issues/1124

cohttp-mirage/src/static.ml
avsm: Good catch, there was no escape I could repro but its a convoluted path. cb2916beee9de5ecea52e4436c8cbb59f1f4e533 now normalises for every request (which also drops an astring dependency)

cohttp/src/path.ml
```diff
@@ -1,7 +1,15 @@
...
+    | ("" | ".") :: rest -> go acc rest
+    | ".." :: rest -> (
+        match acc with [] -> go [] rest | _ :: tl -> go tl rest)
+    | seg :: rest -> go (seg :: acc) rest
   in
-  Filename.concat docroot rel_path
+  path |> String.split_on_char '/' |> go [] |> String.concat "/"
+
+let normalise uri =
+  uri |> Uri.path |> Uri.pct_decode |> remove_dot_segments
```
edwintorok, line 12: We fixed a similar vulnerability a bit differently in XAPI https://xenbits.xen.org/xsa/xsa271-xapi.patch: decode first, resolve dotdot next, and check that the fully resolved path is still inside the docroot: https://github.com/xapi-project/xen-api/pull/3705/changes

Although `remove_dot_segments` here is equivalent, since it doesn't allow any `../` to escape above the docroot.

cohttp-mirage/src/static.ml
```diff
@@ -46,7 +46,8 @@ module HTTP (FS : Mirage_kv.RO) (S : Cohttp_lwt.S.Server) = struct
...
           Logs.info (fun f -> f "request for '%s'" path);
           fn fs (Uri.with_path uri "index.html")
```
edwintorok: Unrelated to the fix here, but this looks wrong.
It turns `/foo/bar/` into `index.html`, and not `/foo/bar/index.html`.

edwintorok (approved):

cohttp-mirage/src/static.ml
avsm: Woops, good catch. I added in a mock test and rejigged the logic to simplify it in e3b229543db7e43fa305e5665c759722cd2dfcb0

PR #1146, opened 2026-08-15
Add write_bigstring_body

toots:
A body whose bytes already live outside the OCaml heap — a memory-mapped file, a buffer filled by a read — has to be copied into a string to be sent, which for a large body is exactly the allocation the caller was avoiding.

## The addition

One function per layer, and the writer a body goes through carries a bigstring path alongside the string one:

```ocaml
(* Cohttp.S.IO *)
val write_bigstring : oc -> Bigstring.t -> int -> int -> unit t

(* Cohttp.S.Http_io *)
val write_bigstring_body : writer -> Bigstring.t -> int -> int -> unit IO.t
```

`Transfer_io.Make(IO).writer` becomes a record of the two writes rather than a bare `string -> unit IO.t`. It is abstract in both `transfer_io.mli` and `S.Http_io`, so that is internal.

`Cohttp_lwt.Body.t` gains a `` `Bigstring `` case, with `of_bigstring` and `to_bigstring`, and the Lwt server and client send it through `write_bigstring_body`. `Body.write_body` keeps its signature and its `` `Bigstring `` arm copies onto the heap, so callers outside cohttp are unaffected.

## No new dependency

`Cohttp.Bigstring` spells the Bigarray type rather than taking it from a library:

```ocaml
type t = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
```

This is the same type as `Bigstringaf.t`, `Lwt_bytes.t` and Async's `Bigstring.t`, so a caller holding any of them passes it directly and `cohttp` gains nothing to depend on.

## Backends

Every IO backend writes without building a string:

| Backend | How |
| --- | --- |
| `cohttp-lwt-unix` | `Lwt_io.write_from_exactly_bigstring` |
| `cohttp-async` | `Writer.write_bigstring` |
| `cohttp-eio` | `Buf_write.cstruct` on `Cstruct.of_bigarray` |
| `cohttp-mirage` | `Channel.write_buffer` on `Cstruct.of_bigarray` |

The two `Buffer`-backed ones (`Cohttp.String_io`, `Cohttp_lwt.String_io`) build a string, since `Buffer` has no bigarray sink and a per-byte `add_char` loop would be slower than one `add_string`.

`Body.to_bigstring` folds each piece of a stream in as it arrives, so a body that is off the heap to keep it off never sits on it whole on the way.

## Testing

`dune build` and the suites for `cohttp`, `cohttp-lwt-unix`, `cohttp-async`, `cohttp-eio` and `cohttp-server-lwt-unix` pass — 118 tests. Note they need `-j1`: several bind fixed ports and collide with each other when dune runs them in parallel, which is pre-existing and not from this change.

Two existing spots needed updating for the widened signatures: the `Test_io` stub in `cohttp/test/test_request.ml`, and a now non-exhaustive match on `Body.t` in `cohttp-lwt-unix/test/test_client.ml`.

`cohttp-eio/tests/test_forward_proxy.ml` is left alone although `dune build @fmt` flags it — that is pre-existing drift against the pinned ocamlformat, unrelated to this change.

### DIFF
--- a/cohttp/src/bigstring.ml
+++ b/cohttp/src/bigstring.ml
@@ -1,0 +1,28 @@
+type t = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
+type write = [ `Copy of t | `Passthrough of t ]
+
+let buffer = function `Copy t | `Passthrough t -> t
+let create len = Bigarray.Array1.create Bigarray.char Bigarray.c_layout len
+let length = Bigarray.Array1.dim
+let sub t ~off ~len = Bigarray.Array1.sub t off len
+
+let check name ~size ~off ~len =
+  if off < 0 || len < 0 || off > size - len then
+    invalid_arg
+      (Printf.sprintf "Cohttp.Bigstring.%s: off=%d len=%d out of bounds for %d"
+         name off len size)
+
+let blit src ~src_off dst ~dst_off ~len =
+  Bigarray.Array1.blit (sub src ~off:src_off ~len) (sub dst ~off:dst_off ~len)
+
+let blit_from_string s ~src_off t ~dst_off ~len =
+  check "blit_from_string" ~size:(String.length s) ~off:src_off ~len;
+  check "blit_from_string" ~size:(length t) ~off:dst_off ~len;
+  for i = 0 to len - 1 do
+    Bigarray.Array1.unsafe_set t (dst_off + i)
+      (String.unsafe_get s (src_off + i))
+  done
+
+let sub_string t ~off ~len =
+  check "sub_string" ~size:(length t) ~off ~len;
+  String.init len (fun i -> Bigarray.Array1.unsafe_get t (off + i))
--- a/cohttp/src/s.ml
+++ b/cohttp/src/s.ml
@@ -60,6 +60,17 @@ module type IO = sig
   (** [write oc s] will block until the complete [s] string is written to the
       output channel [oc]. *)
 
+  val write_bigstring : oc -> Bigstring.write -> int -> int -> unit t
+  (** [write_bigstring oc buf off len] writes [len] bytes of [buf] from [off] to
+      the output channel [oc].
+
+      [`Copy] takes the bytes before returning. [`Passthrough] permits writing
+      them straight out of the caller's buffer, so an implementation that does
+      so leaves them borrowed until the next {!flush} of [oc] completes -- not
+      merely until this returns. An implementation with no way to write from
+      foreign memory may copy a [`Passthrough] buffer; the tag grants a licence
+      rather than demanding zero copies. *)
+
   val flush : oc -> unit t
   (** [flush oc] will return when all previously buffered content from calling
       {!write} have been written to the output channel [oc]. *)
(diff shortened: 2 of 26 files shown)


cohttp/src/bigstring.ml
```diff
@@ -0,0 +1,17 @@
...
+let create len = Bigarray.Array1.create Bigarray.char Bigarray.c_layout len
+let length = Bigarray.Array1.dim
+let sub t ~off ~len = Bigarray.Array1.sub t off len
+
+let blit src ~src_off dst ~dst_off ~len =
+  Bigarray.Array1.blit (sub src ~off:src_off ~len) (sub dst ~off:dst_off ~len)
+
+let blit_from_string s ~src_off t ~dst_off ~len =
+  for i = 0 to len - 1 do
+    Bigarray.Array1.unsafe_set t (dst_off + i)
```
avsm, line 22: This is part of the public interface, so it really needs to be boundschecked.

cohttp/src/s.ml
```diff
@@ -60,6 +60,10 @@ module type IO = sig
   (** [write oc s] will block until the complete [s] string is written to the
       output channel [oc]. *)
 
+  val write_bigstring : oc -> Bigstring.t -> int -> int -> unit t
```
avsm: it's worth thinking about how to minimise copying in this interface. Both Eio and Async prefer 'scheduling' a bigstring, since for a zero copy write the caller's buffer has to stay unmodified until flush. I'm not sure it would fit with Cohttp's IO monad though, I haven't thought that bit through yet.

cohttp/src/bigstring.ml
toots: Addressed!

cohttp/src/s.ml
toots: Great point. I added an explicit `` `Copy of t | `Passthrough of t`` because I think the caller should have to think about it at each call site, it's an important aspect of this API. This is only prescriptive though: some backend have to copy on each call so the real contract is: `` `Copy`` does guarantee that you can re-use the buffer after calling the function. `` `Passthrough`` is best-effort, depending on the underlying backend..

avsm: Hm, I'm not convinced. Why not just make passthrough the default and implement copy over it? Or the caller could just take a copy and then pass that through. Why does the interface need both?

toots: Some backends will copy regardless so making passthrough default can be deceiving.

From a code logic, if you wanted a default, I'd say it should be copy because this is the one safe behavior that is achievable by all backends. But I suspect most people will want or expect passthrough as default.

However, passthrough is risky: the caller should know that the passed data should _not_ be modified after the call returns otherwise, copied data can get corrupted. I've had this issue recently actually.

I feel like this is something most developers will want to know about and have to think about so a short friction that forces an explicit param and, hopefully to read the doc, would be beneficial here.

avsm: Right, but if we're copying, then the point of this PR (to improve performance) is now gone. Also if we leave the copy to the caller, then they can choose between a bigstring or bytes and take care of that themselves. A passthrough being the only option seems better for this interface. And that means just making write_bigstring accept the schedule interface.

The other option is also not to invade Cohttp.S for this, but to put this directly into an extended backend signature that knows what to do with it (cohttp-eio, cohttp-async and cohttp-lwt would all do different things with bigstrings)

toots: Ok. Can you elaborate more on the second option?

PR #1149, opened 2026-09-05
cohttp-eio: continue a partial body delivery from the position already delivered

jeong-sik:
`Reader_flow.single_read` (cohttp-eio/src/utils.ml) hands one body chunk to the caller in as many `single_read` calls as the caller's buffer needs. When a chunk does not fit, the first call copies `buffer[pos..]` correctly, but every later partial delivery copies from offset `0` of the chunk instead of from `pos`:

```ocaml
else
  let () = Cstruct.blit_from_string buffer 0 output 0 output_length   (* should be pos *)
  and () = t.buffered <- Some (buffer, pos + output_length) in
  output_length
```

`t.buffered` still advances, so the bytes that should have been delivered are lost and the start of the chunk is delivered again in their place. Any reader whose buffer is smaller than the remaining chunk twice in a row sees it; an `Eio.Buf_read` over the body flow gets there whenever its free space is smaller than an HTTP chunk. On a server-sent-events stream the corruption shows up as a JSON frame with a span repeated right after itself, e.g. `"model":"x","system_fingerprint","system_fingerprint":"fp_ollama"`.

We hit this from an OCaml client talking to three unrelated OpenAI-compatible providers (Ollama cloud, Moonshot, z.ai) at rates between 0.1% and 56% of streams depending on their chunk sizes; `curl` against the same endpoints never saw a broken frame.

The fix is the one-character change to use `pos`. The added test in `cohttp-eio/tests/test.ml` serves a 3200-byte chunked body and reads it through `Cohttp_eio.Client` 100 bytes at a time; it fails on `main` and passes with the fix.

Disclosure: the diagnosis, patch and test were prepared with AI assistance (Claude); I reviewed them and take responsibility for the change.


talex5 (approved):
This fix looks right to me - thanks!

(it's also surprising that there weren't any tests of the `Client` module in `test.ml` previously!)

